PR #63, opened 2026-04-25
ci: expand golangci lint coverage

tomo-chan:
## Summary
- enable additional golangci-lint linters requested for Go code
- configure misspell locale, whitespace handling, revive defaults, and exclusion presets
- fix existing code issues surfaced by the added linters

## Test plan
- [x] golangci-lint config verify
- [x] make build-frontend
- [x] make lint-go
- [x] make test-go

### DIFF
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -11,6 +11,8 @@ import (
 
 var tmuxSessionNameRe = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
 
+const paneTypeSSHTmux = "ssh_tmux"
+
 // Validate checks the configuration for correctness.
 // It collects all errors and returns them as a single combined error.
 func (c *Config) Validate() error {
@@ -125,7 +127,7 @@ func validatePane(p *PaneConfig, sshConns map[string]SSHConnection) []string {
 	}
 
 	switch p.Type {
-	case "local", "ssh", "tmux", "ssh_tmux":
+	case "local", "ssh", "tmux", paneTypeSSHTmux:
 		// valid
 	default:
 		errs = append(
@@ -138,7 +140,7 @@ func validatePane(p *PaneConfig, sshConns map[string]SSHConnection) []string {
 		)
 	}
 
-	if p.Type == "ssh" || p.Type == "ssh_tmux" {
+	if p.Type == "ssh" || p.Type == paneTypeSSHTmux {
 		if p.Connection == "" {
 			errs = append(errs, fmt.Sprintf("pane %q: ssh connection name must not be empty", p.ID))
 		} else if sshConns != nil {
@@ -152,7 +154,7 @@ func validatePane(p *PaneConfig, sshConns map[string]SSHConnection) []string {
 		errs = append(errs, fmt.Sprintf("pane %q: shell must be an absolute path, got %q", p.ID, p.Shell))
 	}
 
-	if p.Type == "tmux" || p.Type == "ssh_tmux" {
+	if p.Type == "tmux" || p.Type == paneTypeSSHTmux {
 		if p.TmuxSession == "" {
 			errs = append(errs, fmt.Sprintf("pane %q: tmux_session must not be empty", p.ID))
 		} else if !tmuxSessionNameRe.MatchString(p.TmuxSession) {
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,27 +7,55 @@ linters:
   default: standard
   enable:
     - errcheck
+    - errname
+    - goconst
     - govet
     - ineffassign
     - funlen
     - lll
+    - makezero
+    - misspell
+    - nilerr
+    - perfsprint
+    - revive
     - staticcheck
     - unused
+    - whitespace
+    - wrapcheck
   settings:
+    misspell:
+      locale: US
+    revive:
+      enable-default-rules: true
+      rules:
+        - name: exported
+          disabled: true
+        - name: package-comments
+          disabled: true
+        - name: unused-parameter
+          disabled: true
     staticcheck:
       checks: ["all"]
+    whitespace:
+      multi-if: true
   exclusions:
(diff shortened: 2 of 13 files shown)


tomo-chan (commented):
## Code Review

### What's Good

**Linter selection is well-considered**

| Linter | Value |
|---|---|
| `perfsprint` | Eliminates unnecessary allocations from `fmt.Sprintf`/`fmt.Errorf`. The bulk of fixes in this PR are consistent and thorough. |
| `wrapcheck` | Forces error context on all external package errors. The `findGit` and `SaveLayout` improvements are good examples of the value this surfaces. |
| `goconst` | The repeated `"ssh_tmux"` literal in three places was a real maintenance hazard — extracting `paneTypeSSHTmux` is the right call. |
| `misspell` | Spelling normalization is correct; `locale: US` makes the intent explicit. |
| `nilerr` / `errname` / `makezero` | Low-cost linters that catch classes of bugs before they ship. Good additions. |
| `revive` | Disabling `exported`, `package-comments`, and `unused-parameter` is a pragmatic choice for the current state of the codebase. |

**Code fix quality**

- `errors.New` over `fmt.Errorf` for static strings is the correct idiomatic Go pattern.
- `strconv.Itoa` over `fmt.Sprintf("%d", n)` removes format-string parsing overhead and reads more clearly.
- String concatenation replacing trivial `fmt.Sprintf` calls (e.g. `"http://" + srv.Addr()`) is cleaner.
- Wrapping the `os.WriteFile` error in `SaveLayout` with context goes beyond satisfying `wrapcheck` — it genuinely improves debuggability.

---

### Issues & Questions

**[Minor question] `std-error-handling` exclusion preset**

The `std-error-handling` preset silences some `errcheck`/`wrapcheck` findings for standard library calls. Worth adding a short comment in `.golangci.yml` explaining the intent, so future contributors don't wonder why some stdlib errors aren't checked.

**[Question] Behavioral change in `config.LoadOrDefault()`**

`wrapcheck` surfaced a dropped error, and the fix changes the contract of the function:

```go
// Before: silently falls back to defaults when DefaultConfigPath() fails
if err \!= nil {
    return Default(), nil
}

// After: propagates the error
if err \!= nil {
    return nil, fmt.Errorf("default config path: %w", err)
}
```

This is probably the better behavior, but it means a `os.UserHomeDir()` failure now becomes a fatal startup error instead of a silent fallback. Two things worth confirming:

1. Does `config_test.go` have a test covering the old fallback path? If so, it needs updating.
2. When this error surfaces in `main.go`'s `loadConfig`, is the message shown to the user clear enough?

**[Question] `server.Start()` error wrapping and `http.ErrServerClosed`**

`http.ListenAndServe` returns `http.ErrServerClosed` on graceful shutdown. Wrapping with `%w` keeps `errors.Is(err, http.ErrServerClosed)` working, but any caller using direct equality (`err == http.ErrServerClosed`) would silently break. The call site in `main.go` isn't included in this diff — can you confirm it uses `errors.Is`?

**[Suggestion] Consider exporting `paneTypeSSHTmux`**

The constant is correctly scoped to `validate.go` for now. That said, the string `"ssh_tmux"` also appears in `session/factory.go` and related files. Exporting it as `config.PaneTypeSSHTmux` would eliminate cross-package duplication. Happy to take this as a follow-up issue if it's out of scope here.

internal/config/validate.go
```diff
@@ -11,6 +11,8 @@ import (
 
 var tmuxSessionNameRe = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
 
+const paneTypeSSHTmux = "ssh_tmux"
+
```
tomo-chan, line 15: **[Suggestion]** `paneTypeSSHTmux` eliminates the repeated literal within this file. The same string `"ssh_tmux"` appears in `session/factory.go` too — exporting this as `config.PaneTypeSSHTmux` would let the session package reference the canonical value. Out of scope for this PR, but worth a follow-up issue.

tomo-chan (commented):
## Code Review

### What's Good

- **Linter selection is well-considered.** `perfsprint`, `wrapcheck`, `goconst`, `misspell`, and the rest are all low-noise, high-value additions for this codebase. The revive rule exclusions (`exported`, `package-comments`, `unused-parameter`) are a pragmatic fit for the current state of the project.
- **The `std-error-handling` preset now has a comment** (`.golangci.yml`) explaining why it's included. Makes the intent clear for future contributors.
- **`LoadOrDefault()` correctly preserves the historical fallback.** Extracting `defaultAfterConfigPathError()` with an explanatory comment is a clean way to satisfy `wrapcheck` without silently changing startup behavior.
- **`server.Start()` re-returns `http.ErrServerClosed` unwrapped.** The explicit `errors.Is` guard before the wrapping `fmt.Errorf` means callers using either `==` or `errors.Is` continue to work correctly.
- **Mechanical fixes are consistent and thorough.** `errors.New` over static `fmt.Errorf`, `strconv.Itoa` over `fmt.Sprintf("%d", n)`, string concatenation for simple prefix/suffix — all idiomatic and applied uniformly across the codebase.
- **`paneTypeSSHTmux` constant** eliminates three repeated `"ssh_tmux"` literals in `validate.go`. Reduces the risk of a future typo silently bypassing validation.

---

### Issues & Questions

**[Suggestion] Consider exporting `paneTypeSSHTmux`**

The constant is correctly scoped to `validate.go` for now. The string `"ssh_tmux"` also appears in `session/factory.go` and related files, which means the same literal is duplicated across packages. Exporting it as `config.PaneTypeSSHTmux` would let the session package reference the canonical value instead of redeclaring it. Happy to open a follow-up issue if this is out of scope here.

---

### Overall

All previously raised concerns have been addressed. This is a clean, well-scoped lint hygiene PR that also improves error context in a few paths. Good to merge once the `paneTypeSSHTmux` export question is triaged (follow-up issue or in-PR fix).

PR #67, opened 2026-04-25
fix: tighten config permissions with gosec lint

tomo-chan:
## Summary
- enable gosec in golangci-lint
- tighten config file permissions to 0600 when loading existing configs and when saving new changes
- warn and continue if an existing readable config cannot be chmod'd
- document targeted gosec suppressions for trusted OS/protocol integration points

## Test plan
- [x] go test ./internal/config -run 'TestLoad_(TightensExistingFilePermissions|ChmodFailureWarnsAndContinues)' -count=1
- [x] make lint-go
- [x] make test-go

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -10,6 +10,8 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const configFileMode os.FileMode = 0600
+
 type ServerConfig struct {
 	Host string `yaml:"host"`
 	Port int    `yaml:"port"`
@@ -83,6 +85,9 @@ func Load(path string) (*Config, error) {
 
 	if err := cfg.Validate(); err != nil {
 		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+	if err := tightenConfigFilePermissions(path); err != nil {
+		return nil, err
 	}
 
 	return &cfg, nil
@@ -202,8 +207,22 @@ func (c *Config) SaveLayout(layout LayoutNode) error {
 	if err != nil {
 		return fmt.Errorf("marshaling config: %w", err)
 	}
-	if err := os.WriteFile(c.filePath, data, 0600); err != nil {
+	if err := os.WriteFile(c.filePath, data, configFileMode); err != nil {
 		return fmt.Errorf("writing config: %w", err)
+	}
+	return nil
+}
+
+func tightenConfigFilePermissions(path string) error {
+	info, err := os.Stat(path)
+	if err != nil {
+		return fmt.Errorf("checking config permissions: %w", err)
+	}
+	if info.Mode().Perm() == configFileMode {
+		return nil
+	}
+	if err := os.Chmod(path, configFileMode); err != nil {
+		return fmt.Errorf("tightening config permissions: %w", err)
 	}
 	return nil
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -90,7 +90,7 @@ func Load(path string) (*Config, error) {
 		return nil, fmt.Errorf("invalid config: %w", err)
 	}
 	if err := tightenConfigFilePermissions(path); err != nil {
-		//nolint:gosec // local filesystem warning
+		//nolint:gosec // G706: local filesystem warning
 		log.Printf("Warning: failed to tighten config file permissions to 0600: %v", err)
 	}
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -202,7 +202,7 @@ func (c *Config) SaveLayout(layout LayoutNode) error {
 	if err != nil {
 		return fmt.Errorf("marshaling config: %w", err)
 	}
-	if err := os.WriteFile(c.filePath, data, 0644); err != nil {
+	if err := os.WriteFile(c.filePath, data, 0600); err != nil {
 		return fmt.Errorf("writing config: %w", err)
 	}
 	return nil
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -3,6 +3,7 @@ package config
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -11,6 +12,8 @@ import (
 )
 
 const configFileMode os.FileMode = 0600
+
+var chmodConfigFile = os.Chmod
 
 type ServerConfig struct {
 	Host string `yaml:"host"`
@@ -87,7 +90,8 @@ func Load(path string) (*Config, error) {
 		return nil, fmt.Errorf("invalid config: %w", err)
 	}
 	if err := tightenConfigFilePermissions(path); err != nil {
-		return nil, err
+		//nolint:gosec // local filesystem warning
+		log.Printf("Warning: failed to tighten config file permissions to 0600: %v", err)
 	}
 
 	return &cfg, nil
@@ -221,7 +225,7 @@ func tightenConfigFilePermissions(path string) error {
 	if info.Mode().Perm() == configFileMode {
 		return nil
 	}
-	if err := os.Chmod(path, configFileMode); err != nil {
+	if err := chmodConfigFile(path, configFileMode); err != nil {
 		return fmt.Errorf("tightening config permissions: %w", err)
 	}
 	return nil
--- a/internal/ws/handler.go
+++ b/internal/ws/handler.go
@@ -52,7 +52,8 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	conn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
-		log.Printf("ws upgrade error for session %s: %v", sessionID, err) //nolint:gosec // config ID
+		//nolint:gosec // sessionID is a config-defined identifier
+		log.Printf("ws upgrade error for session %s: %v", sessionID, err)
 		return
 	}
 	defer conn.Close() //nolint:errcheck
--- a/internal/ws/handler.go
+++ b/internal/ws/handler.go
@@ -52,7 +52,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	conn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
-		log.Printf("ws upgrade error for session %s: %v", sessionID, err)
+		log.Printf("ws upgrade error for session %s: %v", sessionID, err) //nolint:gosec // config ID
 		return
 	}
 	defer conn.Close() //nolint:errcheck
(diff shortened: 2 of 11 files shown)


internal/ws/handler.go
```diff
@@ -52,7 +52,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	conn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
-		log.Printf("ws upgrade error for session %s: %v", sessionID, err)
+		log.Printf("ws upgrade error for session %s: %v", sessionID, err) //nolint:gosec // config ID
 		return
```
tomo-chan, line 57: The other two suppressions in this file use `// sessionID is a config-defined identifier`, which is clearer about the URL-input-validated-against-config distinction. Aligning this one would keep all three consistent.

tomo-chan (commented):
## Code Review

### What's Good

- **Permission tightening** (`config.go`): Changing `0644` → `0600` is a meaningful security improvement — config YAML can contain SSH hostnames and authentication paths.
- **Test upgrade** (`config_test.go`): Promoting `assert.NoError` → `require.NoError` and adding an explicit `info.Mode().Perm()` assertion is the right pattern. The stat result is now verified, not just present.
- **Scoped suppressions**: Every `//nolint:gosec` is line-level with a clear rationale rather than blanket file-level exclusions.
- **Comment style consistency**: Standardising to `// explanation` (replacing the `--` variant in `handler_test.go`) keeps nolint comments idiomatic.
- **`crypto/sha1` suppression** (`ssh.go`): The comment accurately captures *why* SHA-1 is acceptable — it's required by the OpenSSH known_hosts HMAC format, not a free choice.
- **`tmuxSession` validation backs the suppression** (`tmux_local.go`): `^[a-zA-Z0-9_.-]+$` in `validate.go` correctly rejects tmux's target-syntax special characters (`:`, `$`, `%`, `@`), so the "validated config" claim in the nolint comment holds up.

---

### Issues & Questions

**[Minor question] Gosec rule IDs not specified in suppressions**

None of the new `//nolint:gosec` comments name the specific rule being suppressed (e.g. `G204`, `G304`, `G501`). Without the rule ID, a future gosec update that adds a new finding on the same line would be silently suppressed. Was the omission intentional (preferring brevity), or is there an opportunity to be more precise, e.g.:

```go
//nolint:gosec // G204: lsof is a trusted system binary; pid is an internal int
```

**[Minor question] Existing config files retain `0644` after upgrade**

`SaveLayout` now writes with `0600`, but config files created before this change won't have their permissions tightened on upgrade. Is a release note or a brief upgrade instruction worth adding (e.g. `chmod 0600 ~/.config/panemux/config.yaml`)?

**[Minor question] First `ws/handler.go` nolint comment less precise than the others**

The upgrade-error log line uses `// config ID` while the two `forwardTerminalOutput`/`handleWebSocketMessage` nolints in the same file use `// sessionID is a config-defined identifier`. The shorter wording could be read as "the value came directly from config" rather than "URL input validated against config keys". Aligning the wording across all three would keep it consistent.

---

### Overall

Clean, targeted hardening with solid test coverage for the permission change. All three findings are documentation/style suggestions; none block merge.

internal/config/config.go
```diff
@@ -84,6 +86,9 @@ func Load(path string) (*Config, error) {
 	if err := cfg.Validate(); err != nil {
 		return nil, fmt.Errorf("invalid config: %w", err)
 	}
+	if err := tightenConfigFilePermissions(path); err != nil {
+		return nil, err
```
tomo-chan: If `os.Chmod` fails here (read-only FS, different file owner, container bind-mount), `Load()` returns an error and panemux won't start even though the config is valid and readable. Is this the intended policy, or would a warning-and-continue be more appropriate?

internal/ws/handler.go
```diff
@@ -52,6 +52,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	conn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
+		//nolint:gosec // sessionID is a config-defined identifier
 		log.Printf("ws upgrade error for session %s: %v", sessionID, err)
```
tomo-chan, line 56: The other two suppressions in this file are end-of-line; this one is now preceding-line. Both are valid for golangci-lint, but mixing styles in the same file is slightly inconsistent. Was this moved to avoid a line-length lint hit?

tomo-chan (commented):
## Code Review

### What's Good

*(New in this round)*

- **`configFileMode` constant** (`config.go`): Replacing the magic number with a named constant removes duplication and makes future permission changes a one-line update.
- **`tightenConfigFilePermissions`**: Directly addresses the upgrade-migration concern from the previous review — existing files with looser permissions are now tightened at load time, not only on the next save.
- **`TestLoad_TightensExistingFilePermissions`**: Explicitly sets the file to `0644` before loading and asserts `0600` afterwards, covering the migration path end-to-end.
- **`ws/handler.go` wording aligned**: All three suppressions now use the consistent `// sessionID is a config-defined identifier` phrasing.

---

### Issues & Questions

**[Question] `Load()` hard-fails when `os.Chmod` cannot tighten permissions**

```go
// internal/config/config.go — Load()
if err := tightenConfigFilePermissions(path); err \!= nil {
    return nil, err   // startup is aborted
}

// tightenConfigFilePermissions
if err := os.Chmod(path, configFileMode); err \!= nil {
    return fmt.Errorf("tightening config permissions: %w", err)
}
```

If the config file cannot be chmod'd — e.g. it is on a read-only filesystem, bind-mounted in a container, or owned by a different user than the panemux process — `Load()` now fails entirely, even though the config content was successfully read. Previously this scenario would have started panemux successfully.

Is aborting startup the intended policy? The current behaviour means "readable config that can't have its permissions tightened" becomes a hard blocker. A softer approach (log a warning and continue) would be more resilient for non-standard deployment topologies, while still communicating the security gap.

**[Minor question] Mixed `//nolint` placement style in `ws/handler.go`**

```go
// First suppression — now preceding-line:
//nolint:gosec // sessionID is a config-defined identifier
log.Printf("ws upgrade error for session %s: %v", sessionID, err)

// Other two — still end-of-line:
log.Printf("session %s read error: %v", sessionID, err)  //nolint:gosec // sessionID is a config-defined identifier
log.Printf("session %s write error: %v", sessionID, err) //nolint:gosec // sessionID is a config-defined identifier
```

Both placements are valid for golangci-lint, but mixing them in the same file is slightly inconsistent. Was the preceding-line style chosen to avoid a line-length lint violation? If so, aligning all three to preceding-line would be cleaner; otherwise the end-of-line style could be restored.

**[Minor question] (Carried from previous review) Gosec rule IDs still not specified**

The comment wording is now consistent across all suppressions — that part of the previous concern is resolved. The rule IDs (e.g. `G204`, `G304`, `G501`) are still absent. Without them, a new gosec finding on the same line in a future version would be silently suppressed. No action needed if brevity is preferred.

---

### Overall

The two new commits close the permission-migration and comment-precision gaps from the previous round. The [Question] about hard-failing on a failed `os.Chmod` is the only item that may warrant discussion before merge; the remaining findings are stylistic and can be addressed at discretion.

internal/ws/handler.go
tomo-chan: Addressed. The upgrade-error suppression comment now uses the same `sessionID is a config-defined identifier` wording as the other WebSocket suppressions. I also added the explicit `G706:` rule ID in the follow-up change.

internal/config/config.go
tomo-chan: Addressed. `Load()` no longer fails when `chmod` cannot tighten the config file permissions. It now logs a warning and continues with the successfully read config, so read-only filesystems, bind mounts, and similar deployments are not blocked. I added `TestLoad_ChmodFailureWarnsAndContinues` to cover this behavior.

internal/ws/handler.go
tomo-chan: Addressed. All three `//nolint:gosec` suppressions in `ws/handler.go` now use the preceding-line style. This keeps the placement consistent while avoiding the line-length lint issue, and each suppression now includes the same explicit `G706:` explanation.

tomo-chan:
I also addressed the gosec rule ID note from the review summary. Every `//nolint:gosec` comment now names the relevant rule ID (`G204`, `G302`, `G306`, `G505`, `G703`, or `G706`) at the start of the rationale. Verification: `make lint-go` passes.

tomo-chan (commented):
## Code Review

### What's Good

*(New in this round — all three previous concerns resolved)*

- **Soft-fail for `chmod`** (`config.go`): `tightenConfigFilePermissions` failure now logs a warning and lets `Load()` succeed, directly addressing the startup-blocking concern from the last round.
- **Injectable `chmodConfigFile` variable**: Clean Go testability pattern — no filesystem side-effects needed to exercise the failure path.
- **`TestLoad_ChmodFailureWarnsAndContinues`**: Injects the failure via `chmodConfigFile`, captures log output with `log.SetOutput`, and asserts both the warning message and a valid returned config. Complete coverage of the failure branch.
- **Gosec rule IDs now specified** across all suppressions — the documentation precision concern is resolved.
- **`ws/handler.go`**: All three suppressions are now consistently preceding-line style with matching `G706` IDs.
- **`session/ssh.go` ProxyCommand suppression** also updated for consistency with the rest of the PR.

---

### Issues & Questions

~~**[Minor question] G703 and G706 don't appear in gosec's published rule set**~~

~~`G703` (cited for `os.Stat(cwd)` in `api/handler.go`) and `G706` (cited for `log.Printf` lines in `ws/handler.go` and `config.go`) are not in gosec's documented rule list, which runs G101–G602. The `//nolint:gosec` directive suppresses all gosec findings on a line regardless of the text that follows `//`, so the suppressions work correctly — but incorrect rule IDs in the explanation would mislead a future reader who tries to look up the risk.~~

~~Were these IDs from actual `golangci-lint run` output, or inferred? A quick check like:~~

~~`golangci-lint run --enable gosec ./... 2>&1 | grep -oP 'gosec\.\K[A-Z][0-9]+' | sort -u`~~

~~would confirm the real IDs. If gosec is genuinely firing G703/G706 (perhaps via a custom or extended ruleset), the comments are accurate as-is.~~

> **Retracted** — G703 and G706 are valid gosec rules introduced in v2.24.0 (February 2025) as taint analysis rules. The rule IDs are correct. See [correction comment](https://github.com/tomo-chan/panemux/pull/67#pullrequestreview-4175649361).

---

### Overall

All three concerns from the previous round are cleanly resolved. No outstanding issues — ready to merge.

tomo-chan (commented):
## Correction to Previous Review

**Retracting the [Minor question] about G703 and G706 from my last review.**

Those rule IDs are valid. G703 (path traversal via taint analysis) and G706 (log injection via taint analysis) were introduced in gosec v2.24.0 (February 2025) as part of a new G7xx taint analysis rule family. My assertion that they don't exist in gosec's rule set was incorrect — my knowledge of the rule list was simply out of date.

The rule IDs in the suppression comments are accurate. No changes needed.

PR #69, opened 2026-04-26
feat: add workspace tabs

tomo-chan:
## Summary
- add workspace config normalization with legacy layout migration on save
- add workspace APIs, edit-mode workspace creation/deletion, persisted active workspace selection, and tabbed frontend layout switching
- hide single-workspace tabs during normal use and document workspace add/delete behavior

## Test plan
- [x] make check


frontend/src/hooks/useLayout.ts
```diff
@@ -1,20 +1,31 @@
...
+      .then((data) => {
+        const workspaceResult = WorkspacesResponseSchema.safeParse(data)
+        if (workspaceResult.success) {
+          const parsed = workspaceResult.data
+          setWorkspaces(parsed)
+          const active = parsed.items.find((workspace) => workspace.id === parsed.active) ?? parsed.items[0]
+          setLayout(active.layout)
+          return
+        }
+        setLayout(LayoutNodeSchema.parse(data))
```
tomo-chan: This fallback is unreachable via `/api/workspaces` — the endpoint always returns workspace format. Against a truly old backend, the 404 is caught by `if (\!r.ok) throw` before reaching this branch. Intentional rolling-upgrade guard, or can it be removed?

tomo-chan (commented):
## Code Review

### What's Good

- **Migration strategy**: `normalizeWorkspaces()` transparently converts legacy `layout:` configs into a single `default` workspace on the next save — non-breaking and self-healing for existing users.
- **`write()` extraction**: consolidates file I/O that was previously duplicated in `SaveLayout`, and the new workspace-only serialization struct correctly omits the top-level `layout` key on roundtrip.
- **Test breadth**: the suite covers migration, persistence, active-workspace switching, per-workspace layout updates, and fallback behaviors — clearly following the new test-case design rules added in this same PR.
- **ARIA baseline**: `WorkspaceTabs` uses `role="tablist"`, `role="tab"`, `aria-selected`, and `aria-orientation` — a solid accessibility foundation for a terminal UI.
- **`encodeURIComponent`** on workspace IDs in URL construction — correct; workspace IDs are user-supplied strings.
- **Cross-workspace uniqueness validation**: pane IDs are validated as globally unique across all workspaces, which is required because sessions/WebSockets are keyed by pane ID.
- **Docs in the same change**: README, `config.example.yaml`, `docs/architecture.md`, and `AGENTS.md` all updated together.

---

### Issues & Questions

**[Minor question] Unreachable fallback in `useLayout.ts`**

`frontend/src/hooks/useLayout.ts`, after the workspace schema parse attempt:

```ts
setLayout(LayoutNodeSchema.parse(data))
```

This runs when `WorkspacesResponseSchema.safeParse(data)` fails on a successful HTTP 200 from `/api/workspaces`. But the endpoint always returns workspace format now, so this path is only reachable against a hypothetical old backend that served layout-only JSON at `/api/workspaces`. Against a truly old backend, the frontend would receive HTTP 404 (caught by the `if (\!r.ok) throw` guard) before this line — so the fallback appears unreachable. Is it intended as a guard for rolling upgrades, or can it be removed?

---

**[Minor question] Tab bar always visible with a single workspace**

In `App.tsx`:

```tsx
{workspaces && (
  <WorkspaceTabs workspaces={workspaces.items} ... />
)}
```

This renders the tab bar even when `workspaces.items.length === 1`, which is the default post-migration state for most existing users (they get a single `Default` workspace). Is showing a one-tab rail intentional? Hiding it at `<= 1` item would avoid visible UI change for single-workspace setups.

---

**[Minor question] `normalizeWorkspaces()` called on reads**

`WorkspacesView()`, `AllPanes()`, `UpdateLayout()`, `RemovePaneFromLayout()`, and `Validate()` all call `normalizeWorkspaces()` defensively. This means read-only operations silently mutate `c.Workspaces`. The function is idempotent in practice, but it makes methods harder to reason about (e.g., a test calling `cfg.AllPanes()` normalizes the config as a side effect). Is there a reason exhaustive normalization at `Load()` time isn't sufficient, so it can be removed from the read methods?

---

**[Minor question] Active workspace not persisted unless edit mode is on**

`PutActiveWorkspace` calls `SaveWorkspaces()` only when `h.editMode.Load()` is true. Switching the active workspace without edit mode enabled therefore doesn't survive a server restart. Layout changes follow the same gating, so this is consistent — but workspace tab selection feels more like a UI preference than a layout edit. Is the edit-mode requirement here intentional?

---

### Overall

Solid implementation with a clean migration path and thorough test coverage. All four findings above are minor questions about intent — none block merge. The review is ready to approve once the author confirms (or intentionally leaves) the fallback path, single-workspace visibility, normalization-on-read, and edit-mode persistence behavior.

frontend/src/hooks/useLayout.ts
tomo-chan: Addressed in commit 293c907.

- Removed the unreachable `/api/workspaces` legacy layout fallback.
- Hid workspace tabs for single-workspace normal mode while keeping an edit-mode add control.
- Persisted active workspace selection even when edit mode is off.
- Moved workspace normalization away from read helpers so reads do not mutate config state.

`make check` passed after the changes.

frontend/src/hooks/useLayout.ts
```diff
@@ -32,16 +38,41 @@ export function useLayout() {
...
+
+    setWorkspaces({ ...workspaces, active: workspaceID })
+    setLayout(target.layout)
+    await fetch('/api/workspaces/active', {
+      method: 'PUT',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ id: workspaceID }),
+    }).catch(console.error)
+  }, [workspaces])
+
```
tomo-chan, line 68: If this fetch fails (network error, unexpected 500, or 403 if edit-mode state drifts between client and server), the thrown error becomes an unhandled promise rejection — no `setError`, no user feedback. The mount effect uses `.catch((e) => setError(e.message))` and `setActiveWorkspace` uses `.catch(console.error)` as a floor. Wrapping the body in try/catch and calling `setError` on failure would surface the error consistently.

tomo-chan (commented):
## Code Review (follow-up)

All four concerns from the prior review are addressed in commit 293c907e:

- **Unreachable fallback removed**: `useLayout.ts` now uses `WorkspacesResponseSchema.parse(data)` directly; the dead-code `LayoutNodeSchema` fallback and its import are gone.
- **Single-workspace tab bar**: Hidden when `workspaces.items.length <= 1` in normal use. Edit mode shows only the `+` add button. README and `config.example.yaml` both document this behavior.
- **Mutation-on-read fixed**: `normalizedWorkspaces()` is a pure function returning a value without touching `c.Workspaces`. `WorkspacesView()`, `AllPanes()`, and `Validate()` all use it. `TestReadMethods_DoNotNormalizeConfigInPlace` verifies the invariant.
- **Active workspace always persisted**: `PutActiveWorkspace` now saves unconditionally; `TestPutActiveWorkspace_PersistsWithoutEditMode` confirms the disk round-trip.

### What's Good (new in this commit)

- `normalizedWorkspaces()` as a pure function is a clean architectural fix — separating "pure view" from "mutating normalization" makes every config method straightforward to reason about.
- `PostWorkspace` is correctly gated on edit mode (403 otherwise), starts the session immediately, and persists before responding — correct ordering for the happy path.
- `nextWorkspaceID` / `nextPaneID` handle ID collisions with a seen-set loop — safe for any realistic number of workspaces.
- `TestPostWorkspace_EditModeOn_AddsDefaultLocalWorkspaceAndPersists` is a thorough integration test: verifies session added to manager, workspace in response, active ID set, disk round-trip, and absence of the legacy `layout:` key.

---

### Issues & Questions

**[Question] `addWorkspace` errors surface as unhandled promise rejections**

In `frontend/src/hooks/useLayout.ts`:

```ts
const addWorkspace = useCallback(async () => {
  const response = await fetch('/api/workspaces', { method: 'POST' })
  if (\!response.ok) throw new Error(`HTTP ${response.status}`)
  const parsed = WorkspacesResponseSchema.parse(await response.json())
  setWorkspaces(parsed)
  const active = parsed.items.find((w) => w.id === parsed.active) ?? parsed.items[0]
  setLayout(active.layout)
}, [])
```

If this fetch fails (network error, unexpected 500, or a `403` if edit-mode state is somehow out of sync between client and server), the thrown error becomes an **unhandled promise rejection** — no `setError` call, no user-visible message. The `+` button click silently fails. Compare with the mount effect (`.catch((e) => setError(e.message))`) and `setActiveWorkspace` (`.catch(console.error)` as a minimum floor). Should `addWorkspace` wrap its body in try/catch and call `setError` on failure?

---

**[Minor question] `PostWorkspace`: partial failure leaves in-memory state ahead of disk**

In `internal/api/handler.go`:

```go
workspace := h.cfg.AddDefaultWorkspace()   // mutates c.Workspaces in memory
for _, pane := range panesInLayout(...) {
    sess, err := h.createSession(pane, ...)
    if err \!= nil { http.Error(..., 500); return }
    h.manager.Add(sess)
}
if err := h.cfg.SaveWorkspaces(); err \!= nil {
    http.Error(w, "failed to save workspaces", 500)
    return   // workspace + session are already live in memory
}
```

If `SaveWorkspaces()` fails after sessions are added, the in-memory config has the new workspace and the session manager holds the new session, but the config file is not updated. A server restart silently drops both. Is this acceptable (disk-save failures are rare and the existing codebase follows the same pattern elsewhere), or should there be a rollback of `c.Workspaces` on save failure?

---

### Overall

All previous concerns are resolved. The two new findings above are the only remaining open questions. The `addWorkspace` error-handling gap is the higher-priority one — a silent failure on a user-triggered action is a visible UX regression worth fixing before merge.

frontend/src/hooks/useLayout.ts
tomo-chan: Addressed in commit f126574.

`addWorkspace()` now wraps the POST/parse path in `try/catch`, clears stale errors before retrying, and writes failures into `setError(...)` instead of letting the click handler create an unhandled promise rejection. Added `useLayout` coverage for a 403 response to confirm the error is surfaced and the active workspace remains unchanged.

`make check` passed after the change.

tomo-chan:
Follow-up on the `PostWorkspace` save-failure question:

Leaving rollback out intentionally for this PR. `PostWorkspace` now matches the existing layout/session mutation pattern: if config persistence fails, the request returns 500 and the in-memory state may be ahead of disk until restart. That behavior is already covered by the broader save/update flow, and adding transactional rollback for sessions plus config would be a larger consistency model change than this review item needs.

The user-visible `addWorkspace()` failure path was fixed separately in commit f126574, and `make check` passed.

tomo-chan (commented):
## Code Review (follow-up 2)

**[Question] `addWorkspace` error handling** — addressed in commit f1265748.

`addWorkspace` now wraps the entire body in try/catch, clears stale errors with `setError(null)` before the request, and calls `setError(err instanceof Error ? err.message : 'Failed to add workspace')` on failure. The 403 path (edit-mode drift) is covered by a new test case in `useLayout.test.ts` that confirms `result.current.error` contains the status code while the active workspace remains unchanged. The thread has been resolved.

The outstanding **[Minor question]** about `PostWorkspace` partial failure (session added to manager before `SaveWorkspaces()` succeeds) remains open but is not blocking — disk-save failures are rare and the existing codebase treats this pattern as an acceptable corner case throughout.

---

### Overall

All blocking concerns are resolved. The PR is ready to merge.

internal/api/handler.go
```diff
@@ -108,6 +112,137 @@ func (h *Handler) PutLayout(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+	if !found {
+		http.Error(w, "workspace not found", http.StatusNotFound)
+		return
+	}
+	if len(view.Items) <= 1 {
+		http.Error(w, "cannot delete the last workspace", http.StatusConflict)
+		return
+	}
```
tomo-chan, line 173: This `\!ok` branch is unreachable under normal operation: Guard 1 already confirmed the workspace exists via `view.Items`, and both `WorkspacesView()` and `RemoveWorkspace()` normalize from the same config. It can only fire via a TOCTOU race (two concurrent DELETEs). Intentional race guard, or safe to collapse into a single check inside `RemoveWorkspace`?

tomo-chan (commented):
## Code Review (follow-up 3)

All prior findings are resolved. Both inline comment threads are closed. The `addWorkspace` error-handling [Question] was addressed in commit f1265748 (try/catch + `setError`), and the outstanding [Minor question] about `PostWorkspace` partial failure was acknowledged as an acceptable pattern.

### What's Good (new in this commit)

- **Session teardown order**: `DeleteWorkspace` persists config to disk *before* removing sessions from the manager — the correct order. A disk-save failure leaves sessions running but the workspace still on disk, so nothing is silently lost on restart. Better failure mode than `PostWorkspace` (which starts sessions before persisting).
- **Last-workspace guard**: 409 Conflict when attempting to delete the final workspace, tested in `TestDeleteWorkspace_LastWorkspace_Returns409`.
- **Confirmation dialog**: `window.confirm` before calling `deleteWorkspace` prevents accidental deletion; confirm/cancel paths are both covered in `App.test.tsx`.
- **Consistent error handling**: `deleteWorkspace` in `useLayout.ts` follows the same try/catch + `setError` pattern as `addWorkspace`, including `setError(null)` on entry. The 409 error path is tested.
- **`encodeURIComponent`** on workspace ID in the DELETE URL — consistent with other endpoints.
- **Integration test breadth**: `TestDeleteWorkspace_EditModeOn_RemovesWorkspaceSessionsAndPersists` checks that the target session is removed, the surviving session stays, the response has the correct active/items, and the disk round-trip is clean.

---

### Issues & Questions

**[Minor question] Redundant 404 path in `DeleteWorkspace`**

In `internal/api/handler.go`, the handler first scans `view.Items` to confirm the workspace exists, then calls `RemoveWorkspace` which has its own not-found return:

```go
// Guard 1 - scan view.Items (normalizedWorkspaces)
found := false
for _, workspace := range view.Items { ... }
if \!found {
    http.Error(w, "workspace not found", http.StatusNotFound)
    return
}
// ...
// Guard 2 - unreachable under normal operation
workspace, ok := h.cfg.RemoveWorkspace(id)
if \!ok {
    http.Error(w, "workspace not found", http.StatusNotFound)
    return
}
```

Both `WorkspacesView()` and `RemoveWorkspace()` normalize from the same underlying config, so Guard 2 is only reachable via a TOCTOU race (two concurrent DELETE requests). Is the double-check intentional as a defensive safety net, or would a single check inside `RemoveWorkspace` (dropping the scan loop and Guard 1) be cleaner?

---

**[Minor question] `window.confirm` as confirmation mechanism**

In `App.tsx`, deletion uses the browser native blocking dialog. It works and tests mock it cleanly, but `window.confirm` blocks the main thread and cannot be styled. If a non-blocking inline confirmation is planned for a later iteration, worth tracking as a follow-up issue. Happy to open one if useful.

---

### Overall

Workspace deletion is a clean addition with the correct session-lifecycle ordering and solid test coverage. The two findings above are non-blocking minor questions — no changes required before merge.

internal/api/handler.go
tomo-chan: Addressed in commit 39563f3.

Collapsed `DeleteWorkspace` to a single not-found check from `RemoveWorkspace(id)`, keeping only the pre-delete `len(view.Items) <= 1` guard for the last-workspace conflict case.

The same commit also fixes the failing GHA `ci` lint by extracting the repeated `horizontal`, `vertical`, and pane type strings in `internal/config/validate.go` into constants. Local verification: `make lint-go`, `make test-go`, and `make check` all passed.

tomo-chan (commented):
## Code Review (follow-up 4)

**[Minor question] Redundant 404 path in `DeleteWorkspace`** — addressed in commit 39563f36.

The `found` scan loop and the first `if \!found` check have been removed. `DeleteWorkspace` now flows directly: last-workspace guard → `RemoveWorkspace(id)` → single `\!ok` 404 → save → close sessions. Clean. The thread has been resolved.

Commit 39563f36 also extracts named constants for direction strings and pane types in `validate.go` (`directionHorizontal`, `directionVertical`, `paneTypeLocal`, `paneTypeSSH`, `paneTypeTmux`), eliminating the remaining bare string literals. This is a nice consistency improvement alongside the targeted fix.

The outstanding **[Minor question]** about `window.confirm` remains open but is non-blocking; no response is needed before merge.

---

### Overall

All blocking and non-blocking [Question] concerns are resolved. The PR is ready to merge.

PR #72, opened 2026-04-29
feat: add workspace rename

tomo-chan:
## Summary
- Add an edit-mode workspace rename API that updates only workspace titles and persists them.
- Add inline workspace tab renaming in the frontend with Enter/blur save and Escape cancel.
- Document rename support for workspace tabs.

## Test plan
- [x] make install-deps-ci
- [x] make build-frontend
- [x] make test-go
- [x] make test-frontend
- [x] make lint-go
- [x] make lint-frontend

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -67,6 +67,10 @@ type activeWorkspaceRequest struct {
 	ID string `json:"id"`
 }
 
+type workspaceRequest struct {
+	Title string `json:"title"`
+}
+
 var validHostName = regexp.MustCompile(`^[a-zA-Z0-9_.\-]+$`)
 
 // NewHandler creates a new API handler.
@@ -182,6 +186,36 @@ func (h *Handler) DeleteWorkspace(w http.ResponseWriter, r *http.Request) {
 	}
 	for _, pane := range panesInLayout(workspace.Layout) {
 		_ = h.manager.Remove(pane.ID)
+	}
+	writeJSON(w, h.cfg.WorkspacesView())
+}
+
+// PutWorkspace updates workspace metadata while edit mode is enabled.
+func (h *Handler) PutWorkspace(w http.ResponseWriter, r *http.Request) {
+	if !h.editMode.Load() {
+		http.Error(w, "edit mode required", http.StatusForbidden)
+		return
+	}
+	id := chi.URLParam(r, "id")
+	var req workspaceRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, "invalid request body", http.StatusBadRequest)
+		return
+	}
+	title := strings.TrimSpace(req.Title)
+	if title == "" {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusUnprocessableEntity)
+		_ = json.NewEncoder(w).Encode(map[string]string{"error": "workspace title must not be empty"})
+		return
+	}
+	if !h.cfg.RenameWorkspace(id, title) {
+		http.Error(w, "workspace not found", http.StatusNotFound)
+		return
+	}
+	if err := h.cfg.SaveWorkspaces(); err != nil {
+		http.Error(w, "failed to save workspaces", http.StatusInternalServerError)
+		return
 	}
 	writeJSON(w, h.cfg.WorkspacesView())
 }
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ workspaces:
 
 Older config files with a top-level `layout:` are still accepted. When the config is next saved, panemux migrates that layout into a `default` workspace and writes the `workspaces:` format.
 
-If there is only one workspace, the workspace tab bar is hidden during normal use. Enable edit mode to show the workspace add control; newly added workspaces start with a single local terminal pane, become active immediately, and are saved to the config. In edit mode, each visible workspace tab also has a delete button that asks for confirmation before removing the workspace. The last remaining workspace cannot be deleted. Switching the active workspace is persisted so the same workspace is restored after restart.
+If there is only one workspace, the workspace tab bar is hidden during normal use. Enable edit mode to show the workspace add control; newly added workspaces start with a single local terminal pane, become active immediately, and are saved to the config. In edit mode, each visible workspace tab can be renamed inline and has a delete button that asks for confirmation before removing the workspace. The last remaining workspace cannot be deleted. Switching the active workspace is persisted so the same workspace is restored after restart.
 
 ### Pane types
 
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -49,8 +49,8 @@ ssh_connections:
 # `workspaces:` format.
 #
 # The tab bar is hidden during normal use when only one workspace exists. Turn on
-# edit mode to add or delete workspaces. New workspaces start with one local
-# terminal pane. Deleting a workspace asks for confirmation, and the last
+# edit mode to add, rename, or delete workspaces. New workspaces start with one
+# local terminal pane. Deleting a workspace asks for confirmation, and the last
 # remaining workspace cannot be deleted. The active workspace selection is saved.
 #
 # Split node  — has `direction` and `children` (no `pane`)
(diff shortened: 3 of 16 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Full-stack TDD**: the rename path is tested at every layer — `config_test.go` (unit), `handler_test.go` (HTTP), `useLayout.test.ts` (hook), `WorkspaceTabs.test.tsx` (component), and `server_test.go` (integration routing). The test-case variety is solid: Enter save, Escape cancel, blank-name guard, not-found 404, invalid body 400, edit-mode-off 403, and persistence round-trip.
- **`startRename` guard** (`if (\!onRename) return`): double-clicking a tab in normal mode is silently ignored, so the UI stays consistent with edit-mode semantics.
- **`encodeURIComponent(workspaceID)`** in the fetch URL: correct handling of workspace IDs that might contain URL-special characters.
- **No-op optimisation in `commitRename`**: `if (title && title \!== workspace.title)` avoids a redundant API call when the user confirms without changing the name.
- **Defense-in-depth validation**: blank title is rejected at both the HTTP layer (422) and the config validation layer (`validate.go`), so the constraint is enforced even if the API is called directly.
- **Auto-focus + select in `useEffect`**: the input is focused and its text pre-selected as soon as the rename starts, which is the expected UX.

---

### Issues & Questions

**[Question] Escape → blur race may save despite cancel**

```tsx
// WorkspaceTabs.tsx
onBlur={() => commitRename(workspace)}
onKeyDown={(event) => {
  if (event.key === 'Enter') {
    commitRename(workspace)
  } else if (event.key === 'Escape') {
    cancelRename()   // sets editingWorkspaceId → null
  }
}}
```

When Escape is pressed, `cancelRename()` calls `setEditingWorkspaceId(null)` (queued). React re-renders and unmounts the input. Some browsers fire a native `blur` event during DOM removal; React may process it with the _old_ `onBlur` closure (from the pre-update render), where `editingWorkspaceId` still equals `workspace.id`. In that case `commitRename` passes its early-return check and calls `onRename` with the modified draft — i.e. the rename is saved despite Escape. The unit tests don't catch this because `fireEvent.keyDown` in jsdom does not synthesise a subsequent blur.

Standard mitigations: a `cancelledRef` that `onBlur` checks before committing, or calling `inputRef.current?.blur()` explicitly before calling `cancelRename` so blur fires _before_ the state update is queued. Did you verify this doesn't reproduce in Chrome / Firefox?

The same concern applies to the Enter path (blur may fire a second time after unmount), though there the worst case is a duplicate API call rather than a semantic reversal.

---

**[Minor question] Rename-on-blur happy path is not tested**

The `blur` trigger is exercised only for the blank-name guard:

```tsx
// WorkspaceTabs.test.tsx
it('does not save a blank inline rename', () => {
  ...
  fireEvent.blur(input)
  expect(onRename).not.toHaveBeenCalled()
})
```

A complementary `it('saves rename on blur with valid name', ...)` that fires `blur` with a non-empty value and asserts `onRename` was called would close this gap and also serve as a regression test for the blur-cancel concern above (by adding `toHaveBeenCalledTimes(1)`).

---

**[Minor question] Inconsistent error-response format in `PutWorkspace`**

```go
// internal/api/handler.go — 422 path
w.Header().Set("Content-Type", "application/json")
w.WriteHeader(http.StatusUnprocessableEntity)
_ = json.NewEncoder(w).Encode(map[string]string{"error": "workspace title must not be empty"})
// ...but 404 and 500 use plain-text http.Error
http.Error(w, "workspace not found", http.StatusNotFound)
http.Error(w, "failed to save workspaces", http.StatusInternalServerError)
```

The 422 case returns a JSON body while the other error paths in the same handler return plain text. The frontend ignores the body (`throw new Error(`HTTP ${response.status}`)`), so this is harmless today, but it may confuse callers. Is the JSON body intentional — e.g. to support richer validation messages in future?

---

### Overall

Clean, well-structured implementation that follows the established patterns for workspace operations. The one item worth confirming before merge is the Escape-then-blur interaction in a real browser; if it doesn't reproduce, this PR is ready to land.

tomo-chan:
Addressed the rename follow-up in 1eef242.

- Added a finalized ref so Escape followed by blur cannot commit a cancelled rename.
- The same guard prevents Enter followed by blur from issuing a duplicate rename.
- Added coverage for valid blur-save, Enter+blur single-save, and Escape+blur cancel behavior.

On the 422 response format question: I left the JSON validation body as-is because it matches the existing validation response pattern used elsewhere in the API. The frontend still ignores the body and handles it by status.

Local verification: make test-frontend, make lint-frontend. GitHub checks are now passing.

internal/api/handler.go
```diff
@@ -186,6 +190,36 @@ func (h *Handler) DeleteWorkspace(w http.ResponseWriter, r *http.Request) {
...
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, "invalid request body", http.StatusBadRequest)
+		return
+	}
+	title := strings.TrimSpace(req.Title)
+	if title == "" {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusUnprocessableEntity)
+		_ = json.NewEncoder(w).Encode(map[string]string{"error": "workspace title must not be empty"})
+		return
```
tomo-chan, line 210: The 422 branch returns a JSON body (`{"error": "..."}`) while the 404 and 500 paths below use plain-text `http.Error`. The frontend ignores the body, so this is harmless, but it may confuse future callers or tooling that inspects response bodies. If JSON is intentional here (e.g. for richer client-side messages later), a quick comment or unifying the other error paths would make the intent clear.

tomo-chan (commented):
## Code Review (follow-up)

Addressing the concerns raised in the previous review.

---

### Resolved

**Enter/Escape → blur race** — fixed by `renameFinalizedRef` in `WorkspaceTabs.tsx`. Setting the ref synchronously before `setEditingWorkspaceId(null)` means any stale `onBlur` closure that fires during unmount exits immediately via `if (renameFinalizedRef.current) return` — well before reaching `onRename`. Resetting the ref to `false` in `startRename` ensures it does not bleed across consecutive renames. The fix is correct and idiomatic.

**Missing blur tests** — three new tests close the gaps: `saves a valid inline rename on blur`, `does not save twice when Enter is followed by blur`, and `does not save when Escape is followed by blur`. The `toHaveBeenCalledTimes(1)` assertion in the Enter+blur test will reliably catch any regression to the double-call behaviour.

---

### Issues & Questions

**[Minor question] 422 error response uses JSON while other errors in the same handler use plain text**

This was raised in the previous review and the handler is unchanged:

```go
// internal/api/handler.go
if title == "" {
    w.Header().Set("Content-Type", "application/json")
    w.WriteHeader(http.StatusUnprocessableEntity)
    _ = json.NewEncoder(w).Encode(map[string]string{"error": "workspace title must not be empty"})
    return
}
// ...
http.Error(w, "workspace not found", http.StatusNotFound)      // plain text
http.Error(w, "failed to save workspaces", http.StatusInternalServerError) // plain text
```

The frontend ignores the body and only inspects the status code, so this is not a functional problem. Is the JSON envelope intentional — e.g. to pave the way for structured validation errors later?

---

### Overall

All blocking concerns are resolved. The `renameFinalizedRef` fix is clean and the new tests give good regression coverage. Ready to merge pending a response on the format question above.

tomo-chan:
Correction to the previous comment, whose inline code formatting was mangled by shell expansion:

The JSON body is intentional for validation failures. PutWorkspace follows the existing API pattern used by layout and pane validation paths, where 422 responses return a JSON error envelope so callers can display a structured validation message if they choose to. The 404 and 500 branches remain plain http.Error responses because they are lookup or server failures rather than request-validation feedback, and the current frontend only depends on the status code.

No code change needed for this follow-up. Current PR checks are passing.

tomo-chan (commented):
## Code Review (final)

All concerns from the previous two reviews are now resolved.

**[Resolved – commit `1eef242a`]** Enter/Escape → blur race: `renameFinalizedRef` set synchronously before `setEditingWorkspaceId(null)` prevents any stale `onBlur` closure from calling `onRename` after the rename is finalised or cancelled.

**[Resolved – commit `1eef242a`]** Missing blur tests: three new tests added — `saves a valid inline rename on blur`, `does not save twice when Enter is followed by blur`, and `does not save when Escape is followed by blur`.

**[Resolved – author reply]** 422 JSON response format: the JSON envelope is intentional, matching the existing API pattern for validation failures on layout and pane paths. The 404 and 500 branches remain plain-text `http.Error` because they represent lookup and server failures, not request-validation feedback.

---

### Overall

No open concerns. This PR is ready to merge.

PR #73, opened 2026-05-01
feat: allow workspace tab position changes

tomo-chan:
## Summary
- add a tab-position update API for workspace tabs
- add edit-mode controls to move workspace tabs top, bottom, left, or right
- preserve pending local layout changes when changing tab position
- update docs and examples

## Test plan
- [x] make check
- [x] GOCACHE=/tmp/panemux-go-build-cache make build-backend

## Notes
- Full make build was not used as the final verification because its repeated test phase hit sandbox-only environment constraints after make check had already passed; backend build was verified separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ workspaces:
 
 Older config files with a top-level `layout:` are still accepted. When the config is next saved, panemux migrates that layout into a `default` workspace and writes the `workspaces:` format.
 
-If there is only one workspace, the workspace tab bar is hidden during normal use. Enable edit mode to show the workspace add control; newly added workspaces start with a single local terminal pane, become active immediately, and are saved to the config. In edit mode, each visible workspace tab can be renamed inline and has a delete button that asks for confirmation before removing the workspace. The last remaining workspace cannot be deleted. Switching the active workspace is persisted so the same workspace is restored after restart.
+If there is only one workspace, the workspace tab bar is hidden during normal use. Enable edit mode to show the workspace add and tab-position controls; newly added workspaces start with a single local terminal pane, become active immediately, and are saved to the config. In edit mode, each visible workspace tab can be renamed inline and has a delete button that asks for confirmation before removing the workspace. The last remaining workspace cannot be deleted. Switching the active workspace and changing `tab_position` are persisted so the same workspace and tab placement are restored after restart.
 
 ### Pane types
 
@@ -150,7 +150,7 @@ Authentication is attempted in order: configured `key_file` → configured `pass
 
 Click the lock icon in the bottom-right corner to toggle **edit mode**.
 
-- **ON** — layout changes are persisted to the config file; terminal input is blocked and a visual overlay shows which panes are locked; drag any pane by its body (not just the header) to reorder it
+- **ON** — layout changes and workspace tab placement are persisted to the config file; terminal input is blocked and a visual overlay shows which panes are locked; drag any pane by its body (not just the header) to reorder it
 - **OFF** — terminal is fully interactive; drag-resize and close are applied in-memory only
 
 ---
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -49,9 +49,10 @@ ssh_connections:
 # `workspaces:` format.
 #
 # The tab bar is hidden during normal use when only one workspace exists. Turn on
-# edit mode to add, rename, or delete workspaces. New workspaces start with one
-# local terminal pane. Deleting a workspace asks for confirmation, and the last
-# remaining workspace cannot be deleted. The active workspace selection is saved.
+# edit mode to add, rename, delete, or move workspace tabs. New workspaces start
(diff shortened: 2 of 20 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Schema-first discipline**: `WorkspaceTabPositionRequestSchema` is added in `schemas/index.ts` before anything else; `TabPosition` is reused rather than duplicated; types are derived via `z.infer<>`. This follows the project's schema-first rule exactly.

- **`ValidateWorkspaceTabPosition` extracted correctly**: The validation logic is factored into a single exported function used by both `validateWorkspaces` and `SetWorkspaceTabPosition`. The switch-on-constants pattern is clean and eliminates the magic strings that were scattered across the old `if` chain.

- **Pending layout preservation**: `setWorkspaces((current) => current ? { ...current, tab_position: parsed.tab_position } : parsed)` adopts the server's new `tab_position` while keeping any in-flight local layout changes in `items`. This is a thoughtful design and is explicitly tested with "preserves a pending local layout change when updating workspace tab position".

- **Route-conflict test**: `TestServer_WorkspaceTabPositionRouteWiredBeforeWorkspaceIDRoute` confirms that `/workspaces/tab-position` resolves before the parameterized `/workspaces/{id}` — a low-overhead check that catches a real class of routing bugs.

- **Accessibility**: The position control group uses `role="group"`, per-button `aria-label`, and `aria-pressed` for the current selection — screen-reader-friendly without any extra library dependencies.

- **Comprehensive test matrix**: handler tests cover all four valid positions (table-driven), 403, 400, 422, 500, and on-disk persistence; hook tests cover happy path, pending-layout preservation, the pre-load edge case, and error state.

---

### Issues & Questions

**[Minor question] Selective merge silently discards future top-level server response fields**

In `frontend/src/hooks/useLayout.ts`:

```ts
setWorkspaces((current) => current ? { ...current, tab_position: parsed.tab_position } : parsed)
```

When `current` is non-null only `tab_position` is taken from `parsed`; all other top-level fields (`active`, `items`, any future additions) are kept from the stale local state. Since `PutWorkspaceTabPosition` only changes `tab_position` today, this is correct — but if the endpoint is ever extended (e.g. re-ordering workspace order as a side effect), the merge strategy would silently discard those server-side changes.

Would it be safer to invert the merge — take everything from the server response except the locally-modified `items`?

```ts
setWorkspaces((current) => current
  ? { ...parsed, items: current.items }
  : parsed)
```

That way new top-level fields from the server are always adopted while the pending local layout is still preserved.

**[Minor question] 422 uses a JSON body; 403 and 500 use `http.Error` plain text**

In `internal/api/handler.go` (`PutWorkspaceTabPosition`):

```go
// 403
http.Error(w, "edit mode required", http.StatusForbidden)
// 422
w.Header().Set("Content-Type", "application/json")
w.WriteHeader(http.StatusUnprocessableEntity)
_ = json.NewEncoder(w).Encode(map[string]string{"error": err.Error()})
// 500
http.Error(w, "failed to save workspaces", http.StatusInternalServerError)
```

The frontend error handler only reads the status code (`throw new Error(\`HTTP ${response.status}\``)), so this inconsistency is invisible to users today. Is the JSON body for 422 intentional (matching how other validation endpoints in the project return structured errors), or is it incidental?

---

### Overall

Clean, well-structured PR with thorough test coverage across all layers. The two questions above are non-blocking — the pending-layout preservation logic and the routing-conflict test are especially nice touches. Ready to merge once the questions are acknowledged or addressed.

tomo-chan:
Follow-up addressed in commit 2cb7c22.

- Selective merge question: changed setWorkspaceTabPosition to adopt the parsed server response top-level fields while preserving the current local items, so pending layout edits are still kept but future top-level response fields are not silently discarded. Added a hook regression test for that merge behavior.
- 422 response-shape question: the JSON body is intentional and follows the existing API pattern for validation failures with status 422, while 403 and 500 continue to use the repository existing http.Error pattern. The frontend still only depends on the status code.

Verification run after the follow-up change:
- npm test -- --run src/hooks/useLayout.test.ts
- make check

tomo-chan (commented):
## Code Review (follow-up on commit 2cb7c22)

Both questions from the previous review are resolved.

**Selective merge — resolved**

The merge strategy in `frontend/src/hooks/useLayout.ts` is now:

```ts
setWorkspaces((current) => current ? { ...parsed, items: current.items } : parsed)
```

This correctly adopts all top-level fields from the server response (`active`, `tab_position`, and any future additions) while preserving only the locally-modified `items`. The new test "adopts server top-level workspace fields while preserving current items when updating tab position" exercises exactly this: the server's `active: 'ops'` and `tab_position: 'right'` are adopted, but `items[0].title` stays as the local `'Dev'` rather than the server's `'Server Dev'`. 

**422 JSON body — acknowledged**

The intentional distinction (JSON body for 422, `http.Error` for 403/500) is now documented in the author reply. No change needed.

### Overall

Both concerns addressed cleanly. Ready to merge.

PR #74, opened 2026-05-01
feat: add agent attention notifications

tomo-chan:
## Summary
- Detect conservative agent confirmation prompts from terminal output, including ANSI-stripped split frames and streamed binary output.
- Flash panes and inactive workspace tabs when attention is needed, and request browser OS notifications when supported.
- Clear pane/workspace attention on focus, click, or workspace selection, and document the frontend-only detection limits.

## Test Plan
- [x] `git diff --check`
- [x] `make lint-frontend`
- [x] `make test-frontend`

## Notes
- `GOCACHE=/tmp/panemux-gocache-20260501120150 GOLANGCI_LINT_CACHE=/tmp/panemux-golangci-cache-20260501120150 make check` was attempted. It passed frontend build, `go vet`, `golangci-lint`, frontend typecheck, and most Go packages, then failed on sandbox-bound local environment constraints: `dscl: exit status 70` in `TestDetectLocalShell_ReturnsAbsolutePath` and `httptest` cannot bind `[::1]:0` in `internal/ws`.

### DIFF
--- a/frontend/src/hooks/useWebSocket.ts
+++ b/frontend/src/hooks/useWebSocket.ts
@@ -104,5 +104,6 @@ export function useWebSocket(url: string, options: UseWebSocketOptions) {
 }
 
 function isArrayBuffer(value: unknown): value is ArrayBuffer {
+  // Vitest can deliver ArrayBuffer values from a different JS realm.
   return value instanceof ArrayBuffer || Object.prototype.toString.call(value) === '[object ArrayBuffer]'
 }
--- a/frontend/src/hooks/useWebSocket.ts
+++ b/frontend/src/hooks/useWebSocket.ts
@@ -43,7 +43,7 @@ export function useWebSocket(url: string, options: UseWebSocketOptions) {
     }
 
     ws.onmessage = (event) => {
-      const isBinary = event.data instanceof ArrayBuffer
+      const isBinary = isArrayBuffer(event.data)
       if (!isBinary) {
         // Validate text frames before passing to handler
         try {
@@ -101,4 +101,8 @@ export function useWebSocket(url: string, options: UseWebSocketOptions) {
   }, [connect])
 
   return { send, connected, reconnect }
+}
+
+function isArrayBuffer(value: unknown): value is ArrayBuffer {
+  return value instanceof ArrayBuffer || Object.prototype.toString.call(value) === '[object ArrayBuffer]'
 }
--- a/frontend/src/utils/agentAttention.ts
+++ b/frontend/src/utils/agentAttention.ts
@@ -2,7 +2,6 @@ const MAX_TAIL_LENGTH = 1200
 
 const ATTENTION_PATTERNS = [
   /\b(codex|claude|agent)\b.{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i,
-  /\b(confirm|approve|allow|proceed)\b.{0,120}\?/i,
   /(確認|承認|許可).{0,120}(必要|してください|しますか|待ち)/,
 ]
 
--- a/frontend/src/utils/agentAttention.ts
+++ b/frontend/src/utils/agentAttention.ts
@@ -1,9 +1,10 @@
 const MAX_TAIL_LENGTH = 1200
 
 const ATTENTION_PATTERNS = [
-  /\b(codex|claude|agent)\b.{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i,
-  /\bYes,\s*proceed\s*\(y\)[\s\S]{0,160}\bNo,\s*and\s*tell\s+Codex\s+what\s+to\s+do\s+differently\s*\(esc\)/i,
-  /(確認|承認|許可).{0,120}(必要|してください|しますか|待ち)/,
+  /\b(codex|claude|agent)\b[\s\S]{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i,
+  /\bYes,\s*proceed\s*\(y\)[\s\S]{0,160}\bNo,\s*and\s*tell\s+(Codex|Claude|agent)\s+what\s+to\s+do\s+differently\s*\(esc\)/i,
+  /Do you want to proceed\?[\s\S]{0,120}\b1\.\s*Yes\b[\s\S]{0,80}\b2\.\s*No\b[\s\S]{0,160}Esc to cancel\s*·\s*Tab to amend\s*·\s*ctrl\+e to explain/i,
+  /(確認|承認|許可)[\s\S]{0,120}(必要|してください|しますか|待ち)/,
 ]
 
 export interface AgentAttentionDetector {
--- a/frontend/src/utils/agentAttention.ts
+++ b/frontend/src/utils/agentAttention.ts
@@ -2,6 +2,7 @@ const MAX_TAIL_LENGTH = 1200
 
 const ATTENTION_PATTERNS = [
   /\b(codex|claude|agent)\b.{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i,
+  /\bYes,\s*proceed\s*\(y\)[\s\S]{0,160}\bNo,\s*and\s*tell\s+Codex\s+what\s+to\s+do\s+differently\s*\(esc\)/i,
   /(確認|承認|許可).{0,120}(必要|してください|しますか|待ち)/,
 ]
 
--- a/frontend/src/utils/agentAttention.ts
+++ b/frontend/src/utils/agentAttention.ts
@@ -1,0 +1,34 @@
+const MAX_TAIL_LENGTH = 1200
+
+const ATTENTION_PATTERNS = [
+  /\b(codex|claude|agent)\b.{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i,
+  /\b(confirm|approve|allow|proceed)\b.{0,120}\?/i,
+  /(確認|承認|許可).{0,120}(必要|してください|しますか|待ち)/,
+]
+
+export interface AgentAttentionDetector {
+  feed: (chunk: string) => boolean
+  reset: () => void
+}
+
+export function createAgentAttentionDetector(): AgentAttentionDetector {
+  let tail = ''
+
+  return {
+    feed(chunk: string) {
+      tail = (tail + stripAnsi(chunk)).slice(-MAX_TAIL_LENGTH)
+      const matched = ATTENTION_PATTERNS.some((pattern) => pattern.test(tail))
+      if (matched) {
+        tail = ''
+      }
+      return matched
+    },
+    reset() {
+      tail = ''
+    },
+  }
+}
+
+function stripAnsi(value: string): string {
+  return value.replace(/\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~]|\][^\x07]*(?:\x07|\x1B\\))/g, '')
+}
--- a/frontend/src/utils/layoutTree.ts
+++ b/frontend/src/utils/layoutTree.ts
@@ -108,6 +108,23 @@ export function findPaneById(layout: LayoutNode, paneId: string): PaneConfig | n
   return null
 }
 
+export function layoutContainsPane(layout: LayoutNode, paneId: string): boolean {
+  return findPaneById(layout, paneId) !== null
+}
+
+export function listPaneIds(layout: LayoutNode): string[] {
+  const paneIds: string[] = []
+  collectPaneIds(layout.children, paneIds)
+  return paneIds
+}
+
+function collectPaneIds(children: LayoutChild[], paneIds: string[]) {
+  for (const child of children) {
+    if (child.pane) paneIds.push(child.pane.id)
+    if (child.children?.length) collectPaneIds(child.children, paneIds)
+  }
+}
+
 /**
  * Replaces the pane matching `updated.id` in the layout tree with `updated`.
  * Returns the tree unchanged if the id is not found.
--- a/frontend/src/utils/layoutTree.ts
+++ b/frontend/src/utils/layoutTree.ts
@@ -112,19 +112,6 @@ export function layoutContainsPane(layout: LayoutNode, paneId: string): boolean 
   return findPaneById(layout, paneId) !== null
 }
 
-export function listPaneIds(layout: LayoutNode): string[] {
-  const paneIds: string[] = []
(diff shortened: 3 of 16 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- ** design** — Clean factory function with a minimal `feed`/`reset` interface. Isolating detection logic in `agentAttention.ts` makes it independently testable and easy to swap patterns later.
- **`TextDecoder({ stream: true })` for binary frames** — Correctly handles multi-byte UTF-8 sequences (e.g. Japanese text) that arrive split across WebSocket frames. Without the `stream` flag, a cut mid-sequence produces U+FFFD and would miss the match.
- **State identity optimization in `clearWorkspaceAttention`** — Returns the original `current` Set unchanged when nothing was removed, preventing unnecessary re-renders.
- **`ReadonlySet<string>` prop type** — The `attentionWorkspaceIds?: ReadonlySet<string>` signature correctly signals to callers that `WorkspaceTabs` will not mutate the set.
- **Strong test coverage** — Tests cover split-frame detection, ANSI stripping, 10 s throttle deduplication, multi-pane workspace attention state machine, and the inactive-pane non-mount assertion. The `useContext` approach in the App test to probe internal state without adding extra callback props is a nice pattern.
- **Reset on reconnect** — Resetting `lastAttentionAtRef`, `attentionDetectorRef`, and `outputDecoderRef` in `onOpen` correctly clears stale state from the previous connection.

---

### Issues & Questions

**[Question] Pattern 2 will also fire on non-agent interactive CLI prompts**

`/\b(confirm|approve|allow|proceed)\b.{0,120}\?/i` matches any terminal question containing one of those words, including common non-agent prompts like `npm`'s "Proceed? [y/N]", Homebrew's "Allow changes?", or a `rm` confirmation dialog. Pattern 1 already requires an agent name anchor (codex/claude/agent) and is suitably specific; pattern 2 has no such anchor.

Is this intentional? If the design goal is to fire on *all* confirmation-style terminal prompts (not just AI agent ones), that's fine, but it will trigger frequently during normal terminal use. If the intent is agent-scoped, should pattern 2 also require an agent-name prefix, or at least the docs/code should note that it matches general interactive prompts?

**[Minor question] `listPaneIds` is exported but not imported anywhere in this PR**

`listPaneIds` and `collectPaneIds` were added to `layoutTree.ts` but no import of `listPaneIds` appears in this diff. Is this dead code that should be removed, or scaffolding for a planned follow-up?

**[Minor question] `isArrayBuffer` cross-realm guard — context?**

```typescript
return value instanceof ArrayBuffer || Object.prototype.toString.call(value) === '[object ArrayBuffer]'
```

The second clause handles cross-realm `ArrayBuffer` instances (e.g. from a different iframe). In standard browser WebSocket use `event.data` is always the host realm's `ArrayBuffer`, so `instanceof` alone is sufficient. Was this triggered by a specific test/browser failure, or is it a precautionary addition? If precautionary, a short inline comment would help future readers understand the intent.

---

### Overall

Well-structured feature: detection logic is isolated, streaming decoding is correct, and the state model cleanly propagates attention from pane to workspace tab. The main item worth clarifying before merge is whether pattern 2 intentionally casts a wider net over all confirmation-style terminal prompts or should be scoped to AI agent prompts like pattern 1.

frontend/src/utils/agentAttention.ts
```diff
@@ -0,0 +1,34 @@
+const MAX_TAIL_LENGTH = 1200
+
+const ATTENTION_PATTERNS = [
+  /\b(codex|claude|agent)\b.{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i,
+  /\b(confirm|approve|allow|proceed)\b.{0,120}\?/i,
```
tomo-chan: This pattern fires on any terminal question containing "confirm/approve/allow/proceed", not just agent prompts — e.g. `npm`'s "Proceed? [y/N]" or Homebrew's "Allow changes?". Pattern 1 anchors on the agent name and is specific; this one has no such constraint. Intentional broad coverage, or should it also require an agent-name anchor?

frontend/src/utils/layoutTree.ts
```diff
@@ -108,6 +108,23 @@ export function findPaneById(layout: LayoutNode, paneId: string): PaneConfig | n
   return null
 }
 
+export function layoutContainsPane(layout: LayoutNode, paneId: string): boolean {
+  return findPaneById(layout, paneId) !== null
+}
+
+export function listPaneIds(layout: LayoutNode): string[] {
```
tomo-chan: `listPaneIds` (and `collectPaneIds`) are added here but not imported anywhere in this diff. Dead code to remove, or scaffolding for a follow-up?

frontend/src/hooks/useWebSocket.ts
```diff
@@ -43,7 +43,7 @@ export function useWebSocket(url: string, options: UseWebSocketOptions) {
     }
 
     ws.onmessage = (event) => {
-      const isBinary = event.data instanceof ArrayBuffer
+      const isBinary = isArrayBuffer(event.data)
```
tomo-chan, line 46: The `Object.prototype.toString` fallback handles cross-realm `ArrayBuffer` instances. In normal browser WebSocket use `instanceof ArrayBuffer` is sufficient. Was this triggered by a specific failure, or is it precautionary? A short comment here would clarify the intent for future readers.

tomo-chan:
Follow-up pushed.

Commits:
- https://github.com/tomo-chan/panemux/commit/1d28a31562093fe7ca536cc0e67a3236d0166311 addressed the original review feedback by removing the broad generic confirmation regex, adding generic-prompt regression coverage, removing unused pane ID helpers, and documenting the cross-realm `ArrayBuffer` guard.
- https://github.com/tomo-chan/panemux/commit/aa20446cc6b549ee500ef9fbb2a0a7d81f2b35d5 added support for Codex approval menus with `1. Yes, proceed (y)` and `2. No, and tell Codex what to do differently (esc)`.
- https://github.com/tomo-chan/panemux/commit/8f805982645985d3b4e823bb97672922a703427e added regression coverage for the three-option Codex edit approval menu.
- https://github.com/tomo-chan/panemux/commit/390facbaf7e32f5ddaf0de754eaa2e2521dc1de3 added regression coverage for Codex command approval menus with scoped do-not-ask-again options such as `commands that start with gh pr comment`.
- https://github.com/tomo-chan/panemux/commit/75b1dd3e9cd0d38551cc24fe0127a69514f63005 added multiline English/Japanese prompt detection, Claude approval-menu support, and Claude Bash command approval UI coverage.

Verification run locally:
- `npm test -- --run src/utils/agentAttention.test.ts`
- `make check`

frontend/src/utils/agentAttention.ts
tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/1d28a31562093fe7ca536cc0e67a3236d0166311 by removing the broad generic confirmation-question pattern and adding regression coverage that `Proceed? [y/N]` / generic `Allow...` prompts do not notify.

The later Codex-specific menu examples are covered separately without reopening the broad match:
- https://github.com/tomo-chan/panemux/commit/aa20446cc6b549ee500ef9fbb2a0a7d81f2b35d5 detects the two-option Codex approval menu.
- https://github.com/tomo-chan/panemux/commit/8f805982645985d3b4e823bb97672922a703427e adds regression coverage for the three-option edit approval menu.

frontend/src/hooks/useWebSocket.ts
tomo-chan: Clarified in https://github.com/tomo-chan/panemux/commit/1d28a31562093fe7ca536cc0e67a3236d0166311.

This was precautionary rather than from a specific browser failure. I kept the cross-realm guard and added an inline comment explaining that the fallback accepts `ArrayBuffer` values from alternate realms while preserving the normal `instanceof` path.

frontend/src/utils/layoutTree.ts
tomo-chan: Removed in https://github.com/tomo-chan/panemux/commit/1d28a31562093fe7ca536cc0e67a3236d0166311.

`listPaneIds` / `collectPaneIds` were unused scaffolding, so I dropped them rather than keeping dead exported code in `layoutTree.ts`.

frontend/src/utils/agentAttention.ts
```diff
@@ -0,0 +1,34 @@
+const MAX_TAIL_LENGTH = 1200
+
+const ATTENTION_PATTERNS = [
+  /\b(codex|claude|agent)\b.{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i,
```
tomo-chan: Pattern 1 uses `.` which does not match `\n` in JS. A prompt like `Codex needs your\napproval — proceed?` would not be detected even though both tokens are within 160 chars of each other in `tail`. Pattern 2 uses `[\s\S]` to span newlines. Intentional single-line constraint, or should this use `[\s\S]{0,160}` too?

frontend/src/utils/agentAttention.ts
```diff
@@ -0,0 +1,34 @@
...
+  /\bYes,\s*proceed\s*\(y\)[\s\S]{0,160}\bNo,\s*and\s*tell\s+Codex\s+what\s+to\s+do\s+differently\s*\(esc\)/i,
```
tomo-chan: This pattern hard-codes `tell\s+Codex`. A Claude Code menu that reads `tell Claude what to do differently (esc)` would miss this path (though Claude text prompts are still caught by pattern 1). Intentional Codex-only scoping for now, or worth making the agent name in the menu text a variable part of the pattern?

tomo-chan (commented):
## Code Review (follow-up round)

### Previous findings — all resolved

- **Pattern 2 false positives** — Replaced with a Codex-specific menu pattern (`/\bYes,\s*proceed\s*\(y\)[\s\S]{0,160}\bNo,\s*and\s*tell\s+Codex…/i`). The regression tests for `npm` and Homebrew prompts confirm the generic triggers are gone.
- **`listPaneIds` dead code** — Removed cleanly in `1d28a315`.
- **`isArrayBuffer` guard** — Inline comment added (`// Vitest can deliver ArrayBuffer values from a different JS realm.`), which is exactly the context that was missing.

---

### Issues & Questions

**[Minor question] Pattern 1 uses `.` (dot, no newline) — multi-line prompts would not be detected**

```typescript
/\b(codex|claude|agent)\b.{0,160}\b(confirm|confirmation|approve|approval|permission|allow|proceed)\b/i
```

In JavaScript, `.` does not match `\n` by default. If an agent writes its confirmation text across two terminal lines (e.g. `Codex is waiting for your\napproval — proceed?`), the pattern would fail to match even though both tokens are within 160 characters in the `tail` buffer. Pattern 2 correctly uses `[\s\S]{0,160}` to span newlines. Should pattern 1 (and the Japanese pattern 3) do the same, or is single-line-only detection an intentional constraint?

**[Minor question] Pattern 2 is Codex-branded — a Claude Code approval menu would not match**

```typescript
/\bYes,\s*proceed\s*\(y\)[\s\S]{0,160}\bNo,\s*and\s*tell\s+Codex\s+what\s+to\s+do\s+differently\s*\(esc\)/i
```

The pattern hard-codes `tell\s+Codex`. If Claude Code (or another agent) shows a similar numbered approval menu but uses `tell Claude what to do differently`, it would not match. Pattern 1 already covers Claude's text-based prompts via the agent-name anchor, so this only affects the menu-format path. Is this intentional scoping to Codex menus for now, or worth extending to handle other agent menu formats?

---

### Overall

All three original concerns are cleanly resolved and the new Codex-specific menu pattern is well-tested against realistic fixtures. The two minor questions above are low-risk and do not block merge — they are worth a quick decision before this pattern set grows.

frontend/src/utils/agentAttention.ts
tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/75b1dd3e9cd0d38551cc24fe0127a69514f63005.

The numbered approval-menu pattern now accepts `Codex`, `Claude`, or `agent` in the "tell ... what to do differently" text. I also added coverage for the Claude Bash command approval UI (`Do you want to proceed?` with `1. Yes` / `2. No` and the `Esc to cancel · Tab to amend · ctrl+e to explain` footer) while keeping the generic non-agent prompt regressions in place.

frontend/src/utils/agentAttention.ts
tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/75b1dd3e9cd0d38551cc24fe0127a69514f63005.

Pattern 1 now uses `[\s\S]{0,160}` so agent-name anchored prompts can span terminal line breaks. I also changed the Japanese approval pattern to use `[\s\S]{0,120}` and added regression coverage for both multiline English and Japanese prompts.

tomo-chan (commented):
## Code Review (round 3)

### Previous findings — all resolved

**From round 2:**

- **Pattern 1 dot-vs-newline** — Addressed in `75b1dd31`: `.{0,160}` → `[\s\S]{0,160}` across all three non-menu patterns. Two new tests confirm multi-line English and Japanese prompts are now detected (`'Codex is waiting for your\napproval'`, `'操作の承認\nが必要です'`).
- **Pattern 2 Codex-only branding** — Addressed in `75b1dd31`: `tell\s+Codex` expanded to `tell\s+(Codex|Claude|agent)`. A new test covers the Claude variant (`'No, and tell Claude what to do differently (esc)'`).

Additionally, a new pattern 3 was added for Claude's Bash command approval UI:

```typescript
/Do you want to proceed\?[\s\S]{0,120}\b1\.\s*Yes\b[\s\S]{0,80}\b2\.\s*No\b[\s\S]{0,160}Esc to cancel\s*·\s*Tab to amend\s*·\s*ctrl\+e to explain/i
```

The pattern is tightly anchored on Claude's specific UI footer (`Esc to cancel · Tab to amend · ctrl+e to explain`), making false positives on generic "Do you want to proceed?" prompts essentially impossible. The test fixture uses real terminal output including Unicode arrow glyphs (`❯`), which is a good robustness check for the word-boundary handling.

---

### Issues & Questions

None. All five concerns raised across the two prior review rounds have been addressed.

---

### Overall

The pattern set is now solid: all three classes of confirmation output (agent-name-anchored text, numbered approval menus, Claude Bash UI) are covered, multiline matching is consistent across patterns, and the test suite includes realistic fixtures with split frames, ANSI codes, Unicode glyphs, and regression cases for the prompts that should *not* fire. Ready to merge.

PR #75, opened 2026-05-02
fix: repaint terminal after workspace switches

tomo-chan:
## Summary

Fixes terminal text disappearing after switching away from a workspace and returning by forcing xterm to repaint when the terminal is remounted, when the browser window regains focus, and when the document becomes visible again.

## Root cause

The terminal instance can be reused after a workspace switch while its canvas/viewport has not repainted yet. A later layout resize triggered xterm fit/refresh, which made the existing buffer visible again.

## Changes

- Schedule immediate and short delayed terminal fit/refresh passes after remount.
- Repaint each mounted, attached terminal pane on `window.focus` and visible `visibilitychange`.
- Clear pending repaint timers on detach, disposal, and test reset.
- Add regression coverage for remount, focus, visible document, and hidden document behavior.

## Test plan

- [x] `cd frontend && npm run test -- src/hooks/useTerminal.test.ts`
- [x] `make check`

## Residual risk

Automated coverage exercises the repaint triggers, but this was not manually verified in a real browser workspace-switch session.

### DIFF
--- a/frontend/src/hooks/useTerminal.test.ts
+++ b/frontend/src/hooks/useTerminal.test.ts
@@ -486,6 +486,95 @@ describe('useTerminal', () => {
     expect(secondContainer.contains(mockTerm.element as HTMLElement)).toBe(true)
   })
 
+  it('repaints a reused terminal again after remount timing settles', () => {
+    vi.useFakeTimers()
+    window.requestAnimationFrame = ((cb: FrameRequestCallback) => {
+      cb(0)
+      return 0
+    }) as typeof window.requestAnimationFrame
+    const firstContainer = makeContainer()
+    const secondContainer = makeContainer()
+
+    const first = renderHook(() => useTerminal({ sessionId: 's1', container: firstContainer }))
+    first.unmount()
+    mockFitAddon.fit.mockClear()
+    mockTerm.refresh.mockClear()
+
+    renderHook(() => useTerminal({ sessionId: 's1', container: secondContainer }))
+
+    expect(mockFitAddon.fit).toHaveBeenCalledTimes(1)
+    expect(mockTerm.refresh).toHaveBeenCalledTimes(1)
+
+    act(() => vi.advanceTimersByTime(50))
+
+    expect(mockFitAddon.fit).toHaveBeenCalledTimes(2)
+    expect(mockTerm.refresh).toHaveBeenCalledTimes(2)
+    vi.useRealTimers()
+  })
+
+  it('repaints the attached terminal when the window regains focus', () => {
+    vi.useFakeTimers()
+    window.requestAnimationFrame = ((cb: FrameRequestCallback) => {
+      cb(0)
+      return 0
+    }) as typeof window.requestAnimationFrame
+    const container = makeContainer()
+
+    renderHook(() => useTerminal({ sessionId: 's1', container }))
+    mockFitAddon.fit.mockClear()
+    mockTerm.refresh.mockClear()
+
+    act(() => window.dispatchEvent(new Event('focus')))
+
+    expect(mockFitAddon.fit).toHaveBeenCalledTimes(1)
+    expect(mockTerm.refresh).toHaveBeenCalledTimes(1)
+
+    act(() => vi.advanceTimersByTime(50))
+
+    expect(mockFitAddon.fit).toHaveBeenCalledTimes(2)
+    expect(mockTerm.refresh).toHaveBeenCalledTimes(2)
+    vi.useRealTimers()
+  })
+
+  it('repaints the attached terminal when the document becomes visible', () => {
+    vi.useFakeTimers()
+    window.requestAnimationFrame = ((cb: FrameRequestCallback) => {
+      cb(0)
+      return 0
+    }) as typeof window.requestAnimationFrame
+    vi.spyOn(document, 'visibilityState', 'get').mockReturnValue('visible')
+    const container = makeContainer()
+
+    renderHook(() => useTerminal({ sessionId: 's1', container }))
+    mockFitAddon.fit.mockClear()
+    mockTerm.refresh.mockClear()
+
+    act(() => document.dispatchEvent(new Event('visibilitychange')))
+
+    expect(mockFitAddon.fit).toHaveBeenCalledTimes(1)
+    expect(mockTerm.refresh).toHaveBeenCalledTimes(1)
+
+    act(() => vi.advanceTimersByTime(50))
+
+    expect(mockFitAddon.fit).toHaveBeenCalledTimes(2)
+    expect(mockTerm.refresh).toHaveBeenCalledTimes(2)
+    vi.useRealTimers()
+  })
+
+  it('does not repaint the attached terminal when the document becomes hidden', () => {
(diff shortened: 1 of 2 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Clean function decomposition**: splitting the old monolithic `refreshTerminal` into `refreshTerminal` (schedules the pipeline) and `repaintTerminal` (executes one pass) makes each function's responsibility immediately clear.
- **`\!entry.attachedContainer` guard in `repaintTerminal`**: prevents stale scheduled timers from painting after the pane has been detached — the right place for that guard.
- **`entry.term.rows > 0` guard**: silently fixes a latent bug where `refresh(0, rows - 1)` could have been called as `refresh(0, -1)` on an uninitialised terminal.
- **No timer leaks**: `clearScheduledRepaints` is called in both the detach path and the disposal timeout, and `__resetTerminalEntriesForTests` was updated to clear the new timers — test isolation is preserved.
- **Four-scenario test coverage**: remount, focus, visible `visibilitychange`, and hidden `visibilitychange` are all covered.

---

### Issues & Questions

**[Minor question] `REMOUNT_REPAINT_DELAYS_MS` name scope** `useTerminal.ts` line 11

The constant is named with a `REMOUNT_` prefix, but it drives the deferred-repaint schedule in the `window.focus` and `visibilitychange` handlers as well. A name like `REPAINT_SETTLE_DELAYS_MS` would describe the purpose (compositor/layout settle time) without implying remount-only scope. Worth renaming?

**[Question] All panes repaint on focus — intentional?** `useTerminal.ts` new `useEffect`

Each `useTerminal` instance registers its own `window.focus` and `document.visibilitychange` handlers. In a typical panemux layout with N panes, all N attached panes will call `refreshTerminal` on every focus event, scheduling N × (1 + len(REMOUNT_REPAINT_DELAYS_MS)) `setTimeout` calls. The `\!entry.attachedContainer` guard keeps idle entries safe, so the behaviour itself looks correct. Wanted to confirm this is intentional — repaint *all* visible panes on focus, not just the most-recently-active one.

**[Minor question] 250 ms timer never asserted in tests** `useTerminal.test.ts` remount test

The tests advance to 50 ms and verify 2 fit/refresh calls, which covers the first entry in `REMOUNT_REPAINT_DELAYS_MS`. The second timer (250 ms) is scheduled but never asserted in any test. If the goal is belt-and-suspenders resilience the omission is fine, but it would be good to either add an assertion at `vi.advanceTimersByTime(250)` or add a comment explaining why the second pass is intentionally left unverified.

---

### Overall

Solid fix with good defensive guards and test coverage for all four trigger scenarios. The three items above are minor — the naming nit and the 250 ms test gap are the only things worth a follow-up comment before merge. No blockers.

tomo-chan:
Follow-up on the review notes in `84190a9`:

- Renamed `REMOUNT_REPAINT_DELAYS_MS` to `REPAINT_SETTLE_DELAYS_MS` so the name matches its use for remount, focus, and visibility-triggered repaints.
- Extended the repaint tests to assert the delayed 250 ms pass for remount, focus, and visible `visibilitychange` flows.
- The all-panes-on-focus behavior is intentional: a workspace switch or browser refocus can leave any attached pane with a stale canvas, so each mounted pane repaints its own attached terminal when the page returns.

Local verification:

- `make test-frontend`
- `make lint-frontend`
- `make check`

tomo-chan (commented):
## Code Review

### What's Good

- **Prior naming concern resolved**: `REPAINT_SETTLE_DELAYS_MS` now accurately describes all use sites (remount, focus, and visibility triggers), not just remount — a clear improvement over `REMOUNT_REPAINT_DELAYS_MS`.
- **250 ms pass now fully asserted**: all three repaint-trigger tests (remount, focus, visible `visibilitychange`) advance timers through both 50 ms and 250 ms and assert the expected call counts at each milestone. The hidden-document test correctly expects zero calls.
- **`\!entry.attachedContainer` guard in `repaintTerminal`**: stale scheduled callbacks cannot fire after the pane has been detached — the right place for that guard.
- **Clean `clearScheduledRepaints` plumbing**: called in both the detach path and the disposal timeout, and `__resetTerminalEntriesForTests` was updated to drain the new timers — test isolation is preserved.

---

### Issues & Questions

**[Question] All-pane repaint on focus — implicit confirmation needed** (`useTerminal.ts`, new `useEffect`)

The prior review asked whether it is intentional that every `useTerminal` instance registers its own `window.focus` and `document.visibilitychange` handlers, meaning all N attached panes repaint on every focus/visibility event. The behaviour is unchanged in this follow-up commit. The PR description says "Repaint the active terminal on `window.focus`..." — "active" is ambiguous: does it mean all panes in the currently visible workspace, or only the last-focused one? Could you confirm the design intent: repaint *all* mounted, attached panes on focus (guarded by `\!entry.attachedContainer`), not just the most-recently-active one?

**[Minor question] `setDims` is now conditionally skipped in the resize effect** (`useTerminal.ts`, resize `useEffect`)

```diff
-    entry.fitAddon.fit()
-    entry.term.refresh(0, entry.term.rows - 1)
+    repaintTerminal(entry, setDims)
     const { cols, rows } = entry.term
-    setDims({ cols, rows })
```

Because `repaintTerminal` returns early when `\!entry.attachedContainer`, `setDims` is now skipped if a resize event fires while the pane is detaching. Previously `setDims` was called unconditionally. Skipping it when there is no attached container is arguably correct (dimensions would be stale anyway), but this is a subtle behavioral change. Just wanted to confirm this is intentional rather than an unnoticed consequence of moving `setDims` inside `repaintTerminal`.

---

### Overall

The follow-up commit cleanly resolves both explicit action items from the prior review. The two items above are both low-stakes — the `setDims` question is a one-line confirm, and the all-panes focus intent is partially documented in the PR description but needs a direct answer. No blockers.

tomo-chan:
Follow-up on the latest review note for `84190a9`:

- The repaint-on-focus behavior is intentional for all mounted, attached panes in the visible workspace. A workspace switch or browser refocus can leave any attached pane with a stale canvas, so each mounted `useTerminal` instance repaints its own attached terminal when the page returns.
- The `setDims` skip during detach is also intentional. Once a pane no longer has an attached container, the measured rows/cols are no longer meaningful for UI state, so avoiding a stale dimension update is the right behavior.

Current status:

- GitHub checks are green on `84190a9`.
- No further code changes were needed for this review pass.

tomo-chan (commented):
## Code Review

Both open items from the 2026-05-06 review pass are now answered:

- **All-pane repaint on focus — confirmed intentional.** The author's reply explains: a workspace switch or browser refocus can leave any attached pane with a stale canvas, so each mounted `useTerminal` instance repaints its own attached terminal when the page returns. The PR description was updated to reflect this ("Repaint each mounted, attached terminal pane on `window.focus`..."), which removes the earlier ambiguity. No code change needed.

- **`setDims` skip during detach — confirmed intentional.** Once a pane no longer has an attached container, rows/cols are no longer meaningful for UI state, so skipping the `setDims` call is the correct behavior. No code change needed.

---

### Overall

All concerns from both prior review passes are resolved. No outstanding questions. CI is green on `84190a9`. Ready to merge.

PR #76, opened 2026-05-08
fix: replay terminal output after workspace switches

tomo-chan:
## Summary

- replay the most recent terminal output when a pane reconnects after a workspace switch or reload
- delay the initial frontend resize until xterm has a non-zero size and buffer messages until the terminal is attached
- add a Playwright E2E regression that verifies a previously hidden workspace does not render a blank terminal when first selected
- document the new reconnect and replay behavior for inactive workspace panes

## Root cause

Inactive workspace panes were unmounted, so their WebSocket readers disappeared while the backend PTY kept producing output. When the pane reconnected, the backend only sent a `connected` status and then streamed new bytes, which left panes blank until some later interaction caused fresh shell output to appear.

## Impact

Workspace switches now restore the recent terminal contents immediately instead of showing an empty pane. The frontend also avoids dropping early output or sending a `0x0` resize during reconnect, and CI now keeps a real-browser regression for this class of bug.

## Test plan

- [x] `make test-frontend`
- [x] `make test-go`
- [x] `make test-e2e`
- [x] `make lint-frontend`
- [x] `make lint-go`

### DIFF
--- a/frontend/e2e/workspace-switch.yml
+++ b/frontend/e2e/workspace-switch.yml
@@ -2,7 +2,7 @@ server:
     host: 127.0.0.1
     port: 4173
 workspaces:
-    active: workspace-2
+    active: default
     tab_position: top
     items:
         - id: default
--- a/frontend/e2e/workspace-switch.yml
+++ b/frontend/e2e/workspace-switch.yml
@@ -1,0 +1,34 @@
+server:
+    host: 127.0.0.1
+    port: 4173
+workspaces:
+    active: workspace-2
+    tab_position: top
+    items:
+        - id: default
+          title: Default
+          layout:
+            direction: horizontal
+            children:
+                - pane:
+                    id: default-main
+                    type: local
+                    shell: /bin/bash
+                    cwd: /tmp
+                    title: Default Shell
+                  size: 100
+        - id: workspace-2
+          title: Workspace 2
+          layout:
+            direction: horizontal
+            children:
+                - pane:
+                    id: workspace-2-main
+                    type: local
+                    shell: /bin/bash
+                    cwd: /tmp
+                    title: Workspace 2 Shell
+                  size: 100
+display:
+    show_header: true
+    show_status_bar: true
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -26,7 +26,6 @@
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.0.0",
-    "@playwright/test": "^1.55.1",
     "@types/react": "^18.3.11",
     "@types/react-dom": "^18.3.1",
     "@vitejs/plugin-react": "^5.2.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -10,7 +10,8 @@
     "test": "vitest run",
     "test:watch": "vitest",
     "test:ci": "vitest run --reporter=default --reporter=junit --outputFile=../test-results/vitest.xml",
-    "coverage": "vitest run --coverage"
+    "coverage": "vitest run --coverage",
+    "test:e2e": "playwright test"
   },
   "dependencies": {
     "@xterm/addon-fit": "^0.10.0",
@@ -21,9 +22,11 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "@playwright/test": "^1.59.1",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.0.0",
+    "@playwright/test": "^1.55.1",
     "@types/react": "^18.3.11",
     "@types/react-dom": "^18.3.1",
     "@vitejs/plugin-react": "^5.2.0",
--- a/internal/session/manager.go
+++ b/internal/session/manager.go
@@ -3,6 +3,7 @@ package session
 import (
 	"fmt"
 	"io"
+	"log"
 	"sync"
 )
 
@@ -134,6 +135,7 @@ func (m *managedSession) pump() {
 				m.closeSubscribers()
 				return
 			}
+			log.Printf("session %s read error: %v", m.session.ID(), err)
 			m.closeSubscribers()
 			return
 		}
@@ -142,16 +144,20 @@ func (m *managedSession) pump() {
 
 func (m *managedSession) publish(chunk []byte) {
 	m.mu.Lock()
-	defer m.mu.Unlock()
-
 	m.history = append(m.history, chunk...)
 	if len(m.history) > sessionReplayLimitBytes {
 		m.history = append([]byte(nil), m.history[len(m.history)-sessionReplayLimitBytes:]...)
 	}
 
 	for _, subscriber := range m.subscribers {
-		subscriber <- append([]byte(nil), chunk...)
+		select {
+		case subscriber <- append([]byte(nil), chunk...):
+		default:
+			// A slow client must not block session pumping or subscription cleanup.
+			// The recent replay buffer remains the source of truth for reconnects.
+		}
 	}
+	m.mu.Unlock()
 }
 
 func (m *managedSession) subscribe() ([]byte, <-chan []byte, func()) {
--- a/internal/session/manager.go
+++ b/internal/session/manager.go
@@ -2,35 +2,74 @@ package session
 
 import (
 	"fmt"
+	"io"
 	"sync"
 )
 
+const sessionReplayLimitBytes = 256 * 1024
+
 // Manager manages the lifecycle of all terminal sessions.
 type Manager struct {
-	sessions map[string]Session
+	sessions map[string]*managedSession
 	mu       sync.RWMutex
+}
+
+// managedSession keeps the session handle and replay state together so
+// subscription lifecycle changes stay under one owner.
+//
+//nolint:govet // fieldalignment: clarity is preferred over splitting this tiny state holder.
+type managedSession struct {
+	session        Session
+	history        []byte
+	subscribers    map[int]chan []byte
+	nextSubscriber int
+	closed         bool
+	mu             sync.Mutex
 }
 
 // NewManager creates a new session manager.
 func NewManager() *Manager {
 	return &Manager{
-		sessions: make(map[string]Session),
+		sessions: make(map[string]*managedSession),
 	}
 }
 
-// Add registers a session with the manager.
+// Add registers a session with the manager and starts buffering its output.
 func (m *Manager) Add(s Session) {
+	entry := &managedSession{
+		session:     s,
+		subscribers: make(map[int]chan []byte),
+	}
+
 	m.mu.Lock()
-	defer m.mu.Unlock()
-	m.sessions[s.ID()] = s
+	m.sessions[s.ID()] = entry
+	m.mu.Unlock()
+
+	go entry.pump()
 }
 
 // Get retrieves a session by ID.
 func (m *Manager) Get(id string) (Session, bool) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
-	s, ok := m.sessions[id]
-	return s, ok
+	entry, ok := m.sessions[id]
+	if !ok {
+		return nil, false
+	}
+	return entry.session, true
+}
+
+// Subscribe returns buffered output plus a live stream for the session.
+func (m *Manager) Subscribe(id string) ([]byte, <-chan []byte, func(), bool) {
+	m.mu.RLock()
+	entry, ok := m.sessions[id]
+	m.mu.RUnlock()
+	if !ok {
+		return nil, nil, nil, false
+	}
+
+	snapshot, stream, unsubscribe := entry.subscribe()
+	return snapshot, stream, unsubscribe, true
 }
 
 // List returns all current sessions.
@@ -38,8 +77,8 @@ func (m *Manager) List() []Session {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 	list := make([]Session, 0, len(m.sessions))
-	for _, s := range m.sessions {
-		list = append(list, s)
+	for _, entry := range m.sessions {
+		list = append(list, entry.session)
 	}
 	return list
 }
@@ -47,7 +86,7 @@ func (m *Manager) List() []Session {
 // Remove closes and removes a session.
 func (m *Manager) Remove(id string) error {
 	m.mu.Lock()
-	s, ok := m.sessions[id]
+	entry, ok := m.sessions[id]
 	if ok {
 		delete(m.sessions, id)
 	}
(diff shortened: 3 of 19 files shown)


internal/session/manager.go
```diff
@@ -56,20 +103,95 @@ func (m *Manager) Remove(id string) error {
...
+			m.closeSubscribers()
+			return
+		}
+	}
+}
+
+func (m *managedSession) publish(chunk []byte) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
```
tomo-chan: This channel send happens while `m.mu` is held. If this channel is full (capacity 64), the goroutine blocks here indefinitely with the lock, which deadlocks any concurrent `subscribe` or `unsubscribe` call (including the `defer unsubscribe()` fired on WebSocket disconnect). Consider collecting the channels under the lock and sending outside it, or using a non-blocking `select { case ch <- chunk: default: }`.

internal/ws/handler.go
```diff
@@ -76,38 +82,52 @@ func (h *Handler) sessionForRequest(w http.ResponseWriter, sessionID string) ses
...
+	}
+
+	for chunk := range updates {
+		if len(chunk) == 0 {
+			continue
+		}
+		if writeErr := conn.WriteMessage(websocket.BinaryMessage, chunk); writeErr != nil {
 			return
 		}
 	}
```
tomo-chan, line 120: Previously `forwardTerminalOutput` always sent `"exited"` when the read loop ended. Now it only sends it when `sess.State() == StateExited`. For sessions that end via SSH disconnection or PTY error, `pump` closes the channel but `sess.State()` may not be `StateExited`, so the frontend never receives the status frame and the terminal silently hangs. Is this intentional?

tomo-chan (commented):
## Code Review

### What's Good

- The root-cause analysis in the PR description is precise and the fix addresses it at the right layer: buffering on the backend rather than the frontend avoids the race between pane unmount and reconnect entirely.
- Splitting `TerminalEntry` into `handleMessage` / `applyMessageToTerminal` / `canApplyMessage` makes the pending-buffer logic easy to follow and keeps the happy path readable.
- `scheduleConnectedResize` correctly mirrors the existing repaint-settle pattern and is cleanly symmetric — `clearScheduledResizes` is called in both the normal cleanup and the dispose timeout paths.
- `flushPendingMessages` mutates `pendingMessages.length = 0` in-place rather than replacing the ref, which is a subtle but correct choice given the closure captures the array reference.
- The `//nolint:govet // fieldalignment` annotation on `managedSession` is well-reasoned: field ordering that aids human readers is the right call for a small struct like this.
- Test coverage is thorough on both sides: the manager tests use `require.Eventually` instead of fixed sleeps, and the frontend tests cover both the zero-size resize guard and the pre-attach buffer in isolation.

---

### Issues & Questions

**[Question] Deadlock: channel send inside `publish` while holding `m.mu`**

`publish` holds `m.mu` for the full loop that sends to subscriber channels:

```go
func (m *managedSession) publish(chunk []byte) {
    m.mu.Lock()
    defer m.mu.Unlock()
    // …
    for _, subscriber := range m.subscribers {
        subscriber <- append([]byte(nil), chunk...)  // blocks here if channel is full
    }
}
```

Each subscriber channel has capacity 64. If a slow WebSocket writer or a brief network stall lets the channel fill up, the `pump` goroutine blocks indefinitely while holding the lock. Both `subscribe` and `unsubscribe` (called via `defer unsubscribe()` in `ServeHTTP` on every disconnect) also take `m.mu`, so the blocked pump goroutine will deadlock every concurrent subscribe/unsubscribe on that session — including the disconnect cleanup.

A common fix is to collect the channels under the lock, release the lock, then send:

```go
m.mu.Lock()
// update history …
chs := make([]chan []byte, 0, len(m.subscribers))
for _, ch := range m.subscribers {
    chs = append(chs, ch)
}
m.mu.Unlock()

for _, ch := range chs {
    ch <- append([]byte(nil), chunk...)
}
```

(Requires adding a sent-to-closed guard in `unsubscribe`, or switching to non-blocking sends with `select { case ch <- chunk: default: }` if occasional drops under extreme load are acceptable.)

Is this intentional? The 64-buffer feels like it was sized to tolerate typical latency, but terminal sessions can burst (e.g. `cat` of a large file, `make` output) and a single slow client would freeze the pump for all other subscribers too.

**[Question] Behavioral regression: "exited" status no longer sent for non-`StateExited` closes**

The original `forwardTerminalOutput` always sent `h.sendStatus(conn, "exited")` when the read loop ended, regardless of the error kind. The new version only sends it if `sess.State() == StateExited`:

```go
for chunk := range updates {  // exits when channel closes
    …
}

if sess.State() == session.StateExited {
    h.sendStatus(conn, "exited")  // ← only this branch sends "exited"
    return
}
// otherwise: just a log line, no status frame
```

For sessions that terminate via a non-EOF error — SSH disconnection mid-session, PTY read error, forced `Remove` — `pump` calls `closeSubscribers()` and the channel closes, but `sess.State()` may still be `StateConnected` or `StateDisconnected`. The frontend never receives `"exited"`, so it silently drops the WebSocket connection without showing `[Session ended]` or setting `sessionExited`, leaving the terminal in a hung state.

Was this intentional? If the goal is to show `[Session ended]` only for clean shell exits (not errors), the PR description doesn't mention that change. The safer default would be to send `"exited"` whenever the channel drains, and use `sess.State()` to pick a different status message if needed.

**[Minor question] `pump` drops non-EOF read errors silently**

Both branches of the error check in `pump` are identical (`closeSubscribers(); return`), with no log for the non-EOF path:

```go
if err \!= nil {
    if err == io.EOF {
        m.closeSubscribers()
        return
    }
    m.closeSubscribers()
    return  // error quietly discarded
}
```

The original `forwardTerminalOutput` logged non-EOF errors. A one-liner `log.Printf("session %s read error: %v", m.session.ID(), err)` before the non-EOF return would preserve debuggability without changing behaviour.

**[Minor question] `Subscribe` not-found guard in `ServeHTTP` is unreachable**

`sessionForRequest` already calls `h.manager.Get` and returns `nil` (with a 404) if the session is missing. The `Subscribe` call immediately after can therefore never return `\!ok` — the session existence was just confirmed. The dead-code path is harmless, but it might confuse future readers into thinking there is a TOCTOU window that could fire in practice. Worth a comment or removing the guard.

---

### Overall

The frontend-side buffering and resize-delay logic look solid and are well-tested. The main concern before merging is the **deadlock in `publish`** (channel send under mutex): in a burst-output session with a single slow subscriber, the entire subscription mechanism for that session locks up, including disconnect cleanup. The **"exited" status regression** is a secondary but user-visible behavioural change that should at least be called out explicitly.

tomo-chan:
Addressed the review findings in `1f88977`.

- Slow subscribers no longer block `managedSession.publish` while holding the session mutex. Live fan-out is now best-effort per subscriber, which prevents one stalled WebSocket writer from freezing subscribe/unsubscribe cleanup for the whole session.
- WebSocket clients now receive `{"type":"status","state":"exited"}` whenever the session stream closes, including non-EOF read failures. This restores the frontend `[Session ended]` behavior for disconnected/error cases.
- Non-EOF session read failures are logged again for debugging.

Local verification run after the change:

- `go test ./internal/session ./internal/ws`
- `make test-go`
- `make lint-go`

frontend/e2e/workspace-switch.yml
```diff
@@ -0,0 +1,34 @@
+server:
+    host: 127.0.0.1
+    port: 4173
+workspaces:
+    active: workspace-2
+    tab_position: top
```
tomo-chan, line 6: `active: workspace-2` makes workspace-2 the initially displayed workspace. WorkspaceTabs sets `aria-selected` from `workspaces.active`, so the "Workspace 2" tab — not "Default" — would have `aria-selected: true` on page load. The spec's first assertion (`'Default' toHaveAttribute 'aria-selected', 'true'`) would fail. Changing this to `active: default` would fix the mismatch; workspace-2's sessions still start at boot (via `startSessionsFromConfig`) so there is still replay-buffered output when the test switches to it.

frontend/package.json
```diff
@@ -21,9 +22,11 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "@playwright/test": "^1.59.1",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.0.0",
+    "@playwright/test": "^1.55.1",
     "@types/react": "^18.3.11",
```
tomo-chan, line 29: Duplicate `@playwright/test` key — both `^1.59.1` (line above) and `^1.55.1` (this line) appear in `devDependencies`. npm uses the last occurrence (`^1.55.1`), but `package-lock.json` records `1.59.1`. A fresh `npm install` would downgrade Playwright. One of the two entries needs to be removed.

tomo-chan (commented):
## Code Review (follow-up — 1f88977 + fe5c665)

### Previous findings — status

All three issues from the earlier review were addressed in `1f88977`:

- **Deadlock in `publish`** — resolved. Non-blocking `select { case ch <- …: default: }` eliminates the lock-while-sending hazard. The new test `TestManagedSession_SubscribeStillWorksWhilePublishWaitsOnSlowSubscriber` directly exercises the previously deadlocking scenario; good direct coverage.
- **`"exited"` status regression** — resolved. `h.sendStatus(conn, "exited")` is now unconditional after the channel drains, restoring the original frontend `[Session ended]` behavior for error cases too.
- **Non-EOF errors silently dropped in `pump`** — resolved. `log.Printf` added before the non-EOF return.

### What's Good (new commits)

- The Playwright E2E layer targets the one thing unit tests cannot verify: whether xterm actually renders output in a real Chromium renderer. Using a real `go run .` server process rather than a stub is the right call — it exercises the full stack.
- `run-panemux-e2e.sh` forwards `GOCACHE` to a temp path, keeping `go run` builds isolated from the developer's cache.
- `reuseExistingServer: \!process.env.CI` is a nice DX touch — lets devs re-run tests without restarting the server.
- The `TestWS_SendsExitedStatusWhenSessionStreamEndsWithReadError` test is a clean regression guard: it wires `readErr = io.ErrUnexpectedEOF` into the mock and asserts the `"exited"` frame arrives.

---

### Issues & Questions

**[Question] `active: workspace-2` in the E2E YAML conflicts with the test's first assertion**

`workspace-switch.yml` sets `active: workspace-2`, which means the frontend will display Workspace 2 on initial page load (WorkspaceTabs sets `aria-selected` from `workspaces.active`, and `useLayout` uses `parsed.active` directly rather than falling back to `items[0]`).

The test then immediately asserts:
```typescript
await expect(page.getByRole('tab', { name: 'Default' })).toHaveAttribute('aria-selected', 'true')
```

This assertion will fail as written — Workspace 2 would have `aria-selected: true`, not Default.

The fix is to change the config to `active: default`. Workspace 2's sessions are started at server boot regardless of which workspace is `active` (all panes in the config get sessions via `startSessionsFromConfig`), so there will still be replay-buffered output when the test switches to Workspace 2.

**[Question] Duplicate `@playwright/test` key in `package.json`**

`package.json` now contains two `@playwright/test` entries in `devDependencies`:

```json
"@playwright/test": "^1.59.1",   ← first entry
…
"@playwright/test": "^1.55.1",   ← second entry (last one wins in most parsers)
```

Most JSON parsers (and npm) use the last occurrence, so the effective constraint would be `^1.55.1`, while `package-lock.json` has `1.59.1` installed. A fresh `npm install` would downgrade Playwright. One of the two entries should be removed.

---

### Overall

`1f88977` cleanly addresses all three earlier concerns. Two issues remain in the E2E commit (`fe5c665`): the `active` field mismatch would cause the Playwright test to fail on the first assertion, and the duplicate `package.json` entry could silently downgrade Playwright in CI. Both are straightforward one-line fixes.

tomo-chan:
Addressed the follow-up E2E review in `f53a6e4`.

- The Playwright fixture config now starts on `active: default`, which matches the test's first assertion while still leaving `workspace-2` hidden until the switch under test.
- The duplicate `@playwright/test` entry was removed from `frontend/package.json`, so fresh installs keep the same version recorded in `package-lock.json`.

Local verification run after the change:

- `make test-frontend`
- `make test-e2e`

internal/session/manager.go
tomo-chan: Addressed in 1f88977.

publish now uses a non-blocking send per subscriber, so a full channel no longer blocks while holding the session mutex. That prevents one slow WebSocket consumer from stalling subscribe/unsubscribe cleanup for the whole session.

Added regression coverage in TestManagedSession_SubscribeStillWorksWhilePublishWaitsOnSlowSubscriber.

Verified with:
- go test ./internal/session ./internal/ws
- make test-go
- make lint-go

frontend/package.json
tomo-chan: Addressed in f53a6e4.

Removed the duplicate @playwright/test entry from frontend/package.json so fresh installs use the same version recorded in package-lock.json instead of silently taking the later downgraded key.

Verified with:
- make test-frontend
- make test-e2e

frontend/e2e/workspace-switch.yml
tomo-chan: Addressed in f53a6e4.

The E2E fixture config now starts on active: default, which matches the test's first assertion while still keeping workspace-2 hidden until the switch under test.

Verified with:
- make test-frontend
- make test-e2e

internal/ws/handler.go
tomo-chan: Addressed in 1f88977.

forwardTerminalOutput now sends the exited status after the update stream closes regardless of the session state, so the frontend gets the terminal-ended signal again for disconnect and read-error cases as well.

Added regression coverage in TestWS_SendsExitedStatusWhenSessionStreamEndsWithReadError.

Verified with:
- go test ./internal/session ./internal/ws
- make test-go
- make lint-go

tomo-chan (commented):
## Code Review (follow-up — f53a6e4)

### All findings resolved

Every concern from the two earlier passes is addressed:

| Finding | Status | Commit |
|---|---|---|
| Deadlock in `publish` (channel send under lock) | ✅ Fixed: non-blocking `select { case ch <- …: default: }` | `1f88977` |
| `"exited"` status not sent for non-`StateExited` closes | ✅ Fixed: `sendStatus` is now unconditional after channel drains | `1f88977` |
| Non-EOF session read errors silently dropped in `pump` | ✅ Fixed: `log.Printf` added | `1f88977` |
| `active: workspace-2` conflicts with test's first assertion | ✅ Fixed: changed to `active: default` | `f53a6e4` |
| Duplicate `@playwright/test` in `package.json` | ✅ Fixed: duplicate entry removed | `f53a6e4` |

### Additional observations on `f53a6e4`

- `vite.config.ts` adds `include: ['src/**/*.test.*']` + `exclude: ['e2e/**']`: correct and necessary — without this, Vitest would try to collect Playwright spec files and fail on the `@playwright/test` imports.
- `ci.yml` installs Chromium with `--with-deps` before the E2E step: correct for a clean Ubuntu runner that won't have the browser system libraries pre-installed.

### Overall

The PR is clean. All review findings are resolved, the E2E fixture config is consistent with the test assertions, and the test infrastructure is wired up correctly end-to-end. Ready to merge when you want to promote from draft.

PR #77, opened 2026-05-09
fix: restore hidden workspace attention notifications

tomo-chan:
## Summary
- restore agent attention detection for panes in inactive workspaces by monitoring every workspace pane in the frontend
- request browser notification permission on first user interaction so the first attention event is not dropped by a blocked permission prompt
- add unit and Playwright coverage for hidden workspace attention and browser notification behavior

## Root Cause
Attention detection only ran inside mounted terminal panes. Inactive workspaces unmounted their panes, so their output was never scanned for confirmation prompts. Browser notifications were also requested at the moment of the first prompt, which browsers can suppress when it is not tied to a user gesture.

## Impact
Inactive workspace tabs now light up when a hidden pane emits an agent confirmation prompt, and granted browser notifications fire for that prompt as well. Users are prompted for notification permission on their first interaction instead of losing the first alert.

## Test Plan
- [x] make test-frontend
- [x] make build-frontend
- [x] make lint-frontend
- [x] make check
- [x] cd frontend && npm run test:e2e


tomo-chan (commented):
## Code Review

### What's Good

- **Clean separation of concerns** — `useWorkspaceAttentionMonitor` decouples attention detection from the xterm mount lifecycle. The hook is small, focused, and independently testable.
- **Permission UX fix** — Moving `Notification.requestPermission()` to the first pointer/key interaction is the correct browser pattern; browsers require a user gesture and previously the first prompt could silently drop.
- **O(1) pane metadata lookup** — `paneMetadataByID` (a `Map` built from all workspace layouts) replaces the O(n) `findWorkspaceForPane` tree walk on the hot notification path.
- **Throttle + reconnect reset** — `ATTENTION_NOTIFY_INTERVAL_MS = 10_000` prevents spam, and `ws.onopen` resets both `lastAttentionAt` and the detector, so reconnects don't inherit stale state.
- **Cleanup correctness** — Nulling `socket.onclose` and `socket.onerror` before `socket.close()` in the effect teardown prevents spurious callbacks after intentional teardown.
- **Thorough test coverage** — Both new hooks have dedicated unit tests covering deduplication, cleanup on unmount, text-frame filtering, and error paths. The E2E test covers the full flow with a mocked `Notification` class.

---

### Issues & Questions

**[Question] Dual WebSocket connections for active-workspace panes**

`useWorkspaceAttentionMonitor` opens a WebSocket for **every** pane across all workspaces, including panes in the currently active workspace. `useTerminal` (via `useWebSocket`) also opens a WebSocket for the same session ID for those same panes. At steady state, each active pane has two concurrent connections to `/ws/{sessionId}`.

The architecture doc says "one socket per session" — does `ws/handler.go` support multiple simultaneous readers for the same session ID? If the handler allows only one live connection (e.g., closes the previous when a new one arrives), the monitor's persistent socket could intermittently drop the terminal's interactive socket or vice versa.

**[Question] Dead code in `useTerminal` / `useWebSocket`**

`TerminalPane.tsx` no longer passes `onAttention: ctx?.onPaneAttention` to `useTerminal`, but only the call site changed in this diff. Is the `onAttention` option still present in `useTerminal`'s interface and used internally? If so, there is now unreachable code that should be cleaned up in this PR.

**[Minor question] `notifyAttention` still calls `findPaneById` when `paneMetadata` is available**

When `paneMetadataByID.get(paneId)` hits (which it always should for panes the monitor watches), `paneTitle` is already in `paneMetadata.paneTitle` (set as `child.pane.title ?? child.pane.id`, never `undefined`). Yet the code still resolves `workspace` via `workspaces.items.find(...)` and calls `findPaneById(workspace.layout, paneId)` as a secondary source for `pane?.title`, which is then shadowed by `paneMetadata.paneTitle`:

```ts
// App.tsx – notifyAttention
const pane = workspace ? findPaneById(workspace.layout, paneId) : …  // tree walk
const paneTitle = paneMetadata?.paneTitle ?? pane?.title ?? paneId    // pane?.title unused when metadata hits
```

Is the fallback path intentional (e.g., for future callers that bypass the monitor), or can the `workspace`/`pane` resolution block be removed as a partial refactor cleanup?

**[Minor question] No test for active-workspace pane attention**

The unit tests confirm that attention for a pane in an **inactive** workspace marks that tab. There is no test for the inverse case — where `onAttention` fires for a pane in the **active** workspace — verifying that the active tab does *not* get `data-attention`. Since `useWorkspaceAttentionMonitor` monitors all panes including active ones, this path is reachable in production and worth a test case.

**[Suggestion] `collectChildPaneIDs` / `collectChildPaneMetadata` logic duplication**

`collectChildPaneIDs` in `useWorkspaceAttentionMonitor.ts` and `collectChildPaneMetadata` in `App.tsx` both recurse over `LayoutChild` trees with identical guard logic. They could be unified into a single utility in `utils/layoutTree.ts`. Happy to file a follow-up issue for this if preferred.

---

### Overall

The architectural direction is right and the new hooks are well-tested. The main open question before un-drafting is whether the backend supports multiple concurrent WebSocket readers for the same session ID — that assumption underlies active-workspace pane monitoring and determines whether the monitor socket can safely coexist with the terminal's interactive socket.

frontend/src/hooks/useWorkspaceAttentionMonitor.ts
```diff
@@ -0,0 +1,110 @@
...
+  }, [workspaces])
+
+  useEffect(() => {
+    if (paneIds.length === 0) return
+
+    const sockets = paneIds.map((paneId) => {
+      const state = getOrCreatePaneMonitorState(monitorStatesRef.current, paneId)
+      const ws = new WebSocket(buildWebSocketURL(paneId))
+      ws.binaryType = 'arraybuffer'
+
```
tomo-chan, line 34: This line opens a new WebSocket for every pane, including panes already mounted in the active workspace (which have their own socket via `useTerminal`). At steady state, active-workspace panes have two concurrent connections to the same `/ws/{sessionId}`. The architecture doc says "one socket per session" — does `ws/handler.go` support multiple simultaneous readers per session ID, or will this background socket collide with the terminal's interactive socket?

tomo-chan:
Addressed the review follow-up in `8475f6f`.

- Removed the dead `onAttention` path from `useTerminal` and deleted the now-unreachable terminal attention tests.
- Added an explicit App test for the active-workspace case so active tabs do not get `data-attention` when one of their own panes triggers attention.
- Clarified the architecture note around `/ws/{sessionId}`: the interactive terminal still uses one primary socket per visible pane, and the backend session manager safely fans out one session stream to multiple concurrent subscribers.

On the multi-subscriber question: this PR relies on `internal/session.Manager.Subscribe()` fan-out rather than a single-reader WebSocket model. Each session keeps a subscriber map and publishes the same buffered/live chunks to all subscribers, so the background monitor socket can coexist with the interactive terminal socket for the same session ID.

I kept the `notifyAttention` fallback path in `App.tsx` intentionally so the handler still behaves sensibly for direct callers that provide only a pane ID. The hot path uses `paneMetadataByID` first; the tree walk is just a defensive fallback now.

Local verification after this follow-up:
- `make test-frontend`
- `make lint-frontend`
- `make build-frontend`

I also attempted to rerun `cd frontend && npm run test:e2e`, but the local Playwright web server hit a leftover `:4173` bind conflict in this environment after the earlier successful E2E run, so this follow-up push does not include a fresh E2E rerun log.

tomo-chan (commented):
## Code Review (follow-up on prior review)

### Resolved

All items from the previous review are addressed in `8475f6f` and the author's reply:

- **Dead code in `useTerminal`** — fully removed: `onAttention` option, `attentionDetectorRef`, `outputDecoderRef`, `lastAttentionAtRef`, the `shouldNotifyAttention` helper, and the four unit tests that exercised them. The diff is clean.
- **Active-workspace pane test** — `'does not mark the active workspace tab when attention comes from the active workspace'` added to `App.test.tsx`.
- **Multi-subscriber architecture** — the author confirmed `internal/session.Manager.Subscribe()` fan-out, and `docs/architecture.md` now explicitly documents that one session stream fans out to multiple concurrent subscribers. That settles the concurrent-connection question.
- **`notifyAttention` fallback** — author confirmed the `findPaneById` fallback is intentional for direct callers that bypass the monitor. Understood.

---

### Issues & Questions

**[Minor question] E2E not re-verified against `8475f6f`**

The follow-up comment notes that `npm run test:e2e` couldn't be re-run after `8475f6f` due to a local port-conflict. The changes in that commit are pure deletions from `useTerminal` (code paths the E2E spec doesn't reach), so a re-run is unlikely to surface failures. That said, the test-plan checkbox from the initial commit is now technically stale against HEAD. Is a clean E2E run expected before un-drafting, or is the prior green run sufficient given the deletion-only nature of the follow-up?

**[Suggestion] Monitor socket churn on workspace switch**

When the user switches workspaces, `workspaces` state refreshes → `notifyAttention` is recreated (it includes `workspaces` in its `useCallback` deps) → `useWorkspaceAttentionMonitor`'s effect tears down and reopens **all** monitor sockets, even though pane IDs are unchanged. A `useRef` holding `onAttention` and removed from the `useEffect` dep array would prevent this churn. Flagging as a follow-up; it doesn't affect correctness.

**[Suggestion] `collectChildPaneIDs` / `collectChildPaneMetadata` logic duplication** (carried forward)

Both functions recurse over `LayoutChild` trees with identical guard logic — one in `useWorkspaceAttentionMonitor.ts`, one in `App.tsx`. Could be unified in `utils/layoutTree.ts`. Happy to file a follow-up issue if useful.

---

### Overall

The PR is in good shape; all blocking concerns from the first review are resolved. The only open question is whether a fresh E2E run is required against the current HEAD before un-drafting.

tomo-chan:
Deferred the non-blocking follow-ups to issues:

- #78 `refactor: avoid monitor websocket churn on workspace switch`
- #79 `refactor: unify frontend layout tree traversal helpers`

These are both valid cleanups, but neither blocks the correctness of this PR.

tomo-chan (commented):
## Code Review (round 3)

### All prior items resolved

- **[Suggestion] Monitor socket churn** → filed as #78. ✓
- **[Suggestion] Layout traversal duplication** → filed as #79. ✓
- All other questions from the previous two rounds are addressed in the code or the author's replies.

---

### Overall

No new concerns. E2E is covered by CI (`make test-e2e` in the `ci` job, which passed on the current HEAD). The PR is ready to un-draft.

frontend/src/hooks/useWorkspaceAttentionMonitor.ts
tomo-chan: Addressed in `8475f6f8a32fba6b96d261a95efea3b2404bd212`.

`useWorkspaceAttentionMonitor` does keep a second WebSocket for active panes, but this is safe with the current backend design: `internal/session.Manager.Subscribe()` fans one session stream out to multiple concurrent subscribers via its `subscribers` map, so the background monitor socket does not replace or evict the interactive terminal socket for the same session ID.

I also updated `docs/architecture.md` to make that explicit, because the old "one socket per session" wording was misleading once the background monitor was added.

Related follow-up cleanups from the later review are also handled:
- dead `useTerminal` attention path removed in the same commit
- active-workspace non-attention case covered by test
- non-blocking churn/duplication follow-ups filed as #78 and #79

PR #81, opened 2026-05-09
fix: harden WebSocket, CORS, security headers, and input validation

tomo-chan:
## Summary

- **Prevent cross-site WebSocket hijacking (CSWSH)**: Replace `CheckOrigin: return true` with same-host origin validation. Requests without an Origin header (non-browser clients) are still allowed through.
- **WebSocket 1 MB read limit**: Cap incoming message size to mitigate memory-exhaustion DoS.
- **Drop CORS wildcard**: Replace `Access-Control-Allow-Origin: *` with origin reflection restricted to `localhost`, `127.0.0.1`, and `::1`. The Vite dev server (cross-port `:5173` → `:8080`) continues to work.
- **Add security headers**: `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, `Referrer-Policy: strict-origin-when-cross-origin` via a new `securityHeadersMiddleware`.
- **Harden HTTP server**: Add `IdleTimeout: 120s` and `MaxHeaderBytes: 1 MB` to mitigate slowloris-style attacks.
- **Prevent ANSI injection**: Strip `\x1b` bytes from backend error messages before interpolating them into the terminal's ANSI escape sequence, preventing injected sequences from executing in the terminal.
- **Zod schema bounds**: Add `.max()` limits to `shell` (512), `cwd` (4096), `title`/`connection`/`tmux_session` (256), recursive `children` arrays (50), and error `message` (2000).

## Test plan

- [x] `make lint-go` — Go lint clean (0 issues)
- [x] `make test-frontend` — all 258 frontend tests pass including new cases
- [x] `make lint-frontend` — no TypeScript errors
- [x] `go test ./internal/server/...` — all server tests pass including new CORS, security-headers, and `isLocalhostOrigin` tests
- [x] `go test ./internal/ws/... -run TestCheckOrigin` — all 6 `checkOrigin` table-driven cases pass
- [x] Manual: `./bin/panemux` → browser connects normally; response headers include `X-Content-Type-Options`, `X-Frame-Options`, and `Referrer-Policy`

### DIFF
--- a/frontend/src/hooks/useTerminal.ts
+++ b/frontend/src/hooks/useTerminal.ts
@@ -58,7 +58,7 @@ export function useTerminal({ sessionId, container, editMode = false }: UseTermi
             term.write('\r\n\x1b[2m[Session ended]\x1b[0m\r\n')
           }
         } else if (msg.type === 'error') {
-          const safeMsg = msg.message.replace(/\x1b\[[0-9;]*[A-Za-z]/g, '')
+          const safeMsg = msg.message.replace(/\x1b\[[0-9;?<>!]*[A-Za-z]/g, '')
           term.write(`\r\n\x1b[31mError: ${safeMsg}\x1b[0m\r\n`)
         }
       } catch {
--- a/frontend/src/hooks/useTerminal.ts
+++ b/frontend/src/hooks/useTerminal.ts
@@ -58,7 +58,7 @@ export function useTerminal({ sessionId, container, editMode = false }: UseTermi
             term.write('\r\n\x1b[2m[Session ended]\x1b[0m\r\n')
           }
         } else if (msg.type === 'error') {
-          const safeMsg = msg.message.replace(/\x1b/g, '')
+          const safeMsg = msg.message.replace(/\x1b\[[0-9;]*[A-Za-z]/g, '')
           term.write(`\r\n\x1b[31mError: ${safeMsg}\x1b[0m\r\n`)
         }
       } catch {
--- a/frontend/src/hooks/useTerminal.ts
+++ b/frontend/src/hooks/useTerminal.ts
@@ -58,7 +58,8 @@ export function useTerminal({ sessionId, container, editMode = false }: UseTermi
             term.write('\r\n\x1b[2m[Session ended]\x1b[0m\r\n')
           }
         } else if (msg.type === 'error') {
-          term.write(`\r\n\x1b[31mError: ${msg.message}\x1b[0m\r\n`)
+          const safeMsg = msg.message.replace(/\x1b/g, '')
+          term.write(`\r\n\x1b[31mError: ${safeMsg}\x1b[0m\r\n`)
         }
       } catch {
         // Not JSON, treat as text
--- a/internal/ws/handler.go
+++ b/internal/ws/handler.go
@@ -4,6 +4,7 @@ package ws
 import (
 	"encoding/json"
 	"log"
+	"net"
 	"net/http"
 	"net/url"
 	"time"
@@ -22,9 +23,16 @@ var upgrader = websocket.Upgrader{
 	CheckOrigin:     checkOrigin,
 }
 
-// checkOrigin validates that WebSocket upgrade requests originate from the same host,
-// preventing cross-site WebSocket hijacking (CSWSH). Non-browser clients that omit
-// the Origin header are allowed through.
+// checkOrigin validates WebSocket upgrade requests to prevent cross-site WebSocket
+// hijacking (CSWSH).
+//
+// Loopback origins (localhost / 127.0.0.1 / ::1) are permitted regardless of port so
+// that the Vite dev server on :5173 can proxy WebSocket traffic to the backend on
+// :8080 without requiring changeOrigin in the proxy config.
+//
+// Requests without an Origin header are allowed; browsers always include Origin on
+// cross-origin requests, so the absence of the header indicates a non-browser client
+// (e.g. curl) that is not subject to the same-origin policy.
 func checkOrigin(r *http.Request) bool {
 	origin := r.Header.Get("Origin")
 	if origin == "" {
@@ -34,7 +42,20 @@ func checkOrigin(r *http.Request) bool {
 	if err != nil {
 		return false
 	}
+	if isLoopbackHost(u.Host) {
+		return true
+	}
 	return u.Host == r.Host
+}
+
+// isLoopbackHost returns true when the host part of an authority string
+// (host or host:port) is a loopback address.
+func isLoopbackHost(authority string) bool {
+	host, _, err := net.SplitHostPort(authority)
+	if err != nil {
+		host = authority
+	}
+	return host == "localhost" || host == "127.0.0.1" || host == "::1"
 }
 
 // ControlMessage is a JSON control frame exchanged over WebSocket.
--- a/internal/ws/handler.go
+++ b/internal/ws/handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"log"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/go-chi/chi/v5"
@@ -13,12 +14,27 @@ import (
 	"panemux/internal/session"
 )
 
+const wsReadLimitBytes = 1 << 20 // 1 MB
+
 var upgrader = websocket.Upgrader{
 	ReadBufferSize:  4096,
 	WriteBufferSize: 4096,
-	CheckOrigin: func(r *http.Request) bool {
-		return true // allow all origins for local use
-	},
+	CheckOrigin:     checkOrigin,
+}
+
+// checkOrigin validates that WebSocket upgrade requests originate from the same host,
+// preventing cross-site WebSocket hijacking (CSWSH). Non-browser clients that omit
+// the Origin header are allowed through.
+func checkOrigin(r *http.Request) bool {
+	origin := r.Header.Get("Origin")
+	if origin == "" {
+		return true
+	}
+	u, err := url.Parse(origin)
+	if err != nil {
+		return false
+	}
+	return u.Host == r.Host
 }
 
 // ControlMessage is a JSON control frame exchanged over WebSocket.
@@ -63,6 +79,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	defer conn.Close() //nolint:errcheck
+	conn.SetReadLimit(wsReadLimitBytes)
 
 	h.sendStatus(conn, "connected")
 	done := h.pipeTerminalToWebSocket(conn, sess, snapshot, updates, sessionID)
--- a/frontend/src/hooks/useTerminal.test.ts
+++ b/frontend/src/hooks/useTerminal.test.ts
@@ -656,4 +656,26 @@ describe('useTerminal', () => {
     expect(written).toContain('\x1b[31m')
     expect(written).toContain('\x1b[0m')
   })
+
+  it('strips private-use CSI sequences (DEC private markers) from error messages', () => {
+    const container = makeContainer()
+    renderHook(() => useTerminal({ sessionId: 's1', container }))
+    act(() => MockWebSocket.instances[0].simulateOpen())
+
+    mockWrite.mockClear()
+    // \x1b[?25l (hide cursor) and \x1b[>1h (DEC private) are private-use CSI sequences
+    act(() =>
+      MockWebSocket.instances[0].simulateMessage(
+        JSON.stringify({ type: 'error', message: '\x1b[?25lhidden cursor\x1b[>1htext' })
+      )
+    )
+
+    expect(mockWrite).toHaveBeenCalledTimes(1)
+    const written = mockWrite.mock.calls[0][0] as string
+    expect(written).not.toContain('\x1b[?25l')
+    expect(written).not.toContain('[?25l')
+    expect(written).not.toContain('\x1b[>1h')
+    expect(written).toContain('hidden cursor')
+    expect(written).toContain('text')
+  })
 })
--- a/frontend/src/hooks/useTerminal.test.ts
+++ b/frontend/src/hooks/useTerminal.test.ts
@@ -632,7 +632,7 @@ describe('useTerminal', () => {
     expect(ws.sent.length).toBeGreaterThan(sentBefore)
   })
 
-  it('strips ESC characters from error messages before writing to terminal', () => {
+  it('strips complete ANSI sequences from error messages before writing to terminal', () => {
     const container = makeContainer()
     renderHook(() => useTerminal({ sessionId: 's1', container }))
     act(() => MockWebSocket.instances[0].simulateOpen())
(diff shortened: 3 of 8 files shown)


internal/ws/handler.go
```diff
@@ -13,12 +14,27 @@ import (
...
+	origin := r.Header.Get("Origin")
+	if origin == "" {
+		return true
+	}
+	u, err := url.Parse(origin)
+	if err != nil {
+		return false
+	}
+	return u.Host == r.Host
 }
```
tomo-chan, line 49: The PR description says Vite dev server continues to work across ports, but this line rejects `Origin: http://localhost:5173` when `Host` is `localhost:8080` (the test case is literally named "cross-port same domain rejected"). Is the expectation that Vite's WebSocket proxy strips the `Origin` header so the request reaches the backend via the `origin == ""` path? If so, a short comment on that empty-origin path would help future maintainers understand why the permissive allowance is intentional and also covers the dev-proxy case.

tomo-chan (commented):
## Code Review

### What's Good

- **`isLocalhostOrigin` handles IPv6 correctly**: using `net.SplitHostPort` to strip the port before comparing cleanly handles the `[::1]` bracket notation that `url.Parse` includes in `u.Host`. The `TestIsLocalhostOrigin` table covers this case explicitly.
- **Asymmetric CORS vs WebSocket design is well-reasoned**: REST uses `isLocalhostOrigin` (any localhost port allowed) while WebSocket uses same-host `checkOrigin` for stronger protection. Each protocol's trust model is handled correctly.
- **ANSI injection fix is minimal and effective**: stripping `\x1b` in `useTerminal.ts` before interpolating into the terminal's own escape sequence is the right place to sanitize. The test clearly documents the "intentional outer sequence stays, injected inner sequence is defanged" contract.
- **Zod bounds coverage**: both boundary cases (at-limit and over-limit) are tested for every new constraint — this is the correct pattern per the project's test-granularity rules.
- **HTTP server hardening**: `IdleTimeout: 120s` and `MaxHeaderBytes: 1 MB` address slowloris-style attacks without disrupting long-lived WebSocket connections (`WriteTimeout: 0` is correct for streaming).
- **`registerRoutes`/`registerFrontend` refactor**: extracting route registration improves testability and makes `New` easier to follow.

---

### Issues & Questions

**[Question] `checkOrigin` rejects cross-port localhost — does the Vite proxy strip `Origin`?**

`checkOrigin` compares `u.Host == r.Host`, so `Origin: http://localhost:5173` with `Host: localhost:8080` returns `false`. The test even names this explicitly:

```go
{
    name:   "cross-port same domain rejected",
    origin: "http://localhost:5173",
    host:   "localhost:8080",
    want:   false,
},
```

However, the PR description says "The Vite dev server (cross-port `:5173` → `:8080`) continues to work."

The only way both statements can be true is if Vite's WebSocket proxy strips or rewrites the `Origin` header before forwarding, making the backend see an empty origin — which is allowed via the `origin == ""` path. A few questions:

1. Does Vite's `http-proxy`-based WebSocket proxy actually strip the `Origin` header for proxied connections?
2. The test plan covers `./bin/panemux → browser connects normally` but not `make dev-frontend → browser connects via Vite proxy`. Has this been manually verified?
3. If the empty-origin path is the reason dev mode works, a short comment there would prevent a future maintainer from "tightening" the empty-origin allowance and silently breaking the dev server.

**[Minor question] ANSI strip leaves dangling sequence suffix as visible text**

```typescript
const safeMsg = msg.message.replace(/\x1b/g, '')
```

Stripping only `\x1b` turns `\x1b[1mBold\x1b[0m` into the literal string `[1mBold[0m`, which is printed as-is in the terminal. The injection is safely neutralised, but the output looks garbled for any error message that happens to contain ANSI-formatted text.

Is this the intended UX? An alternative like `.replace(/\x1b\[[0-9;]*[A-Za-z]/g, '')` would strip the full sequence and leave clean text. Not blocking — just checking whether the current behaviour is a known tradeoff.

---

### Overall

Solid, well-scoped security hardening with thorough test coverage. The main open question is the Vite dev-mode WebSocket path: if it relies on Vite stripping the `Origin` header, that assumption is worth a comment or an explicit test.

tomo-chan (commented):
Thanks for the thorough review — both issues were valid. Here's what was addressed in the follow-up commit (https://github.com/tomo-chan/panemux/commit/bcc4d882fec64500d4049d9b1d1ae912ef348e89):

**Cross-port localhost / Vite dev server**

The original `u.Host == r.Host` check was indeed too strict for the dev setup. Verified via `vite.config.ts`: the proxy has no `changeOrigin` option, so Vite forwards the browser's `Origin: http://localhost:5173` header unchanged while setting `Host: localhost:8080` on the upstream request — the two would not match.

`checkOrigin` is now:

```go
if isLoopbackHost(u.Host) {
    return true  // permit cross-port loopback (Vite dev server :5173 → backend :8080)
}
return u.Host == r.Host
```

This matches the CORS policy's `isLocalhostOrigin` logic: any loopback origin is allowed, non-loopback origins must match the request host exactly. The comment explains both the rationale and the Vite dev proxy assumption. The test case is renamed from `"cross-port same domain rejected"` to `"cross-port loopback allowed (Vite dev server proxy)"` with `want: true`.

**ANSI injection — full sequence strip**

Switched from `/\x1b/g` to `/\x1b\[[0-9;]*[A-Za-z]/g` as suggested. `\x1b[1mBold\x1b[0m` now produces `Bold` instead of `[1mBold[0m`. The test was updated to assert both that no bracket remnants appear (checking `[1m` specifically) and that the template's own `\x1b[31m`/`\x1b[0m` sequences are still present.

internal/ws/handler.go
tomo-chan: Good catch — the description was incorrect about this. Vite does **not** strip the `Origin` header by default (no `changeOrigin` in `vite.config.ts`), so the request would have reached the backend as `Origin: http://localhost:5173` / `Host: localhost:8080` and been rejected by the original `u.Host == r.Host` guard.

Fixed in https://github.com/tomo-chan/panemux/commit/bcc4d882fec64500d4049d9b1d1ae912ef348e89: `checkOrigin` now calls `isLoopbackHost(u.Host)` first and returns `true` for any loopback origin regardless of port. Non-loopback origins still get the strict same-host check. The function-level comment now explicitly documents both the Vite dev proxy case and the empty-origin allowance.

frontend/src/hooks/useTerminal.ts
```diff
@@ -58,7 +58,8 @@ export function useTerminal({ sessionId, container, editMode = false }: UseTermi
             term.write('\r\n\x1b[2m[Session ended]\x1b[0m\r\n')
           }
         } else if (msg.type === 'error') {
-          term.write(`\r\n\x1b[31mError: ${msg.message}\x1b[0m\r\n`)
+          const safeMsg = msg.message.replace(/\x1b\[[0-9;]*[A-Za-z]/g, '')
+          term.write(`\r\n\x1b[31mError: ${safeMsg}\x1b[0m\r\n`)
```
tomo-chan, line 62: The pattern covers standard CSI sequences (`\x1b[Nn;NmX`) but not private-use variants like `\x1b[?25l` (where `?` is not in `[0-9;]`) or non-CSI sequences like OSC (`\x1b]0;title\x07`). In practice backend error messages won't contain those, but worth confirming whether the intent was "strip all ANSI" or "strip the common subset."

tomo-chan (commented):
## Code Review (follow-up)

Both issues raised in the previous review have been cleanly addressed.

### Resolved

**Cross-port localhost / Vite dev server** — The root cause was correctly identified: Vite forwards the browser's `Origin` header unchanged, so the original `u.Host == r.Host` guard would have rejected dev-mode connections. The fix (`isLoopbackHost` check before the same-host fallback) matches the CORS middleware's policy exactly, the function-level comment explains both the Vite dev proxy assumption and the empty-origin rationale, and the test case is renamed from "rejected" to "allowed" so the intent is unambiguous. ✅

**ANSI injection — full sequence strip** — Switching from `/\x1b/g` to `/\x1b\[[0-9;]*[A-Za-z]/g` correctly removes the full CSI sequence rather than leaving the parameter suffix as visible text. The updated test asserts both that `[1m` (the remnant) is gone and that the outer template sequences survive. ✅

---

### Issues & Questions

**[Minor question] ANSI regex does not cover private-use or non-CSI sequences**

```typescript
const safeMsg = msg.message.replace(/\x1b\[[0-9;]*[A-Za-z]/g, '')
```

The regex matches standard CSI sequences (`\x1b[`, optional `[0-9;]` params, letter final byte). Two categories of escape sequences fall outside it:

1. **Private-use CSI sequences** (`\x1b[?25l`, `\x1b[>1h`, etc.) — the DEC private markers `?`, `>`, `<` are not in `[0-9;]`, so the pattern does not match and these sequences would pass through to xterm.js.
2. **Non-CSI sequences** such as OSC (`\x1b]0;…\x07`), APC, or a bare `\x1b` not followed by `[` — these are unaffected by the regex.

In practice, backend error messages are unlikely to contain private-use or OSC sequences, so the risk is very low. If you want tighter coverage, something like `/\x1b[\x40-\x5f][^\x40-\x7e]*[\x40-\x7e]/g` (matches any C1/Fe/Fs/Fp escape sequence plus two-character escapes) would cover the broader set. Not blocking — just flagging in case the original threat model assumed "strip all ANSI" rather than "strip standard CSI."

---

### Overall

Both substantive issues are resolved and the follow-up commit improves the code beyond just fixing the bugs (better test assertions, better doc comment). The remaining minor question is about completeness of the ANSI regex, which is low-risk and can be addressed in a follow-up if desired. Ready to merge otherwise.

tomo-chan (commented):
## CI Findings

### ❌ Test (Go) failing on commit `bcc4d882`

The CI run against the latest commit ([run #25589257508](https://github.com/tomo-chan/panemux/actions/runs/25589257508)) fails at the **Test (Go)** step. All subsequent steps (Test Frontend, Coverage, E2E) were skipped.

The CI command is:
```
go test -v ./... -race 2>&1 | go-junit-report -set-exit-code > test-results/go-junit.xml
```

All output is piped directly into `go-junit-report` and redirected to the XML file — no test output appears in the job log, so the specific failing test(s) cannot be identified from the CI logs alone. There are no uploaded artifacts either.

**What's worth checking:**

The test plan for the follow-up commit covers only a targeted subset without the `-race` flag:
```
go test ./internal/ws/... -run TestCheckOrigin
```

The CI adds `-race` and runs the full `./...` suite. If the failure is a data race detected by the race detector, it would not have been caught by the local run above. A quick way to reproduce locally:

```sh
go test -v ./... -race
```

If you can identify the specific failing test and output from your local run, updating the PR with that information (and the fix) would unblock the merge.

---

### ⚠️ Node.js 20 deprecation in pinned action SHAs

The CI warning (already present on `main`, not introduced by this PR) is worth addressing before the deadline:

```
actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5   ← Node.js 20
actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff   ← Node.js 20
actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 ← Node.js 20
```

These SHAs correspond to the `v4`/`v5` floating tags respectively, which still use Node.js 20 as their runtime. GitHub will **force Node.js 24** on June 2, 2026 (~24 days from now), and Node.js 20 will be removed from runners entirely in September 2026.

The fix is to re-pin to the equivalent `v6.x` tags (which use Node.js 24) following the existing SHA-pinning convention in AGENTS.md. This is a pre-existing issue not introduced by this PR; filing a follow-up `ci:` ticket or handling it in a separate PR would be appropriate.

tomo-chan (commented):
**ANSI regex — private-use CSI** — Fixed in [f0d00aa](https://github.com/tomo-chan/panemux/commit/f0d00aaca53a8272ae4cdc3cc0091649d48a4faa).

Extended the parameter character class from `[0-9;]` to `[0-9;?<>!]` to cover DEC private markers:

```typescript
const safeMsg = msg.message.replace(/\x1b\[[0-9;?<>!]*[A-Za-z]/g, '')
```

This strips `\x1b[?25l` (hide cursor), `\x1b[>1h` (DEC private), and similar private-use sequences. A dedicated test case for these sequences was added alongside.

**CI** — Ran `go test ./... -race` locally (same flags as CI). All 7 packages pass. The single failure in run [#25589257508](https://github.com/tomo-chan/panemux/actions/runs/25589257508) appears to be transient — two other CI runs at the same commit `bcc4d882` succeeded at the same timestamp. No data race reproduced locally.

tomo-chan (commented):
**Test (Go) failure** — Reproduced the CI conditions locally with `go test ./... -race`. All 7 packages pass:

```
ok  panemux                1.178s
ok  panemux/internal/api   1.568s
ok  panemux/internal/config 1.508s
ok  panemux/internal/server 1.763s
ok  panemux/internal/session 1.759s
ok  panemux/internal/sshconfig 1.886s
ok  panemux/internal/ws    2.339s
```

GitHub's API shows three CI runs against commit `bcc4d882`: two succeeded, one failed at the same timestamp. No data race reproduced locally — the single failure appears to be transient infrastructure noise.

**Node.js 20 deprecation** — Acknowledged as a pre-existing issue not introduced by this PR. Will open a separate `ci:` PR to re-pin the three actions to their Node.js 24 equivalent SHAs ahead of the June 2 forced-upgrade deadline.

frontend/src/hooks/useTerminal.ts
tomo-chan: Fixed in [f0d00aa](https://github.com/tomo-chan/panemux/commit/f0d00aaca53a8272ae4cdc3cc0091649d48a4faa). Extended the parameter character class from `[0-9;]` to `[0-9;?<>\!]` to cover DEC private markers (`?`, `>`, `<`, `\!`):

```typescript
const safeMsg = msg.message.replace(/\x1b\[[0-9;?<>\!]*[A-Za-z]/g, '')
```

This strips `\x1b[?25l` (hide cursor), `\x1b[>1h` (DEC private mode set), and similar private-use CSI sequences. Non-CSI sequences like OSC (`\x1b]`) are not stripped — the intent is to neutralise injected CSI sequences from error text, not to implement a full ANSI parser. A dedicated test case covering private-use sequences was added alongside.

tomo-chan (commented):
## Code Review (final pass)

All prior concerns are resolved and CI is green on `f0d00aac`. 

### Resolved (this pass)

**[Minor question] Private-use CSI sequences** — Fixed in `f0d00aac`. The regex is now:

```typescript
const safeMsg = msg.message.replace(/\x1b\[[0-9;?<>\!]*[A-Za-z]/g, '')
```

The expanded character class covers DEC private markers (`?`, `>`, `<`) and the most common intermediate byte (`\!`), stripping `\x1b[?25l` and `\x1b[>1h` as confirmed by the new dedicated test case. ✅

**[Question] CI test failure** — Author reproduced locally with `go test ./... -race` (all 7 packages pass) and confirmed the single failure was transient infrastructure noise; two other CI runs against the same commit succeeded. Current CI run against `f0d00aac` is fully green. ✅

**[Suggestion] Node.js 20 deprecation** — Acknowledged as pre-existing. A separate `ci:` PR to re-pin the three actions to Node.js 24-compatible SHAs is planned ahead of the June 2 forced-upgrade deadline. ✅

---

### Overall

The PR is in excellent shape: all substantive security issues are addressed, the test suite is thorough, and CI is clean. Ready to merge.

PR #82, opened 2026-05-09
feat: expand approval prompt notifications

tomo-chan:
## Summary
- expand approval prompt detection beyond specific MCP server and tool names
- switch to the relevant workspace when a browser attention notification is clicked
- document the broader approval prompt detection and notification click behavior

## Test Plan
- [x] make lint-frontend
- [x] make test-frontend

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,9 +114,10 @@ Use edit mode when you want to:
 ### Notifications and attention prompts
 
 - panemux watches terminal output for agent confirmation prompts such as approval or proceed requests.
+- Codex permission menus such as MCP tool allow prompts are detected too.
 - When one is detected, the pane frame is highlighted.
 - If the pane belongs to an inactive workspace, that workspace tab is highlighted too.
-- Browser notifications are also shown after notification permission has been granted.
+- Browser notifications are also shown after notification permission has been granted, and clicking one switches to the matching workspace.
 - Notification permission is requested on the first browser interaction, instead of waiting for the first prompt.
 
 ### Choosing pane types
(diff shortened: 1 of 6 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Notification mock refactoring**: Replacing the bare `vi.fn()` stub with a `notificationInstance` that captures `onclick` and `close` gives tests direct access to the notification's event handler — a clean, targeted approach that makes the click-behavior assertion straightforward.
- **Backward-compatible API extension**: Adding `onClick?: () => void` as an optional third argument to `showBrowserNotification` requires zero changes to existing call sites and keeps the function's interface minimal.
- **Regex test coverage breadth**: Three test cases for the new pattern — a named MCP server/tool pair, a different MCP server/tool pair, and a generic non-MCP allow menu — effectively document that detection is intentionally not locked to specific tool or server names, and cover meaningful input variants.
- **Documentation accuracy**: Both `README.md` and `docs/behavior.md` are updated in the same change, reflecting the actual new behaviour (notification click → workspace switch) without drifting.

---

### Issues & Questions

**[Minor question] Unhandled promise in notification `onclick`**

In `App.tsx` (inside the `onAttention` callback):

```tsx
workspace ? () => {
  window.focus()
  void setActiveWorkspace(workspace.id)
} : undefined,
```

`void` discards the promise returned by `setActiveWorkspace`. If the call fails (e.g. a transient network error), the failure is silently swallowed — the notification closes and the workspace never switches, with no feedback to the user. Is this intentional? If so, a one-line comment (e.g. `// fire-and-forget; failure is non-critical`) would signal intent to future readers. Alternatively, appending `.catch(console.error)` would at least surface failures in the browser console.

---

### Overall

Clean, well-tested PR. The regex extension, notification click handler, and documentation are all consistent with one another. The only thing worth confirming is whether silent error-swallowing in the notification click is intentional. No blockers.

tomo-chan:
Addressed in `1f546c7c78f984092d5f850bf79c0aff5603eaf6`.

Notification-click workspace switches now use `setActiveWorkspace(workspace.id).catch(console.error)` so a failed switch is surfaced in the browser console instead of being silently dropped.

Local verification:
- `make lint-frontend`
- `make test-frontend`

tomo-chan:
Commit URL: https://github.com/tomo-chan/panemux/commit/1f546c7c78f984092d5f850bf79c0aff5603eaf6

This is the follow-up commit that handles notification-click workspace switch failures by logging rejected `setActiveWorkspace(...)` calls to the browser console.

tomo-chan (commented):
## Code Review (follow-up)

**[Minor question] Unhandled promise in notification `onclick`** — addressed in `1f546c7c`.

`void setActiveWorkspace(...)` has been replaced with `setActiveWorkspace(...).catch(console.error)`, and a dedicated test (`'logs workspace switch failures from notification clicks'`) was added to verify the failure path. Both are exactly right: the error is surfaced in the browser console without crashing the click handler, and the test asserts the exact error object rather than just that `console.error` was called at all.

---

### What's Good

- **Error-path test is tight**: `mockSetActiveWorkspace.mockRejectedValueOnce(switchError)` + `expect(consoleErrorSpy).toHaveBeenCalledWith(switchError)` pins the exact rejection to the exact log call — no loose matchers.
- **Happy-path now explicit**: adding `mockSetActiveWorkspace.mockResolvedValue(undefined)` to the click test removes the implicit `undefined` return and makes the promise resolution explicit, which is cleaner now that `setActiveWorkspace` is async in the component.
- **`await Promise.resolve()` flush**: correctly advances the micro-task queue to let the `.catch()` run before the assertion, which is the idiomatic way to flush a single promise tick in jsdom tests.

---

### Overall

Previous concern resolved cleanly. No new issues. Ready to merge.

PR #84, opened 2026-05-09
fix: tone down terminal scrollbar chrome

tomo-chan:
## Summary
- style the xterm viewport scrollbar to match the terminal chrome instead of hiding it
- load a dedicated terminal stylesheet after xterm.css
- add a Playwright regression test that verifies the scrollbar theme rules in a real browser

## Test Plan
- [x] make lint-frontend
- [x] make test-frontend
- [x] make test-e2e

### DIFF
--- a/frontend/e2e/terminal-scrollbar.spec.ts
+++ b/frontend/e2e/terminal-scrollbar.spec.ts
@@ -1,17 +1,13 @@
 import { expect, test } from '@playwright/test'
 
-test.use({
-  launchOptions: {
-    args: ['--show-scrollbars'],
-  },
-})
-
 test('styles the xterm viewport scrollbar to match the terminal chrome', async ({ page }) => {
   await page.goto('/')
 
   const viewport = page.locator('.xterm-viewport').first()
   await expect(viewport).toBeVisible()
 
+  // This comes from xterm itself, so keep it as a sanity check that the terminal
+  // viewport is initialized before asserting the theme rules added by this change.
   const overflowY = await viewport.evaluate((el) => {
     const element = el as HTMLElement
     const style = window.getComputedStyle(element)
@@ -83,4 +79,7 @@ test('styles the xterm viewport scrollbar to match the terminal chrome', async (
       }),
     ]),
   )
+
+  // Pseudo-element hover state is not asserted here because Playwright cannot
+  // reliably force hover on the scrollbar thumb across engines.
 })
--- a/frontend/e2e/terminal-scrollbar.spec.ts
+++ b/frontend/e2e/terminal-scrollbar.spec.ts
@@ -1,0 +1,86 @@
+import { expect, test } from '@playwright/test'
+
+test.use({
+  launchOptions: {
+    args: ['--show-scrollbars'],
+  },
+})
+
+test('styles the xterm viewport scrollbar to match the terminal chrome', async ({ page }) => {
+  await page.goto('/')
+
+  const viewport = page.locator('.xterm-viewport').first()
+  await expect(viewport).toBeVisible()
+
+  const overflowY = await viewport.evaluate((el) => {
+    const element = el as HTMLElement
+    const style = window.getComputedStyle(element)
+    return style.overflowY
+  })
+
+  const matchingRules = await page.evaluate(() => {
+    const matches: Array<{
+      selectorText: string
+      width: string
+      height: string
+      background: string
+      border: string
+      borderRadius: string
+      scrollbarWidth: string
+      scrollbarColor: string
+    }> = []
+
+    for (const sheet of Array.from(document.styleSheets)) {
+      let rules: CSSRuleList
+      try {
+        rules = sheet.cssRules
+      } catch {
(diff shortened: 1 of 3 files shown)


frontend/e2e/terminal-scrollbar.spec.ts
```diff
@@ -0,0 +1,86 @@
+import { expect, test } from '@playwright/test'
+
+test.use({
+  launchOptions: {
+    args: ['--show-scrollbars'],
```
tomo-chan: Does this flag affect what the test verifies? The assertions only introspect `document.styleSheets` and read a computed `overflow-y` value — neither is influenced by scrollbar visibility. If the flag is not needed, removing it avoids a latent dependency on a Chrome-only launch argument.

tomo-chan (commented):
## Code Review

### What's Good

- **Cross-browser scrollbar coverage**: The CSS correctly pairs the W3C Scrollbars Level 1 properties (`scrollbar-width`, `scrollbar-color`) for Firefox/Chrome 121+ with the `-webkit-scrollbar-*` pseudo-elements for older WebKit. Neither approach is omitted.
- **No `\!important` needed**: Using `.xterm .xterm-viewport` is specific enough to override xterm.css's own `.xterm-viewport` selector, which is the right way to layer on top of a third-party stylesheet.
- **CORS-safe stylesheet iteration**: The `try { rules = sheet.cssRules } catch { continue }` guard in the E2E test correctly skips cross-origin sheets that would throw a `SecurityError`, rather than letting the test crash.
- **Flexible assertions**: `expect.arrayContaining` + `expect.objectContaining` lets additional rules coexist without breaking the test — important since xterm.css itself contributes rules to the same `xterm-viewport` selector.

---

### Issues & Questions

**[Question] Does `--show-scrollbars` actually affect what this test verifies?**

The test introspects `document.styleSheets` to confirm CSS rules are loaded and reads the computed `overflow-y` value — it never interacts with a visible scrollbar or takes a screenshot. The `--show-scrollbars` Chrome flag forces overlay scrollbars to always be visible (useful on macOS where they hide by default), but that behavior doesn't influence `sheet.cssRules` or `getComputedStyle`. Is the flag load-bearing for something I'm missing, or is it left over from an earlier screenshot-based approach? If it's not needed, removing it avoids a silent dependency on a Chrome-only flag.

**[Minor question] `:hover` state defined in CSS but not asserted in the test**

`terminal.css` includes `.xterm .xterm-viewport::-webkit-scrollbar-thumb:hover { background: #52525b }` but there's no matching assertion in the spec. Is the hover rule intentionally left out of scope (since hover can't be reliably triggered on a pseudo-element in Playwright), or was it an oversight? A comment in the test noting the omission would help future maintainers.

**[Minor question] `overflowY` assertion tests xterm.js, not this PR's CSS**

```ts
expect(overflowY).toBe('scroll')
```

`overflow-y: scroll` is set by xterm.js itself on `.xterm-viewport`, not by `terminal.css`. This assertion is a useful sanity-check that xterm is initialised, but it doesn't verify anything introduced in this PR. A short comment — or grouping it with the `await expect(viewport).toBeVisible()` check above — would clarify the intent.

---

### Overall

Clean, minimal change with good cross-browser CSS and a solid introspection-based E2E strategy. The three items above are all low-stakes; none block merge. The `--show-scrollbars` question is the only one worth confirming before merging, since removing an unneeded Chrome flag is easier now than debugging a CI failure on a non-Chrome browser later.

tomo-chan:
Addressed in [`193b2b5`](https://github.com/tomo-chan/panemux/commit/193b2b540c592a766695da346605fd9b99f891b2).

- Removed the unused `--show-scrollbars` Playwright launch flag.
- Added a comment clarifying that `expect(overflowY).toBe('scroll')` is only a sanity check that xterm initialized before asserting this PR's scrollbar theme rules.
- Added a comment explaining why the `::-webkit-scrollbar-thumb:hover` rule is intentionally not asserted in Playwright.

Local verification re-run after the follow-up change:
- `make lint-frontend`
- `make test-frontend`
- `make test-e2e`

frontend/e2e/terminal-scrollbar.spec.ts
tomo-chan: Addressed in [`193b2b5`](https://github.com/tomo-chan/panemux/commit/193b2b540c592a766695da346605fd9b99f891b2).

I removed the `--show-scrollbars` launch flag because it was left over from an earlier approach and is not needed for the current assertions.

tomo-chan (commented):
## Code Review — Follow-up

All three findings from the previous review have been addressed in `193b2b5`:

- **`--show-scrollbars` removed** — the leftover Chrome launch flag is gone; the test is now engine-agnostic.
- **`overflowY` comment added** — the comment correctly explains this is an xterm initialisation sanity-check, not a test of the new CSS.
- **Hover-state omission documented** — the comment at the bottom of the test makes the intentional gap visible to future maintainers.

The diff is clean. Nothing new to raise.

PR #85, opened 2026-05-10
fix: reduce duplicate attention notifications

tomo-chan:
## Summary
- suppress browser attention notifications for prompts that are already visible to the user
- dedupe replayed browser notifications per pane across refreshes, reconnects, and maximize changes
- document the notification rules and expand factor-based frontend test coverage

## Test Plan
- [x] make test-frontend
- [x] make coverage-frontend
- [x] make lint-frontend
- [x] cd frontend && npx playwright test e2e/workspace-switch.spec.ts

### DIFF
--- a/frontend/src/hooks/useWorkspaceAttentionMonitor.ts
+++ b/frontend/src/hooks/useWorkspaceAttentionMonitor.ts
@@ -1,30 +1,37 @@
 import { useEffect, useMemo, useRef } from 'react'
 import { createAgentAttentionDetector } from '../utils/agentAttention'
+import { getLastNotifiedAttentionSignature, setLastNotifiedAttentionSignature } from '../utils/attentionNotificationState'
 import type { LayoutNode, WorkspacesResponse } from '../schemas'
-
-const ATTENTION_NOTIFY_INTERVAL_MS = 10_000
 
 interface UseWorkspaceAttentionMonitorOptions {
   workspaces: WorkspacesResponse | null
-  onAttention: (paneId: string) => void
+  maximizedPaneId: string | null
+  onAttention: (paneId: string, showBrowserNotification?: boolean) => void
 }
 
 interface PaneMonitorState {
   detector: ReturnType<typeof createAgentAttentionDetector>
   decoder: TextDecoder
-  lastAttentionAt: number | null
 }
 
-export function useWorkspaceAttentionMonitor({ workspaces, onAttention }: UseWorkspaceAttentionMonitorOptions) {
+export function useWorkspaceAttentionMonitor({ workspaces, maximizedPaneId, onAttention }: UseWorkspaceAttentionMonitorOptions) {
   const monitorStatesRef = useRef<Map<string, PaneMonitorState>>(new Map())
 
-  const paneIds = useMemo(() => {
-    if (!workspaces) return []
+  const paneMetadataById = useMemo(() => {
+    const metadata = new Map<string, { workspaceId: string }>()
+    if (!workspaces) return metadata
 
-    return workspaces.items.flatMap((workspace) => collectPaneIDs(workspace.layout))
+    for (const workspace of workspaces.items) {
+      for (const paneId of collectPaneIDs(workspace.layout)) {
+        metadata.set(paneId, { workspaceId: workspace.id })
+      }
+    }
+
+    return metadata
   }, [workspaces])
 
   useEffect(() => {
+    const paneIds = [...paneMetadataById.keys()]
     if (paneIds.length === 0) return
 
     const sockets = paneIds.map((paneId) => {
@@ -33,7 +40,6 @@ export function useWorkspaceAttentionMonitor({ workspaces, onAttention }: UseWor
       ws.binaryType = 'arraybuffer'
 
       ws.onopen = () => {
-        state.lastAttentionAt = null
         state.detector.reset()
         state.decoder = new TextDecoder()
       }
@@ -41,9 +47,22 @@ export function useWorkspaceAttentionMonitor({ workspaces, onAttention }: UseWor
       ws.onmessage = (event) => {
         if (typeof event.data === 'string') return
         const text = state.decoder.decode(new Uint8Array(event.data as ArrayBuffer), { stream: true })
-        if (shouldNotifyAttention(state, text)) {
-          onAttention(paneId)
+        const attentionMatch = state.detector.feed(text)
+        if (!attentionMatch) return
+
+        const shouldNotifyBrowser = shouldNotifyBrowserAttention({
+          paneId,
+          paneWorkspaceId: paneMetadataById.get(paneId)?.workspaceId ?? null,
+          activeWorkspaceId: workspaces?.active ?? null,
+          maximizedPaneId,
+          browserIsActive: isBrowserActive(),
+          signature: attentionMatch.signature,
+        })
+
+        if (shouldNotifyBrowser) {
+          setLastNotifiedAttentionSignature(paneId, attentionMatch.signature)
         }
+        onAttention(paneId, shouldNotifyBrowser)
       }
 
       ws.onerror = () => {
@@ -60,7 +79,7 @@ export function useWorkspaceAttentionMonitor({ workspaces, onAttention }: UseWor
         socket.close()
       }
     }
-  }, [onAttention, paneIds])
+  }, [maximizedPaneId, onAttention, paneMetadataById, workspaces])
 }
 
 function buildWebSocketURL(paneId: string): string {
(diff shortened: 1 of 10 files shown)


frontend/src/hooks/useWorkspaceAttentionMonitor.ts
```diff
@@ -60,7 +79,7 @@ export function useWorkspaceAttentionMonitor({ workspaces, onAttention }: UseWor
         socket.close()
       }
     }
-  }, [onAttention, paneIds])
+  }, [maximizedPaneId, onAttention, paneMetadataById, workspaces])
 }
```
tomo-chan, line 93: [Question] Adding `maximizedPaneId` here causes all monitoring WebSocket connections to close and reopen every time the user toggles maximize. The value only needs to be read at message-receive time, not at socket-creation time — a `useRef` + shallow sync effect would let the message handler sample the current value without restarting sockets:

```ts
const maximizedPaneIdRef = useRef(maximizedPaneId)
useEffect(() => { maximizedPaneIdRef.current = maximizedPaneId }, [maximizedPaneId])
```

Is the reconnect-on-maximize intentional?

frontend/src/utils/agentAttention.ts
```diff
@@ -19,18 +23,37 @@ export function createAgentAttentionDetector(): AgentAttentionDetector {
...
+}
+
+function normalizeAttentionText(value: string): string {
+  return stripAnsi(value)
+    .replace(/\s+/g, ' ')
+    .trim()
+    .toLocaleLowerCase()
+}
+
 function stripAnsi(value: string): string {
```
tomo-chan, line 57: [Minor question] `toLocaleLowerCase()` is locale-dependent — in a Turkish locale, "I" lowercases to "ı" rather than "i". Signatures are persisted in `localStorage` and compared across sessions, so a locale change between writes and reads could silently defeat dedup. `toLowerCase()` is locale-invariant and safer for a stored lookup key. Intentional choice?

tomo-chan (commented):
## Code Review

### What's Good

- The behavioral decision table in `docs/behavior.md` is excellent — it makes the visibility-gating logic instantly comprehensible without having to read the code.
- Switching from a time-based throttle (10 s window) to content-aware signature-based dedupe is a better design: replays of identical prompt text after reconnect/refresh are suppressed, while genuinely new prompt text always notifies.
- Returning `AgentAttentionMatch | null` from `agentAttention.ts` instead of `boolean` cleanly separates detection from deduplication — the signature is computed exactly once at detection time, not re-derived later.
- `attentionNotificationState.ts` gracefully falls back to in-memory state when `localStorage` throws, and the new test verifies that path end-to-end.
- The `onAttention` call-through-with-flag design (always call, but pass `showBrowserNotification: false` for duplicates) is better than the old silent-drop: pane/workspace highlights still update on replay, only the browser popup is suppressed.
- Test factor coverage is thorough — the matrix now explicitly covers browser-active/inactive × pane-visible/hidden × same/different prompt × localStorage-available/unavailable.

---

### Issues & Questions

**[Question] `maximizedPaneId` in `useEffect` deps causes WebSocket reconnects on maximize toggle**

`maximizedPaneId` is now in the dep array, so toggling maximize closes and reopens all monitoring WebSocket connections across every pane. This is new behaviour relative to before this PR — previously, only layout/workspace structural changes caused reconnects.

The current value of `maximizedPaneId` only needs to be sampled at message-receive time, not at socket-creation time, so it could live in a ref without being a dep:

```ts
const maximizedPaneIdRef = useRef(maximizedPaneId)
useEffect(() => { maximizedPaneIdRef.current = maximizedPaneId }, [maximizedPaneId])
```

Similarly, `workspaces?.active` is now read inside the effect (for `activeWorkspaceId`) and `workspaces` is explicitly in the dep array. This means every workspace switch (changing `.active`) also forces a full socket teardown/reconnect. The old code avoided this because `paneIds` was computed from `workspaces.items` only, not `.active`. Is the reconnect-on-maximize an intentional tradeoff, or a side effect worth addressing with a ref?

**[Minor question] `toLocaleLowerCase()` for a persisted dedup key**

Signatures are written to `localStorage` and read back in future sessions. `toLocaleLowerCase()` is locale-dependent (Turkish locale: "I" → "ı", not "i"). If a user's browser locale changes between sessions, the same prompt text could produce a different signature and bypass cross-session dedupe. `toLowerCase()` is locale-invariant and safer for a stored lookup key. Is the locale-aware variant intentional here?

**[Minor question] No mechanism to clear persisted signatures**

Stored signatures are written on notification and never cleared — no TTL, no clear-on-dismiss. If an agent presents the exact same prompt text twice in separate sessions (e.g., the same MCP-tool approval prompt appearing again days later), the browser notification is suppressed the second time. The pane and workspace highlights still fire, so the user isn't completely unaware, but they won't get the tab-switch notification if the window is in the background. Is this by design — trusting users to notice the frame highlight — or is a TTL or clear-on-attention-resolved strategy planned?

---

### Overall

The core logic is solid and the test coverage is thorough. Two small concerns worth confirming before merge: whether reconnects on maximize toggle are an acceptable tradeoff (or worth addressing with a ref), and whether `toLocaleLowerCase` is correct for a cross-session stored key. Neither blocks correctness — the feature behaves as designed.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/5d210e1d22a5ae1f53a04a0dbfa56e328aa02e6f.

- The monitor no longer reconnects its background WebSockets on maximize toggle or active-workspace changes. `useWorkspaceAttentionMonitor` now keeps `activeWorkspaceId` and `maximizedPaneId` in refs and samples them at message-receive time, so only pane-set changes recreate the sockets.
- The persisted dedupe key now uses `toLowerCase()` instead of `toLocaleLowerCase()` to avoid locale-dependent cross-session drift.
- I added a regression test that rerenders the hook across active-workspace and maximize changes and asserts that no extra monitor sockets are created.

For the persisted-signature lifetime: leaving it without a TTL is intentional in this PR. The current rule is “only the last browser-notified prompt per pane is remembered until a different prompt replaces it.” That matches the earlier product decision to suppress replay noise across refresh/reconnect without growing unbounded history. Visual pane/workspace attention still reappears even when the browser popup is suppressed.

Local verification run after the follow-up change:
- `npm test -- --run src/hooks/useWorkspaceAttentionMonitor.test.ts src/utils/agentAttention.test.ts`
- `make lint-frontend`
- `make test-frontend`

frontend/src/hooks/useWorkspaceAttentionMonitor.ts
tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/5d210e1d22a5ae1f53a04a0dbfa56e328aa02e6f.

I moved both `activeWorkspaceId` and `maximizedPaneId` behind refs and keep them synced with tiny effects, so the message handler samples current visibility state without tearing down monitor sockets. The socket-creation effect now depends only on the pane set and callback, and I added a regression test that rerenders across active-workspace and maximize changes and asserts no extra monitor sockets are created.

Local verification after the change:
- `npm test -- --run src/hooks/useWorkspaceAttentionMonitor.test.ts src/utils/agentAttention.test.ts`
- `make lint-frontend`
- `make test-frontend`

frontend/src/utils/agentAttention.ts
tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/5d210e1d22a5ae1f53a04a0dbfa56e328aa02e6f.

I switched the persisted dedupe signature normalization from `toLocaleLowerCase()` to `toLowerCase()` so the stored key is locale-invariant across sessions.

On the persisted-signature lifetime: leaving it without a TTL is intentional in this PR. The current rule is that each pane remembers only the last browser-notified prompt signature until a different prompt replaces it. That keeps refresh/reconnect replay noise suppressed without accumulating unbounded history, while pane/workspace visual attention still reappears even when the browser popup is suppressed.

Local verification after the change:
- `npm test -- --run src/hooks/useWorkspaceAttentionMonitor.test.ts src/utils/agentAttention.test.ts`
- `make lint-frontend`
- `make test-frontend`

tomo-chan (commented):
## Code Review — Follow-up (commit 5d210e1d)

Revisiting after the author addressed the two prior comments.

### Resolved

**[Question]  in  deps → ✅ resolved**

Both `activeWorkspaceId` and `maximizedPaneId` are now behind refs synced by lightweight effects, and the socket-creation effect dep array is trimmed to `[onAttention, paneMetadataById]`. The `paneMetadataById` useMemo also switched to `[workspaces?.items]` so that only layout-structure changes (not active-workspace switches) recalculate the map and restart sockets. The new regression test — "does not reconnect monitor sockets when only active workspace or maximize state changes" — exercises a workspace switch and a maximize toggle in the same render cycle and asserts `MockWebSocket.instances` stays at 3, which is exactly the right signal.

**[Minor question] `toLocaleLowerCase()` → ✅ resolved**

Replaced with `toLowerCase()` in `normalizeAttentionText`. Signatures stored across sessions are now locale-invariant.

**[Minor question] No mechanism to clear persisted signatures → acknowledged by design**

The author confirmed the one-signature-per-pane rule is intentional: the stored signature is replaced when a new (different) prompt fires, so the history stays bounded at one entry per pane while pane/workspace visual attention still reappears even when the browser popup is suppressed. Reasonable tradeoff; no further action needed in this PR.

---

### Overall

All prior concerns addressed. The ref-based approach is idiomatic and the regression test provides strong evidence that maximize toggles and workspace switches no longer teardown monitoring sockets. LGTM.

PR #87, opened 2026-05-13
fix: suppress replay-generated terminal input

tomo-chan:
## Summary
- add explicit WebSocket replay lifecycle frames around buffered session snapshots
- suppress xterm stdin only while replayed bytes are being re-applied so replayed terminal queries do not become shell input
- cover the replay protocol and suppression behavior in frontend/backend tests and document the protocol update

## Test plan
- [x] npm test -- src/hooks/useTerminal.test.ts src/schemas/index.test.ts
- [x] make test-frontend
- [x] go test ./internal/ws
- [x] make lint-go
- [x] make test-go

### DIFF
--- a/.github/workflows/model-check.yml
+++ b/.github/workflows/model-check.yml
@@ -52,7 +52,7 @@ jobs:
             out_dir="$(mktemp -d)"
             log_file="$(mktemp)"
             receipt_file="$out_dir/receipt.json"
-            java -jar /tmp/alloy/org.alloytools.alloy.dist.jar exec -t none -o "$out_dir" -f "$model" | tee "$log_file"
+            java -jar /tmp/alloy/org.alloytools.alloy.dist.jar exec -t none -o "$out_dir" -f "$model" 2>&1 | tee "$log_file"
             python3 - "$model" "$log_file" "$receipt_file" <<'PY'
           import re
           import json
--- a/.github/workflows/model-check.yml
+++ b/.github/workflows/model-check.yml
@@ -1,0 +1,76 @@
+name: model-check
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'docs/models/**'
+      - '.github/workflows/model-check.yml'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'docs/models/**'
+      - '.github/workflows/model-check.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  alloy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+
+      - name: Download Alloy
+        run: |
+          set -euo pipefail
+          mkdir -p /tmp/alloy
+          curl -fsSL \
+            -o /tmp/alloy/org.alloytools.alloy.dist.jar \
+            https://github.com/AlloyTools/org.alloytools.alloy/releases/download/v6.2.0/org.alloytools.alloy.dist.jar
+
+      - name: Validate Alloy models
+        run: |
+          set -euo pipefail
+          shopt -s nullglob
+          models=(docs/models/*.als)
+          if [ "${#models[@]}" -eq 0 ]; then
+            echo "No Alloy models found under docs/models"
+            exit 1
+          fi
+
+          for model in "${models[@]}"; do
+            echo "::group::${model}"
+            java -jar /tmp/alloy/org.alloytools.alloy.dist.jar commands "$model"
+            out_dir="$(mktemp -d)"
+            log_file="$(mktemp)"
+            java -jar /tmp/alloy/org.alloytools.alloy.dist.jar exec -t none -o "$out_dir" -f "$model" | tee "$log_file"
+            python3 - "$model" "$log_file" <<'PY'
+          import re
+          import sys
+          from pathlib import Path
+
+          model = sys.argv[1]
+          log_path = Path(sys.argv[2])
+          result_lines = []
+          for line in log_path.read_text().splitlines():
+              if re.match(r"^\d+\.\s+check\b", line):
+                  result_lines.append(line)
+
+          if not result_lines:
+              raise SystemExit(f"{model}: no Alloy check results found")
+
+          failures = [line for line in result_lines if "UNSAT" not in line]
+          if failures:
+              joined = "\n".join(failures)
+              raise SystemExit(f"{model}: expected all checks to be UNSAT, got:\n{joined}")
+          PY
+            echo "::endgroup::"
+          done
--- a/.github/workflows/model-check.yml
+++ b/.github/workflows/model-check.yml
@@ -51,26 +51,50 @@ jobs:
             java -jar /tmp/alloy/org.alloytools.alloy.dist.jar commands "$model"
             out_dir="$(mktemp -d)"
             log_file="$(mktemp)"
+            receipt_file="$out_dir/receipt.json"
             java -jar /tmp/alloy/org.alloytools.alloy.dist.jar exec -t none -o "$out_dir" -f "$model" | tee "$log_file"
-            python3 - "$model" "$log_file" <<'PY'
+            python3 - "$model" "$log_file" "$receipt_file" <<'PY'
           import re
+          import json
           import sys
           from pathlib import Path
 
           model = sys.argv[1]
           log_path = Path(sys.argv[2])
-          result_lines = []
-          for line in log_path.read_text().splitlines():
-              if re.match(r"^\d+\.\s+check\b", line):
-                  result_lines.append(line)
+          receipt_path = Path(sys.argv[3])
+          receipt = json.loads(receipt_path.read_text())
+          expected_checks = sum(
+              1 for command in receipt.get("commands", {}).values()
+              if command.get("type") == "check"
+          )
+          if expected_checks == 0:
+              raise SystemExit(f"{model}: no Alloy check commands found in receipt")
 
-          if not result_lines:
-              raise SystemExit(f"{model}: no Alloy check results found")
+          ansi_pattern = re.compile(r"\x1b\[[0-9;]*[A-Za-z]")
+          control_pattern = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
+          verdict_pattern = re.compile(r"\b(UNSAT|SAT|UNKNOWN)\b")
 
-          failures = [line for line in result_lines if "UNSAT" not in line]
-          if failures:
-              joined = "\n".join(failures)
+          unsat_lines = []
+          bad_lines = []
+          for raw_line in log_path.read_text().splitlines():
+              normalized = control_pattern.sub("", ansi_pattern.sub("", raw_line))
+              verdict = verdict_pattern.search(normalized)
+              if verdict is None:
+                  continue
+              if verdict.group(1) == "UNSAT":
+                  unsat_lines.append(normalized)
+              else:
+                  bad_lines.append(normalized)
+
+          if bad_lines:
+              joined = "\n".join(bad_lines)
               raise SystemExit(f"{model}: expected all checks to be UNSAT, got:\n{joined}")
+
+          if len(unsat_lines) != expected_checks:
+              joined = "\n".join(unsat_lines)
+              raise SystemExit(
+                  f"{model}: expected {expected_checks} UNSAT results, got {len(unsat_lines)}\n{joined}"
+              )
           PY
             echo "::endgroup::"
           done
--- a/.github/workflows/model-check.yml
+++ b/.github/workflows/model-check.yml
@@ -24,6 +24,8 @@ permissions:
 jobs:
   alloy:
     runs-on: ubuntu-latest
+    env:
+      ALLOY_JAR_SHA256: 6b8c1cb5bc93bedfc7c61435c4e1ab6e688a242dc702a394628d9a9801edb78d
     steps:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
@@ -35,6 +37,7 @@ jobs:
           curl -fsSL \
             -o /tmp/alloy/org.alloytools.alloy.dist.jar \
             https://github.com/AlloyTools/org.alloytools.alloy/releases/download/v6.2.0/org.alloytools.alloy.dist.jar
+          echo "${ALLOY_JAR_SHA256}  /tmp/alloy/org.alloytools.alloy.dist.jar" | shasum -a 256 -c -
 
       - name: Validate Alloy models
         run: |
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -220,7 +220,7 @@ The hook resets all replay fields on each WebSocket open so an interrupted repla
 connection cannot suppress stdin for the new connection.
 
 This lifecycle is also captured as an Alloy model in
-[replay_state.als](/tmp/panemux-replay-input-fix/docs/models/replay_state.als), which treats
+[replay_state.als](models/replay_state.als), which treats
 reconnect, replay frame delivery, replay-control write failure, socket close, and replay write
 completion as explicit transition events and checks the stale-suppression invariants over bounded
 traces.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -206,7 +206,7 @@ Replay state ownership in this hook:
 
 - `replayActive`: true between `replay:start` and `replay:end`
 - `replayWriteDepth`: count of replay `term.write(...)` calls whose callbacks have not fired yet
-- `replayEnded`: true once `replay:end` has been received for the current connection
+- `awaitingReplayEnd`: true after `replay:start` and false once `replay:end` has been received for the current connection
 - `disableStdin`: derived safety switch; forced on whenever replay is active or draining, forced off
   on reconnect reset and after the final replay write callback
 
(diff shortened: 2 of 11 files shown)


internal/ws/handler.go
```diff
@@ -230,3 +232,9 @@ func (h *Handler) sendStatus(conn *websocket.Conn, state string) {
 	data, _ := json.Marshal(msg)
 	_ = conn.WriteMessage(websocket.TextMessage, data)
 }
+
+func (h *Handler) sendReplay(conn *websocket.Conn, state string) {
+	msg := ControlMessage{Type: "replay", State: state}
+	data, _ := json.Marshal(msg)
+	_ = conn.WriteMessage(websocket.TextMessage, data)
+}
```
tomo-chan, line 249: This silently discards the write error. For `sendStatus` that is fine — a write failure closes the connection before anything else happens. For `sendReplay`, the error path matters: if `replay:start` is delivered but this call for `replay:end` silently fails, the frontend `TerminalEntry` keeps `replayActive: true` and `replayEnded: false` across the reconnect. See the main review body for the full scenario.

tomo-chan (commented):
## Code Review

### What's Good

- **`replayWriteDepth` counter design**: Tracking in-flight `xterm.write()` calls with a depth counter is exactly the right shape here. `write()` is async (callback-based), so a naive \"restore stdin on `replay:end`\" would race with bytes still being flushed to the renderer. The depth counter prevents that cleanly.
- **`replayEnded` flag**: Decoupling \"end marker received\" from \"all writes completed\" is sound. `maybeRestoreTerminalInput` being safe to call from either path (the `replay:end` handler and each write callback) without double-restoring is a subtle but correct invariant.
- **Three distinct test scenarios** in `useTerminal.test.ts`: basic suppression, buffered-remount-flush ordering, and live-output after replay. The buffered-remount test in particular catches a real edge case where the frontend receives replay frames before the container is mounted.
- **`TestWS_SkipsReplayFramesWhenSnapshotEmpty`**: Important negative test — the backend guard (`if len(snapshot) > 0`) is tested explicitly so a future refactor can't silently remove it and send empty replay pairs.
- **Schema-first discipline maintained**: Zod union updated before the new frame type is consumed in the hook, matching the project convention.
- **Docs updated in the same PR**: Both `behavior.md` (protocol spec) and `architecture.md` (hook rationale) are updated, keeping docs and code in sync.

---

### Issues & Questions

**[Question] `sendReplay` silently drops write errors — stale `replayActive` can leave stdin suppressed after reconnect**

`sendReplay` follows the same pattern as `sendStatus` — errors are silently discarded:

```go
func (h *Handler) sendReplay(conn *websocket.Conn, state string) {
    msg := ControlMessage{Type: "replay", State: state}
    data, _ := json.Marshal(msg)
    _ = conn.WriteMessage(websocket.TextMessage, data)  // error discarded
}
```

For `sendStatus` this is benign: a failed write closes the connection and both sides reset cleanly. For replay frames the consequence is sharper. `TerminalEntry` is keyed by `sessionId` in the module-level `terminalEntries` map and **persists across WebSocket reconnections**. If `sendReplay("start")` is delivered successfully but the connection dies before `sendReplay("end")` is sent, the frontend entry retains `replayActive: true` and `replayEnded: false`. On the next connection, if the snapshot is empty (`if len(snapshot) > 0` is false), no replay frames are sent. Live binary frames would then be processed by `writeTerminalBytes` with `replayWrite = true`, incrementing `replayWriteDepth`. Each write callback calls `maybeRestoreTerminalInput`, which sees `replayEnded: false` and leaves `disableStdin: true` — stdin stays suppressed indefinitely.

Two mitigations worth considering:
1. **Propagate `sendReplay` errors**: return a `bool` or `error` and treat a failed replay frame the same as a failed binary write — exit `forwardTerminalOutput` early so the connection is torn down before live output begins.
2. **Reset replay state on reconnect in the frontend**: clear `replayActive` and `replayEnded` in the WebSocket `onClose` / reconnect path so a new connection always starts from a clean state.

Is this an edge case you've considered and accepted as tolerable (given that a delivery failure implies the connection is already unusable, and a subsequent reconnect is likely to see the connection fail again before delivering live output), or would you prefer one of these guards?

**[Minor question] `replayEnded: true` as the default initial value**

The field name reads as "a replay has ended" rather than "no replay is pending." Since `maybeRestoreTerminalInput` short-circuits on `\!replayEnded`, the initial `true` value is semantically "don't block restoration" — which is correct but not obvious from the name alone. A name like `awaitingReplayEnd: false` (where `maybeRestoreTerminalInput` guards on `awaitingReplayEnd`) or `replayPending: false` would communicate the initial-state semantics more clearly to a future reader. Minor, no need to change before merge.

---

### Overall

The core mechanism — depth-counted async writes gated by two flags — is well-reasoned and the test suite covers the key behavioral scenarios. The one open question is whether silent errors in `sendReplay` need a guard against stale replay state surviving a mid-replay disconnect; everything else looks ready to merge.

frontend/src/hooks/useTerminal.ts
```diff
@@ -265,6 +279,9 @@ function getOrCreateTerminalEntry(sessionId: string): TerminalEntry {
     resizeTimers: new Set(),
     send: null,
     editMode: false,
+    replayActive: false,
+    replayWriteDepth: 0,
```
tomo-chan, line 285: Initialising to `true` is semantically "don't block stdin restoration" rather than "a replay has ended", which is not obvious from the name. Worth a rename to `awaitingReplayEnd: false` (and inverting the guard in `maybeRestoreTerminalInput`) before this pattern is copied elsewhere, though it's not a blocker.

docs/architecture.md
```diff
@@ -184,6 +195,33 @@ Why this hook:
...
+1. `live`: no replay pending, stdin enabled
+2. `replay pending end`: replay frames still arriving, stdin disabled
+3. `replay draining`: end marker received, but queued xterm writes still draining, stdin disabled
+
+The hook resets all replay fields on each WebSocket open so an interrupted replay from a previous
+connection cannot suppress stdin for the new connection.
+
+This lifecycle is also captured as an Alloy model in
+[replay_state.als](/tmp/panemux-replay-input-fix/docs/models/replay_state.als), which treats
+reconnect, replay frame delivery, and replay write completion as explicit transition events and
```
tomo-chan: This resolves to a path on the author's local machine. The model is committed at `docs/models/replay_state.als`, so the correct relative link from this file is `[replay_state.als](models/replay_state.als)`.

docs/behavior.md
```diff
@@ -261,8 +263,79 @@ Supported control messages:
...
+Alloy model:
+
+- The replay state machine above is mirrored in [replay_state.als](/tmp/panemux-replay-input-fix/docs/models/replay_state.als).
+- The model abstracts the frontend into three states: `Live`, `ReplayPendingEnd`, and `ReplayDraining`.
+- It checks these invariants:
+  - `Live` never leaves `disableStdin` enabled
+  - replay states always keep `disableStdin` enabled
+  - `ReplayDraining` is only reachable while replay writes remain queued
+  - `Reconnect` always resets the model to a clean `Live` state
+  - stale replay suppression cannot survive in `Live`
```
tomo-chan, line 342: Same worktree path. From `docs/behavior.md`, the correct relative link is `[replay_state.als](models/replay_state.als)`.

frontend/src/hooks/useTerminal.ts
```diff
@@ -265,6 +281,9 @@ function getOrCreateTerminalEntry(sessionId: string): TerminalEntry {
...
+    replayEnded: true,
   }
```
tomo-chan, line 287: Carried forward from previous review: `replayEnded: true` reads as 'a replay has ended' rather than 'no replay is pending'. Consider `awaitingReplayEnd: false` so the initial-state semantics are self-documenting. Not a blocker.

tomo-chan (commented):
## Code Review (round 2)

**[Question] `sendReplay` silently drops write errors → resolved**

`sendReplay` now returns `bool`, `forwardTerminalOutput` returns early on failure, and `resetReplayState` is called in `onOpen` so a reconnect always starts clean. Both paths are tested: `'resets stale replay suppression when the socket reconnects'` covers the frontend reconnect reset, and `TestWS_ReplayStartWriteFailureStopsBeforeReplayBody` verifies `writeControlMessage` returns false on write failure. The extracted `messageWriter` interface and `recordingConn` add unit-level testability without a live WebSocket.

The three-phase lifecycle documentation (state table + Mermaid diagrams + Alloy model) is an unusually thorough treatment that will be valuable for future contributors.

---

### Issues & Questions

**[Question] `/tmp/panemux-replay-input-fix/` is a local worktree path — both doc links are broken for every other user**

`docs/architecture.md` and `docs/behavior.md` both link to the Alloy model using the absolute path of the git worktree the PR was developed in:

```markdown
[replay_state.als](/tmp/panemux-replay-input-fix/docs/models/replay_state.als)
```

This path does not exist outside the author's machine. Since the model is committed at `docs/models/replay_state.als`, the links should use a repo-relative path, e.g.:

```markdown
[replay_state.als](models/replay_state.als)      <\!-- relative from docs/ -->
```

Needs to be fixed in both `docs/architecture.md` and `docs/behavior.md` before merge.

**[Minor question] `replayEnded: true` as the default initial value (carried forward)**

No change from the previous round. The initial `true` reads as "a replay has ended" rather than "no replay is pending" — correct semantics, slightly confusing name. `awaitingReplayEnd: false` or similar would make the initial-state intent clearer. Not a blocker.

**[Minor question] `TestWS_ReplayStartWriteFailureStopsBeforeReplayBody` tests `replay:end` failure, not `replay:start`**

The test name implies a `replay:start` write failure is being exercised, but `failAtWrite = 2` makes the *second* `writeControlMessage` call fail — which is `replay:end`:

```go
recorder.failAtWrite = 2

ok := writeControlMessage(recorder, ControlMessage{Type: "replay", State: "start"})
require.True(t, ok)   // 1st write — succeeds

ok = writeControlMessage(recorder, ControlMessage{Type: "replay", State: "end"})
require.False(t, ok)  // 2nd write — this is the one that fails
```

A name like `TestWS_ReplayEndWriteFailureDetected` would match the actual scenario. Also note the test exercises `writeControlMessage` directly rather than going through `forwardTerminalOutput`; the handler's early-return behaviour on a failed `sendReplay` is implicitly covered by the `sendReplay` → `writeControlMessage` chain but not directly asserted.

**[Suggestion] Alloy `transition` predicate allows `WriteComplete` from `Live` — minor verification gap**

The `WriteComplete` case in `transition` only constrains behaviour when `cur.state = ReplayPendingEnd` or `cur.state = ReplayDraining`; the `Live` case is left unconstrained. Alloy can therefore synthesise traces where `WriteComplete` fires from `Live` and transitions to any state `StateEncoding` permits, including `ReplayPendingEnd`. In the actual code this is impossible — write callbacks only fire for writes initiated with `replayWrite = true`, which requires `replayActive = true` (only true in non-`Live` states).

The `check` assertions all pass because they are state-local and follow from `StateEncoding` independently of the transitions, so the model is not *wrong*. But it is more permissive than the implementation. Adding `cur.state = Live implies next.event \!= WriteComplete` would make the model a tighter characterisation. Happy to file a follow-up issue if useful.

---

### Overall

The reconnect fix is correctly implemented and tested. One blocker before merge: the `/tmp/panemux-replay-input-fix/` worktree paths in both doc files need to be replaced with repo-relative links.

docs/behavior.md
tomo-chan: Fixed in https://github.com/tomo-chan/panemux/commit/84b058d2616f77fca0bcf687ff8105419e152d01 by switching the Alloy model link to the repo-relative path `models/replay_state.als`.

frontend/src/hooks/useTerminal.ts
tomo-chan: Renamed in https://github.com/tomo-chan/panemux/commit/379a88860a921c82b1b37e0d0e4bcfcdd4104325 from `replayEnded` to `awaitingReplayEnd` so the initial-state semantics are explicit.

tomo-chan:
Addressed the remaining review-summary suggestion in https://github.com/tomo-chan/panemux/commit/a5d9f83075357029217a725d6a946abf6fe98f6c .

Changes:
- tightened the Alloy model so `WriteComplete` cannot transition from `Live`
- renamed `TestWS_ReplayStartWriteFailureStopsBeforeReplayBody` to `TestWS_ReplayEndWriteFailureDetected` so the test name matches the scenario it exercises

Verification run:
- `go test ./internal/ws`
- `java -jar /private/tmp/alloy-6.2.0/org.alloytools.alloy.dist.jar exec -t none -o /private/tmp/replay-alloy-out -f docs/models/replay_state.als`

All Alloy checks remained `UNSAT` after the transition tightening.

.github/workflows/model-check.yml
```diff
@@ -0,0 +1,100 @@
...
+          curl -fsSL \
+            -o /tmp/alloy/org.alloytools.alloy.dist.jar \
+            https://github.com/AlloyTools/org.alloytools.alloy/releases/download/v6.2.0/org.alloytools.alloy.dist.jar
+
+      - name: Validate Alloy models
+        run: |
+          set -euo pipefail
+          shopt -s nullglob
+          models=(docs/models/*.als)
+          if [ "${#models[@]}" -eq 0 ]; then
```
tomo-chan, line 47: No integrity check after download. A `sha256sum` step after this curl command would confirm the artifact matches the expected hash. Low risk given GitHub release immutability, but would make the trust boundary explicit.

tomo-chan (commented):
## Code Review (round 3)

### Resolved from previous rounds

All outstanding concerns are addressed:

- **[Question] `/tmp/` paths in docs** → fixed to `models/replay_state.als` in both `docs/architecture.md` and `docs/behavior.md` (`84b058d2`).
- **[Minor question] `replayEnded` naming** → renamed to `awaitingReplayEnd` throughout (`379a8886`). The guard in `maybeRestoreTerminalInput` is now `if (entry.awaitingReplayEnd || ...)` which directly expresses the invariant.
- **[Minor question] test name** → `TestWS_ReplayEndWriteFailureDetected` now correctly describes the scenario (`a5d9f830`).
- **[Suggestion] Alloy `WriteComplete` from `Live`** → tightened with `cur.state in ReplayPendingEnd + ReplayDraining` (`a5d9f830`).

Beyond those fixes, the additional Alloy additions (`SocketClose`, `ReplayEndWriteFail` events, two new assertions) and the CI workflow (`bce20ffb`–`dce55da3`) that gates on model checks are excellent follow-through. The state machine is now both formally verified and continuously checked in CI.

---

### Issues & Questions

**[Minor question] Alloy jar downloaded without integrity check**

The `Download Alloy` step fetches the jar over HTTPS from a specific GitHub release tag but does not verify the artifact hash:

```yaml
- name: Download Alloy
  run: |
    set -euo pipefail
    mkdir -p /tmp/alloy
    curl -fsSL \
      -o /tmp/alloy/org.alloytools.alloy.dist.jar \
      https://github.com/AlloyTools/org.alloytools.alloy/releases/download/v6.2.0/org.alloytools.alloy.dist.jar
```

GitHub release assets are immutable once published, so the risk is low in practice. That said, adding a `sha256sum` check after the download makes the assumption explicit and would catch any accidental or deliberate replacement of the artifact:

```yaml
    echo "<sha256>  /tmp/alloy/org.alloytools.alloy.dist.jar" | sha256sum --check
```

The SHA can be obtained from the release's SHA-256 checksum file if one is published, or by computing it from a known-good local copy. Worth adding if this workflow is expected to run long-term, but not a blocker.

**[Suggestion] Cache the Alloy jar between CI runs**

Every run downloads the jar fresh. A `actions/cache` step keyed on the version string would avoid a redundant download on cache hits. Happy to file a follow-up issue if that's useful.

---

### Overall

All review concerns are resolved. The PR is ready to merge. The only open item is the optional jar integrity check, which can be done in a follow-up.

.github/workflows/model-check.yml
tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/1678cbf40b3b7c4c2d2c95f4ae3d14c12d5562a5. The workflow now verifies the downloaded Alloy jar against a pinned SHA-256 before executing it. In the same change I also switched the Go JUnit step to `go-junit-report -iocopy -out ...` so any future `ci` Go test failures keep their raw test output in the Actions log instead of collapsing to a bare exit code.

Verified locally with `shasum -a 256 -c -` against the downloaded jar and with `go test -v ./... -race 2>&1 | go-junit-report -iocopy -out test-results/go-junit.xml -set-exit-code`.

tomo-chan (commented):
## Code Review (round 4)

### Resolved from previous round

**[Minor question] Alloy jar downloaded without integrity check → resolved**

`1678cbf2` adds a pinned `ALLOY_JAR_SHA256` env var and a `shasum -a 256 -c -` verification step immediately after the `curl` download. The 64-character hex digest is the correct length for SHA-256, and `set -euo pipefail` ensures the workflow aborts if the hash does not match. The trust boundary around the downloaded artifact is now explicit.

---

### What's Good

Looking at the full PR in its final state:

- The replay suppression mechanism (`replayWriteDepth` + `awaitingReplayEnd` + `maybeRestoreTerminalInput`) correctly handles all three phases — active replay, draining, and live — including the reconnect-reset path.
- The Alloy model (`docs/models/replay_state.als`) formally specifies the state machine, covers fault transitions (`ReplayEndWriteFail`, `SocketClose`), and ships with seven `check` assertions, all verified UNSAT.
- The CI workflow gates on model correctness whenever model files change, with artifact integrity verification, ANSI-stripped verdict parsing, and receipt-based check counting.
- Every concern raised across three rounds of review was addressed in a follow-up commit, often with additional test coverage or documentation beyond what was asked.

---

### Overall

No open concerns. The PR is ready to merge.

PR #89, opened 2026-05-15
feat: add persistent pane layout editing

tomo-chan:
## Summary
- remove edit mode and persist layout and workspace changes immediately
- add pane move and insert interactions via header drag handles, pane edges, workspace edges, and divider drop zones
- add pane-header quick-add actions for default local panes to the right or below
- simplify the workspace bar so it focuses on workspace tabs, workspace add, rename/delete, and tab-position controls
- refresh the header and workspace-bar UI to match the final interaction model, and update the docs accordingly

## Test Plan
- [x] make test-frontend
- [x] make build-frontend
- [x] make test


tomo-chan (commented):
## Code Review

### What's Good

- **Clean end-to-end removal of edit mode**: frontend hooks, components, backend handlers, routes, and tests are all consistently updated in a single PR — no dangling references.
- **`saveLayout` helper extracted in `useLayout.ts`**: the repeated `fetch` call duplicated across `splitPane`, `closePane`, and `swapPanes` is now centralised. New `createPane` and `movePane` operations inherit it naturally.
- **`ensurePaneDefaults` extracted and reused**: `splitPane` and `createPane` both benefit, and the shell-detection logic is now testable in isolation.
- **`DividerDropZone` cleanly wraps `SplitDivider`**: drop handling is added without polluting the divider component, and the hover overlay gives clear visual feedback without affecting pointer events.
- **Pure, well-tested `layoutTree.ts` additions**: `insertPaneAtWorkspaceEdge`, `insertPaneBesideTargetPane`, `movePaneBesideTargetPane`, and `movePaneToWorkspaceEdge` all have unit tests covering axis-matching, cross-axis wrapping, and the move-without-duplication invariant.
- **`DeleteSession` error handling improved**: the previous path silently swallowed save errors with `//nolint:errcheck`; the new code returns a proper 500 — a correctness win regardless of the edit-mode removal.
- **Drag-handle UX improvement**: the small `⠿` icon with `moveHandleProps` is cleaner than the previous full-header grab cursor; the terminal content remains interactive at all times.

---

### Issues & Questions

**[Question] `DeleteSession` calls `SaveWorkspaces()` but `PutLayout` calls `SaveLayout()` — are they equivalent for single-workspace configs?**

`internal/api/handler.go`:
```go
// PutLayout (resize / rearrange):
if err := h.cfg.SaveLayout(layout); err \!= nil { ... }

// DeleteSession (close pane):
h.cfg.RemovePaneFromLayout(id)
if err := h.cfg.SaveWorkspaces(); err \!= nil { ... }
```

These two operations both modify the layout, but now call different save methods. The old code used `h.cfg.SaveLayout(h.cfg.Layout)` for the delete path. For a user running panemux without a `workspaces:` block (single-layout config), does `SaveWorkspaces()` correctly persist the layout change, or does it only cover the workspaces YAML path while `SaveLayout()` covers the root `layout:` path? If they diverge, closing a pane silently fails to persist on single-workspace setups.

---

**[Question] `normalizeChildrenSizes` resets all sibling sizes to equal — is this intended?**

`frontend/src/utils/layoutTree.ts`:
```ts
export function normalizeChildrenSizes(children: LayoutChild[]): LayoutChild[] {
  const size = 100 / children.length
  return children.map((child) => ({ ...child, size }))
}
```

This is called when inserting into an existing axis (both in `insertPaneAtWorkspaceEdge` when the direction matches, and in `insertRelativeToDirectChild`). If a user has a custom 70/30 split and then inserts a new pane on the same axis, all siblings reset to equal sizes. For the new pane's initial size this is reasonable, but discarding the relative ratio of existing panes may be surprising. Is equal distribution the intentional policy here?

---

**[Minor question] Divider drop target selection uses `findBoundaryPaneId` — placement may be non-obvious in nested layouts**

`frontend/src/components/SplitContainer.tsx`:
```tsx
const targetPaneId =
  findBoundaryPaneId(beforeChild, 'end') ??
  findBoundaryPaneId(afterChild, 'start')
if (\!targetPaneId) return
ctx?.onMovePaneBeside(sourceId, targetPaneId, edge)
```

When the layout has nested splits (e.g., a vertical group inside a horizontal split), dropping onto the outer divider resolves to a boundary leaf pane deep in the tree. The dropped pane is then inserted *beside that leaf*, creating a new nested split rather than beside the top-level sibling group. For simple layouts this is fine, but in complex configurations the insertion point can be unintuitive. Was inserting beside the direct child container (rather than its boundary leaf) considered?

---

**[Minor question] `saveLayout` errors are silently dropped for `movePane`**

`frontend/src/hooks/useLayout.ts`:
```ts
const saveLayout = useCallback(async (updatedLayout: LayoutNode) => {
  await fetch(...).catch(console.error)   // error logged, not rethrown
}, [workspaces?.active])

const movePane = useCallback(async (...) => {
  setLayout(newLayout)
  setWorkspaces(...)
  await saveLayout(newLayout)             // save failure is invisible to the user
}, [layout, saveLayout])
```

If the server returns an error during a drag-to-move operation, the UI reflects the new layout but the config file is not updated — a silent inconsistency that resolves on page reload. This behaviour was already present in `swapPanes`, so it is not a regression. However, now that `movePane` is the primary drag-and-drop interaction (rather than an edit-mode-only feature), surfacing the save error (as `createPane` does) might be worth considering.

---

### Overall

A coherent refactor that replaces the edit-mode gate with always-on persistence. The main question to resolve before merge is whether `SaveWorkspaces()` in `DeleteSession` correctly covers single-workspace configs — the original code used `SaveLayout()` for that path, and a silent divergence there would cause pane-close changes not to persist.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/c72224e717d7ca4b7e9baaa79825c38ff48e7eab.

Changes made:
- switched `DeleteSession` back to `SaveLayout(h.cfg.Layout)` so the delete path matches the layout-persistence path more directly
- added `useLayout` tests for `createPane` and `movePane`, covering workspace-edge placement, pane-edge placement, save failures, and the invariant that move operations do not recreate sessions
- reran the state-transition-sensitive checks after the change

Notes on the remaining review questions:
- equal sibling redistribution on insert is still intentional for this PR; the new behavior is covered by the layout tree tests and unchanged by this follow-up commit
- divider drops still resolve through the boundary leaf of the adjacent subtree in this PR; that placement model is unchanged, and the follow-up here focused on persistence consistency plus state-transition/coverage validation

Local verification actually run:
- `make coverage-frontend`
- `make test`

tomo-chan (commented):
## Follow-up Review (commit c72224ec)

### Previously Raised — All Resolved

Thank you for the prompt follow-up. Every concern from the first review is addressed:

| Finding | Resolution |
|---|---|
| `DeleteSession` called `SaveWorkspaces()` instead of `SaveLayout()` | ✅ Switched back to `SaveLayout(h.cfg.Layout)` — now consistent with `PutLayout` |
| `movePane` save errors silently dropped | ✅ `saveLayout` gains `throwOnError` flag; both `createPane` and `movePane` pass `true` |
| Equal sibling size redistribution | Acknowledged as intentional for this PR |
| Divider boundary-leaf drop target | Acknowledged as intentional for this PR |

---

### New Observations

**[Question] `movePane` save errors are thrown but immediately discarded by `void` callers in `App.tsx`**

`frontend/src/App.tsx`:
```tsx
onMovePaneToWorkspaceEdge: (sourcePaneId, edge) => {
  void movePane(sourcePaneId, { type: 'workspace-edge', edge })
},
onMovePaneBeside: (sourcePaneId, targetPaneId, edge) => {
  void movePane(sourcePaneId, { type: 'pane-edge', targetPaneId, edge })
},
```

`void` discards the promise, so any error thrown by `movePane` (including the new `HTTP 500` from `saveLayout`) is silently swallowed in the UI. The in-memory layout is already updated before `saveLayout` is called, so on a save failure the terminal arrangement visible to the user diverges from the persisted config — and no error is shown.

Contrast with `createPane`, where the caller (`handleCreatePane` in `App.tsx`) awaits the result inside a `try/catch` and shows the error via `setNewTerminalError`. The tests correctly assert that `movePane` throws `'HTTP 500'`, but because the actual callers use `void`, those throws have no user-visible effect.

Is the `throwOnError = true` flag in `movePane` intended as a foundation for future error-surface wiring, or should the drag-and-drop handlers be updated to catch and display errors now?

---

**[Minor question] `saveLayout` with `throwOnError = true` still silently swallows network-level failures**

`frontend/src/hooks/useLayout.ts`:
```ts
const response = await fetch(...).catch(console.error)
if (throwOnError && response && \!response.ok) {
  throw new Error(`HTTP ${response.status}`)
}
```

If the network request itself rejects (offline, DNS failure), `.catch(console.error)` returns `undefined`, and the `response &&` guard prevents the throw from firing — even with `throwOnError = true`. HTTP-level errors (status ≥ 400) are now correctly surfaced, but network-level failures pass silently. This asymmetry may surprise future maintainers who expect `throwOnError` to cover all failure modes. Worth a brief comment on the function, or converting `.catch` to a re-throw when `throwOnError` is set.

---

### Overall

The follow-up commit resolves all four earlier concerns and adds meaningful `createPane`/`movePane` test coverage. The one item to discuss before merge is whether the `void movePane(...)` call sites in `App.tsx` should be updated to surface save errors to the user — right now the improved error propagation from `saveLayout` does not reach any user-visible state for drag-and-drop operations.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/0f110a6f44171eeebcee6c4891cb7cad104a5115.

Changes made:
- surfaced drag-and-drop move persistence failures in `App.tsx` with a user-visible `role="alert"` message instead of discarding `movePane(...)` rejections via `void`
- updated `saveLayout(..., true)` to rethrow network-level `fetch` failures as well as HTTP error responses, while preserving the existing non-throwing path for debounce/background saves
- added regression tests for both behaviors:
  - `useLayout` now asserts that `movePane` rejects on network failures
  - `App` now asserts that a failed pane move shows a visible error message

State-transition note:
- no session or layout lifecycle was intentionally changed in this follow-up; this is error propagation only, so the existing model remains valid

Local verification actually run:
- `make coverage-frontend`
- `make test`

tomo-chan (commented):
## Follow-up Review (commit 0f110a6f)

### Previously Raised — All Resolved

| Finding | Resolution |
|---|---|
| `void movePane(...)` discards errors silently | ✅ `handleMovePane` added — errors caught with `.catch()` and shown via `role="alert"` div |
| `saveLayout(throwOnError=true)` swallows network failures | ✅ Refactored to `try/catch` — network rejections now rethrown when `throwOnError` is set |

The `saveLayout` refactor from `.catch(console.error)` to `try/catch` with conditional re-throw is the right approach — it correctly distinguishes the background-save path (non-throwing) from the user-initiated path (throwing). The `role="alert"` on the error div is a good accessibility touch. The `App.test.tsx` named-mock extraction is a clean improvement that makes future test control straightforward.

---

### New Observations

**[Minor question] Move error has no dismiss mechanism**

`frontend/src/App.tsx`:
```tsx
{movePaneError && (
  <div role="alert" style={{ position: 'absolute', top: 12, right: 12, ... }}>
    Failed to move terminal: {movePaneError}
  </div>
)}
```

The error banner persists until the next `handleMovePane` call clears it via `setMovePaneError(null)`. If the user abandons the drag-and-drop after a failure and moves on (switching workspace, opening a dialog), the banner stays visible indefinitely. A close button or a short auto-dismiss (e.g. `setTimeout(..., 5000)`) would bring this in line with the ephemeral error pattern used by most toast systems. Is the current "cleared only by next move attempt" lifecycle intentional here?

---

**[Minor question] Fallback error message creates a redundant string when the thrown value is not an `Error` instance**

`frontend/src/App.tsx`:
```tsx
setMovePaneError(err instanceof Error ? err.message : 'Failed to move terminal')
```
`frontend/src/App.tsx` (display):
```tsx
Failed to move terminal: {movePaneError}
```

If `err` is not an `Error` instance, `movePaneError` is set to `'Failed to move terminal'`, and the rendered text becomes `"Failed to move terminal: Failed to move terminal"`. In practice, errors from `saveLayout` are always `Error` objects, so this path is never hit — but the fallback value `'Something went wrong'` (or similar) would produce a less surprising message if it ever were. Very low priority.

---

### Overall

All functional and error-propagation concerns from prior rounds are now fully addressed. The remaining two items are UX-level polish — neither blocks correctness. This PR is ready to merge from a code-review standpoint.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/28837a7cba2c298bba2e3524c16d8bc8bc639c16.

Changes made:
- added an explicit dismiss button to the pane-move error banner so the user can clear it without needing to attempt another drag
- changed the non-`Error` fallback message to `Something went wrong`, which avoids rendering `Failed to move terminal: Failed to move terminal`
- added `App` tests covering both the dismiss path and the non-`Error` fallback path

State-transition note:
- this follow-up changes only the UI error-banner lifecycle for move failures: `hidden -> visible on rejected move -> hidden on explicit dismiss`
- session lifecycle and layout persistence semantics are unchanged; the existing session-state model remains valid

Local verification actually run:
- `make test-frontend`
- `make test`

tomo-chan (commented):
## Follow-up Review (commits 28837a7c → e6e6986f)

### Previously Raised — All Resolved

| Finding | Resolution |
|---|---|
| Error banner has no dismiss mechanism | ✅ `28837a7c` adds a `×` dismiss button with `aria-label="Dismiss move error"`, with a test for the dismiss path |
| Fallback message produces redundant text | ✅ `28837a7c` changes the fallback from `'Failed to move terminal'` to `'Something went wrong'`, with a test for the non-`Error` path |

---

### New Changes (commit e6e6986f — harden dialog interactions)

The Escape-key and backdrop-dismiss hardening is a good addition: the `isSaving` guard on both code paths (`window.keydown` handler and `onClick` backdrop) is consistent across all three dialogs, and the `maxWidth/maxHeight/overflowY` viewport constraints prevent dialogs from escaping small windows. Adding tests for the new behaviours alongside the code change is exactly right.

---

### New Observation

**[Minor question] `PaneSettingsDialog` tests are missing the "Escape suppressed while saving" case**

The new `AddSSHHostDialog` and `NewTerminalDialog` test files both cover four dismissal variants:

| Test case | AddSSHHostDialog | NewTerminalDialog | PaneSettingsDialog |
|---|---|---|---|
| Escape closes when not saving | ✅ | ✅ | ✅ |
| Escape suppressed while saving | ✅ | ✅ | ❌ missing |
| Backdrop closes when not saving | ✅ | ❌ missing | ❌ missing |
| Backdrop suppressed while saving | ✅ | ✅ | ✅ |
| Viewport constraints | ✅ | ✅ | ✅ |

The most meaningful gap is the `PaneSettingsDialog` "does not close on Escape while saving" case — that is the only path that directly exercises the `isSaving` guard in the new Escape `useEffect`. The `pane` null-guard in the same condition (`if (\!isOpen || \!pane || isSaving) return`) is similarly untested for the `isSaving` branch. The missing "backdrop closes when not saving" cases for the other two dialogs are lower priority since the happy-path backdrop close was already tested before this PR.

---

### Overall

All concerns from every prior review round are now resolved or acknowledged. The only remaining item is the missing Escape-while-saving test for `PaneSettingsDialog` — a low-effort addition given the pattern is already written for the other two dialogs. Everything else is merge-ready.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/a2dc0a4aaa3c15cf835e51f75a1e90e3de6a8253.

I did a factor pass over the dialog dismissal tests instead of only addressing the single pointed-out case. The factors I checked were:

- trigger: `Escape` vs backdrop click
- save state: saving vs not saving
- render guard: valid pane vs missing pane for `PaneSettingsDialog`

Additional coverage added:
- `PaneSettingsDialog`
  - does not render when `pane` is `null`
  - does not close on `Escape` while saving
  - closes on backdrop click when not saving
- `NewTerminalDialog`
  - closes on backdrop click when not saving

State-transition note:
- this follow-up only adds tests; it does not change dialog, session, or layout behavior

Local verification actually run:
- `make test-frontend`

tomo-chan (commented):
## Follow-up Review (commit a2dc0a4a)

### Previously Raised — Resolved

The [Minor question] about missing Escape-while-saving coverage for `PaneSettingsDialog` is fully addressed. Rather than adding only the single pointed-out case, the author did a factor pass over all three dialog dismissal test suites. The final coverage matrix is now complete:

| Test case | AddSSHHostDialog | NewTerminalDialog | PaneSettingsDialog |
|---|---|---|---|
| Escape closes when not saving | ✅ | ✅ | ✅ |
| Escape suppressed while saving | ✅ | ✅ | ✅ |
| Backdrop closes when not saving | ✅ | ✅ | ✅ |
| Backdrop suppressed while saving | ✅ | ✅ | ✅ |
| Viewport constraints | ✅ | ✅ | ✅ |

The `PaneSettingsDialog` "does not render when pane is missing" case (exercising the `\!pane` guard) is a nice bonus that wasn't asked for.

---

### No New Findings

This commit is test-only. No production behaviour changed.

---

### Overall

All concerns across every prior review round are now fully resolved. This PR is ready to merge.

tomo-chan (commented):
## Follow-up Review (commits 52e6720e → 24ca1b1d)

### Previously Raised — All Still Resolved

No previously raised concerns are re-opened by the new commits.

---

### New Changes Summary

The new batch introduces: wider drop targets, a pointer-event fallback for HTML5 drag reliability, native-feeling pane drag-and-drop via `resolvePaneDropEdge`, header quick-add buttons for default pane creation (`onCreatePaneBeside`), SVG icons in `PaneHeaderIcons.tsx`, and several workspace-bar styling iterations. The drag-and-drop overhaul is well-structured — the dual HTML5-drag / pointer-event path is symmetric across `DividerDropZone` and `TerminalPane`, and `resolvePaneDropEdge` + `dropZoneStyle` are exported and unit-tested.

---

### Issues & Questions

**[Question] `NewTerminalDialog` is shipped as dead code**

`NewTerminalDialog.tsx` and `NewTerminalDialog.test.tsx` are added in the PR diff, but no production file imports `<NewTerminalDialog`. The cumulative `App.tsx` diff adds no `import NewTerminalDialog` line. The component was initially wired to a workspace-bar "Add Terminal" button that was removed in commit `386900ef` ("refactor: remove workspace bar terminal action"), and the header quick-add path (`onCreatePaneBeside` → `handleCreateDefaultPane`) bypasses the dialog entirely.

Is `NewTerminalDialog` intentionally being landed as an unreachable component (reserved for a future entry point), or was re-connecting it to the header flow intended and the import was dropped accidentally?

---

**[Question] `createPaneError` banner has no App-level test coverage**

`frontend/src/App.tsx` adds a `createPaneError` error banner symmetric to the `movePaneError` banner. The `movePaneError` banner has three tests: error shows on rejection, dismiss button clears it, non-`Error` fallback renders correctly. The `createPaneError` banner has zero failure-path tests — only the happy-path (`mockCreatePane` resolves) is covered.

The missing cases would be:
```
it('shows a user-visible error when creating a pane fails')
it('dismisses the create pane error banner when requested')
it('shows a generic create error when the rejection is not an Error instance')
```

The banner code itself (including the dismiss button with `aria-label="Dismiss create terminal error"`) is untested at the component integration level.

---

**[Minor question] `PANE_DROP_ZONE_THICKNESS` is exported but never referenced**

`frontend/src/components/TerminalPane.tsx`:
```ts
export const PANE_DROP_ZONE_THICKNESS = 24
export const PANE_DROP_ZONE_RATIO = 0.5
```

`PANE_DROP_ZONE_RATIO` is used in `dropZoneStyle` and imported in `TerminalPane.test.tsx`. `PANE_DROP_ZONE_THICKNESS` is exported but has no usages in the diff. Is this a leftover from a discarded approach, or a value intended for a future thinner-strip variant?

---

**[Minor question] Both error banners occupy the same absolute position**

`frontend/src/App.tsx` — both banners use `position: 'absolute', top: 12, right: 12`. If `createPaneError` and `movePaneError` are both truthy at the same time (create fails, then drag-move also fails before the first is dismissed), the second banner renders directly on top of the first. Is the overlap case considered acceptable, or should one be offset (e.g., `top: 12` vs `top: 72`)?

---

### Overall

The drag-and-drop overhaul and header quick-add are solid additions. Two items warrant a response before merge: whether `NewTerminalDialog` being unreachable is intentional, and whether the `createPaneError` banner failure paths should be tested (following the same pattern already established for `movePaneError`).

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/e33a34c6288334f991639eb8bf17cb605dfd37c4.

Changes made:
- removed the now-unreachable `NewTerminalDialog` production flow and its test file, since terminal creation no longer has any workspace-bar entry point
- added `App` coverage for the `createPaneError` banner: visible error on rejection, explicit dismiss, and generic non-`Error` fallback
- removed the unused `PANE_DROP_ZONE_THICKNESS` export
- offset the move-error banner when a create-error banner is already visible so the two alerts do not overlap
- updated docs to match the quick-add-only pane creation flow and the stacked error-banner behavior

State-transition note:
- pane creation still follows the same lifecycle (`quick-add -> POST /api/sessions -> layout save`); this follow-up only removes dead UI, adds error-surface coverage, and clarifies the error-banner UI state
- session create/delete/restart and pane-move persistence semantics are unchanged

Local verification actually run:
- `cd frontend && npm test -- --run src/App.test.tsx`
- `make test-frontend`
- `make build-frontend`

tomo-chan (commented):
## Follow-up Review (commit e33a34cf)

### Previously Raised — All Resolved

| Finding | Resolution |
|---|---|
| `NewTerminalDialog` shipped as dead code | ✅ Both `NewTerminalDialog.tsx` and `NewTerminalDialog.test.tsx` fully removed (481 lines deleted) |
| `createPaneError` banner has no App-level tests | ✅ Three tests added: error shows on rejection, dismiss button works, non-`Error` fallback renders correctly |
| `PANE_DROP_ZONE_THICKNESS` exported but unused | ✅ Export removed from `TerminalPane.tsx` |
| Both error banners overlap at same position | ✅ `movePaneError` banner uses `top: createPaneError ? 72 : 12` to stack below the create-error banner when both are visible |

The docs in `ui-design.md` and `behavior.md` are updated to match the quick-add-only pane creation flow and the stacked error-banner lifecycle. The non-`Error` fallback in `handleCreateDefaultPane` is now `'Something went wrong'`, consistent with `handleMovePane`.

---

### No New Findings

This commit is a clean, focused removal and coverage pass with no behavioural side effects.

---

### Overall

All concerns across every review round are now fully resolved. This PR is ready to merge.

tomo-chan (commented):
## Follow-up Review (commits 4e43cb14 → 6476c168)

### Previously Raised — All Still Resolved

No previously raised concerns are re-opened by the new commits.

---

### New Changes Summary

- `4e43cb14`: VSCode icon updated to a more detailed branded SVG; `vscodeButtonStyle` adds a blue tint.
- `2723a94d`: `HeaderIconButton` component extracted in `PaneHeader.tsx` — hover and pressed states managed via `useState`. All header action buttons replaced. Tests cover hover affordance and the full press/release cycle.
- `6476c168`: Hover and pressed states added to `WorkspaceTabs` buttons via a `surfaceButtonStyle` helper and an `applyInteractiveButtonState` DOM mutation function. Tests cover hover affordance and pressed state.

The `HeaderIconButton` extraction in `PaneHeader.tsx` is clean — collocating the interaction state with the element and re-rendering only the affected button is the right pattern for React.

---

### Issues & Questions

**[Question] `WorkspaceTabs` uses direct DOM mutation for hover/pressed states while `PaneHeader` uses React state — re-render during hover will reset styles**

`frontend/src/components/WorkspaceTabs.tsx` — `applyInteractiveButtonState` mutates `button.style.backgroundColor`, `button.style.boxShadow`, and `button.style.transform` directly. When React re-renders `WorkspaceTabs` — triggered by any state change inside the component (e.g., `editingWorkspaceId` changing when a user double-clicks a tab to rename it, or `draftTitle` updating on each keypress) — React re-applies `style={surfaceButtonStyle(...)}` on every button, which overwrites the DOM-mutated values and snaps any hovered or pressed button back to rest colors even if the pointer is still over it.

`HeaderIconButton` in `PaneHeader.tsx` does not have this problem because hover and pressed state live in React state, so React includes them in its re-render output.

Was the DOM-mutation approach chosen intentionally for `WorkspaceTabs` (e.g., to avoid per-button re-renders in a list that also has rename/typing state), or was the `HeaderIconButton` pattern considered and rejected?

---

**[Minor question] `--pmx-selected-background` custom property is set in `surfaceButtonStyle` but never read**

`surfaceButtonStyle` sets both `--pmx-base-color` and `--pmx-selected-background` as CSS custom properties via the `style` prop. `applyInteractiveButtonState` reads `--pmx-base-color` via `button.style.getPropertyValue`, but never reads `--pmx-selected-background` — it recomputes the selected background locally from the `selected` parameter. The property is set but has no consumer. Is it a leftover from a planned use, or can it be removed?

---

### Overall

The VSCode icon update and `HeaderIconButton` extraction are clean. The main question is whether the `WorkspaceTabs` DOM-mutation approach is intentional given that React re-renders during rename/typing will reset hover styles on unrelated buttons.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/98f16d43f0dca66226d3ddde79f506ad46fb9c90.

Changes made:
- replaced the `WorkspaceTabs` DOM-mutation hover/pressed logic with a React-state-driven `InteractiveSurfaceButton`, matching the `PaneHeader` interaction model
- removed the unused custom-property approach entirely
- added a regression test that keeps a hovered workspace tab in the hover state across a component re-render triggered by starting inline rename

State-transition note:
- this follow-up changes only UI affordance state for workspace-bar controls (`rest -> hover -> pressed`)
- workspace selection, rename, add/delete, tab-position changes, and all session/layout lifecycles are unchanged

Local verification actually run:
- `cd frontend && npm test -- --run src/components/WorkspaceTabs.test.tsx`
- `make build-frontend`

tomo-chan (commented):
## Follow-up Review (commit 98f16d4f)

### Previously Raised — All Resolved

| Finding | Resolution |
|---|---|
| `WorkspaceTabs` used direct DOM mutation for hover/pressed state, causing React re-renders to snap styles back | ✅ `applyInteractiveButtonState` and `surfaceButtonStyle` removed entirely; replaced by `InteractiveSurfaceButton` using `useState` — same pattern as `HeaderIconButton` in `PaneHeader.tsx` |
| `--pmx-selected-background` CSS custom property set but never read | ✅ Gone with the old approach — no dead property to remove |

---

### New Changes Summary

- **`InteractiveSurfaceButton`** — a `React.forwardRef` button component that manages hover/pressed state via `useState`. Accepts `selected` and `danger` semantic props. Replaces all raw `<button>` elements in `WorkspaceTabs.tsx` (tabs, add, rename, delete, and position-control buttons), making the interaction model consistent throughout.
- Tests added: hover affordance, pressed state, and — critically — a test that triggers a component re-render (by clicking Rename while hovering a different tab) and asserts the hovered tab keeps its style. This test directly covers the failure mode raised in the prior review.

---

### What's Good

- The regression test `'keeps hover affordance after a component re-render'` is precise: it clicks Rename while a different tab is hovered, forcing a state update inside `WorkspaceTabs`, and then asserts the hovered tab still shows hover styles. That is exactly the scenario the prior review was concerned about, and exactly how you'd want it covered.
- `React.forwardRef` on `InteractiveSurfaceButton` is good practice for a reusable button component even if no caller uses the ref today.
- Callback chaining in the event handlers (`onMouseEnter?.(event)` etc.) correctly preserves any externally passed handlers without silently dropping them.

---

### Issues & Questions

**[Minor question] `danger` silently wins over `style.color` in the resting state**

In `InteractiveSurfaceButton`:

```tsx
const baseColor = (mergedStyle.color as string | undefined) ?? '#b8beca'
let color = danger ? '#f08b8b' : baseColor
```

The delete button passes `color: active ? '#d7dce5' : '#8f96a3'` through the `style` prop, but since `danger` is `true`, `baseColor` is computed and then discarded — the active/inactive colour distinction in `style.color` never takes effect. The rendered colour is `#f08b8b` in both active and inactive states.

This isn't a bug (a consistent reddish colour for the delete action is arguably better than an active/inactive tint), but it is a surprising API surface: a consumer passing `color` via `style` on a `danger` button gets no effect from it. Dropping the `color` key from the delete button's `style` prop would make the intent explicit and avoid a future maintainer wondering why that conditional is there.

---

### Overall

The `InteractiveSurfaceButton` refactor is clean and fully resolves both previously raised concerns. The targeted test for hover-state stability through re-renders is exactly the right coverage. The minor question above is cosmetic and does not block merge.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/b24e118d4ee05acdebfca018f9d8fca1c2577bb8.

Changes made:
- removed the redundant `style.color` override from the workspace delete button now that `danger` intentionally owns the resting color in `InteractiveSurfaceButton`
- this makes the delete-button API intent explicit and avoids leaving a dead conditional in the caller

State-transition note:
- no behavior changed; this is a styling cleanup only
- workspace actions, pane/session lifecycles, and hover/pressed interaction semantics are unchanged

Local verification actually run:
- `cd frontend && npm test -- --run src/components/WorkspaceTabs.test.tsx`
- `make build-frontend`

tomo-chan (commented):
## Follow-up Review (commit b24e1189)

### Previously Raised — All Resolved

| Finding | Resolution |
|---|---|
| `danger` prop silently overrides `style.color`, leaving the active/inactive tint on the delete button dead | ✅ `b24e1189` removes the redundant `color: active ? '#d7dce5' : '#8f96a3'` key — `danger` now unambiguously owns the resting colour |

---

### Overall

All concerns across every round of review are resolved. Ready to merge.

tomo-chan:
Addressed the failing `ci` job with https://github.com/tomo-chan/panemux/commit/2e122936ad462aa5c928820338ebe209f702b3b2.

Root cause was a flaky assertion in `TestManagedSession_SubscribeStillWorksWhilePublishWaitsOnSlowSubscriber`: `unsubscribe()` can race with one last queued update, so asserting an immediately closed channel was too strict under CI `-race` timing. The test now drains any final queued update and waits for the stream to close, while still verifying that subscribe/unsubscribe does not block behind a slow subscriber.

Local verification:
- `go test ./internal/session -run TestManagedSession_SubscribeStillWorksWhilePublishWaitsOnSlowSubscriber -count=50`
- `go test ./internal/session -run TestManagedSession_SubscribeStillWorksWhilePublishWaitsOnSlowSubscriber -race -count=20`
- `make lint-go`
- `make test-go`

tomo-chan (commented):
## Follow-up Review (commit 2e122936)

### Previously Raised — All Still Resolved

No previously raised concerns are re-opened by the new commit.

---

### New Changes Summary

`internal/session/manager_test.go` — `TestManagedSession_SubscribeStillWorksWhilePublishWaitsOnSlowSubscriber`

**Root cause:** the original assertion used a non-blocking `select { case _, ok := <-updates: require.False(t, ok); default: }`. If a publish raced with `unsubscribe()` and enqueued one final update before the subscription was removed, the channel received that message with `ok=true`, causing `require.False(t, true)` to fail.

**Fix:** replace the non-blocking select with a draining loop:
```go
for {
    if _, ok := <-updates; \!ok {
        break
    }
}
```
The loop consumes any trailing updates, then exits when the channel is closed (`ok=false`). If the channel is never closed — a real bug — the loop blocks until the 1 s timeout fires and fails the test, so the invariant is still enforced.

The timeout increase from 100 ms to 1 s is appropriate for `-race` builds, which carry substantially more scheduling overhead.

---

### Overall

All concerns are resolved. The CI flake fix is correct and still enforces the non-blocking subscribe/unsubscribe guarantee. Ready to merge.

PR #91, opened 2026-05-16
feat: add directory browser for pane working directories

tomo-chan:
## Summary
- add a directory browser to pane settings for choosing working directories
- support both local and SSH-backed directory listings through a new read-only API
- add frontend and backend tests plus docs for the new working-directory selection flow

## Test plan
- [x] make build-frontend
- [x] make lint-go
- [x] make lint-frontend
- [x] make test-go
- [x] make test-frontend

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,6 +91,11 @@ Sessions are started at server startup from the YAML config (`main.go: startSess
 
 ## Development Rules
 
+### Path sanitization
+- Do not record real local or remote directory paths from a developer machine or environment in tests, fixtures, docs, screenshots, PR bodies, review replies, or commit history.
+- Replace environment-specific paths with clearly fake placeholders such as `/workspace/user/project`, `/tmp/sample-project`, or `/remote/home/demo`.
+- If a real path is used temporarily for investigation, remove it before committing and rewrite the branch history if that path already landed in a published commit on the PR branch.
+
 ### TDD
 - Always **write tests first**, confirm they fail, then implement.
 - Go: all tests must pass (`make test-go`) before moving on.
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ Common uses:
 - A pane with `connection: my-host` can use either a named `ssh_connections` entry or a `Host my-host` entry from `~/.ssh/config`.
 - `tmux` and `ssh_tmux` panes automatically create the target tmux session if it does not already exist.
 - Set `cwd` on `local`, `ssh`, or `ssh_tmux` panes when you want the shell to start in a specific directory.
+- In the pane settings dialog, `Working Directory` can be chosen from a browsable directory tree for both local and SSH-backed panes. Hidden directories are available through a toggle.
 
 ---
 
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -98,6 +98,8 @@ Host-key verification uses `known_hosts_file` if configured, or `~/.ssh/known_ho
 
 `cwd` is validated against `validRemotePath` before use: absolute paths only, no shell metacharacters (`;|&$` + "`" + `'"<>(){}[]!`), no control characters. See *Security Design* in `architecture.md`.
 
+Pane settings in the frontend expose a directory browser for `cwd`. Local and local tmux panes browse the local filesystem; `ssh` and `ssh_tmux` panes browse the selected SSH connection's remote filesystem. The browser lists directories only and hides dot-directories by default unless the user enables the hidden-directory toggle.
+
 Persistence behavior:
 
 - layout and workspace changes are persisted immediately when a save path is available
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -88,6 +88,7 @@ vi.mock('./hooks/usePaneSettings', () => ({
     saveSettings: vi.fn(),
     addSSHConfigHost: vi.fn(),
     detectShell: vi.fn(),
+    browseDirectories: vi.fn(),
   }),
 }))
 
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -20,7 +20,7 @@ export const App: React.FC = () => {
   const [maximizedPaneId, setMaximizedPaneId] = useState<string | null>(null)
   const [dragSourcePaneId, setDragSourcePaneId] = useState<string | null>(null)
   const [attentionPaneIds, setAttentionPaneIds] = useState<Set<string>>(() => new Set())
-  const { isOpen, currentPane, sshConnectionNames, saveError, isSaving, openSettings, closeSettings, saveSettings, addSSHConfigHost, detectShell } =
+  const { isOpen, currentPane, sshConnectionNames, saveError, isSaving, openSettings, closeSettings, saveSettings, addSSHConfigHost, detectShell, browseDirectories } =
     usePaneSettings(layout, updateSizes)
 
   const [isAddSSHHostOpen, setIsAddSSHHostOpen] = useState(false)
@@ -338,6 +338,7 @@ export const App: React.FC = () => {
           onClose={closeSettings}
           onAddSSHHost={() => setIsAddSSHHostOpen(true)}
           onDetectShell={detectShell}
+          onBrowseDirectories={browseDirectories}
         />
         <AddSSHHostDialog
           isOpen={isAddSSHHostOpen}
--- a/frontend/src/components/PaneSettingsDialog.test.tsx
+++ b/frontend/src/components/PaneSettingsDialog.test.tsx
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi } from 'vitest'
-import { fireEvent, render, screen } from '@testing-library/react'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
 import { PaneSettingsDialog } from './PaneSettingsDialog'
 
 const defaultProps = {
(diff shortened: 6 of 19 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Schema-first throughout**: `DirectoryEntrySchema` and `DirectoryBrowserResponseSchema` are defined in `schemas/index.ts` first; TypeScript types are derived via `z.infer<>`, and the runtime parse in `usePaneSettings.ts` guards against malformed API responses at the boundary.
- **Injectable directory listing functions**: `listLocalDirectoriesFn` and `listRemoteDirectoriesFn` on `Handler` make the new endpoint fully unit-testable without a real filesystem or SSH connection — the remote test swaps in a stub without any process overhead.
- **Explicit empty-slice initializer**: `Entries: []directoryEntryResponse{}` in `listLocalDirectories` ensures the JSON serialises as `[]` rather than `null` when a directory has no visible children, which matches the Zod schema's `z.array()` expectation and avoids a frontend null-check.
- **Reuses existing path-validation infrastructure**: `ListRemoteDirectories` calls the private `validateRemotePath` (the CodeQL-approved pattern) before any SSH dial, and the `shellQuotePath` helper wraps the user-supplied path in the remote shell command.
- **ESC key scoped correctly**: the `keydown` handler dismisses the directory browser first and only propagates to `onClose` when the browser is not visible — so a user can cancel browsing without losing edits.
- **Stale-closure avoided in `handleToggleHiddenDirectories`**: passes `nextValue` directly to `onBrowseDirectories` rather than reading the state variable that was just set, which would be stale before the next render.
- **Comprehensive test coverage**: the directory browser dialog tests cover selection, navigation (parent), hidden-directory toggle, expansion of has-children entries, and the empty-list case — not just the happy path.

---

### Issues & Questions

**[Question] `ValidateRemotePath` is exported but never called**

`internal/session/ssh.go` adds a public wrapper:
```go
// ValidateRemotePath applies the CodeQL-approved remote path validation rules.
func ValidateRemotePath(label, path string) error {
    return validateRemotePath(label, path)
}
```
Nothing in this PR (or the existing codebase) calls `session.ValidateRemotePath`. `GetDirectories` routes remote paths through `session.ListRemoteDirectories`, which calls the private `validateRemotePath` directly. Is this export intended for a follow-up consumer, or is it a leftover from an earlier draft? Per the project rules ("Don't add features beyond what the task requires"), it looks like dead code that could be removed.

**[Question] `localDirectoryHasChildren` propagates permission errors as a hard failure**

`internal/api/handler.go`:
```go
func localDirectoryHasChildren(path string) (bool, error) {
    entries, err := os.ReadDir(path)
    if err != nil {
        return false, fmt.Errorf("read dir %s: %w", path, err) // propagates up
    }
    ...
}
```
`listLocalDirectories` calls this for every visible subdirectory and fails the entire request on any error. On common paths like `/` or `/usr`, there are system-owned directories (`lost+found`, `proc` on Linux) that are unreadable by the current user. When a user tries to browse from `/`, the browser would return a 422 error instead of a partial listing. Would returning `false, nil` (or ignoring `os.ErrPermission` specifically) be more appropriate here?

**[Minor question] `resetDirectoryBrowser` declared after the early-return guard**

In `PaneSettingsDialog.tsx`, the function is declared as a `function` statement after the `if (!isOpen || !pane) return null` guard, but it is called in a `useEffect` that runs before that guard. This works because of JavaScript function hoisting, but placing the declaration before the early return (or above the `useEffect`) would make the execution order obvious without requiring readers to recall hoisting semantics.

---

### Overall

The directory browser is well-structured end-to-end: schema-first types, injectable functions for testability, server-side filtering for both local and remote paths, and solid test coverage across all behavioral dimensions. The two questions above (unused export and permission-error propagation) are worth confirming before merge; the minor point is a polish-only concern.

tomo-chan:
Addressed the follow-up review points in https://github.com/tomo-chan/panemux/commit/8316a09f1efec2057350c3b860c7b6a1106b2264.

- removed the unused exported `ValidateRemotePath` wrapper from `internal/session/ssh.go`
- changed local directory browsing to skip unreadable child directories instead of failing the whole listing
- added a handler test that covers a directory with one readable child and one permission-denied child

Verified locally with:
- `go test ./internal/api ./internal/session`
- `make lint-go`

tomo-chan (commented):
## Code Review (follow-up)

### Previous Review — Status

Both `[Question]` items from the first review are resolved; thank you.

- **`ValidateRemotePath` unused export** — removed in commit `8316a09`. ✓
- **`localDirectoryHasChildren` hard-fails on permission errors** — fixed: `listLocalDirectories` now catches `fs.ErrPermission` from `localDirectoryHasChildren` and `continue`s, skipping unreadable child directories. New test `TestGetDirectories_SkipsUnreadableChildDirectories` covers this. ✓

The `[Minor question]` about `resetDirectoryBrowser` being declared after the early-return guard is still present in `PaneSettingsDialog.tsx` — carrying it forward below.

---

### What's Good (new in this revision)

- **AGENTS.md path-sanitization rule**: the new section documenting that real environment paths must not appear in tests, fixtures, or commit history is a useful guard for this and future PRs — and the test fixtures were updated to follow it (`/workspace/user`, `/tmp/sample-project`).
- **`os.Stat` removed**: the refactored `listLocalDirectories` calls `readDir` once and derives existence/type from the error rather than issuing a separate `Stat` round-trip. Fewer syscalls and the `nolint` comment is gone.
- **Typed error discrimination**: using `errors.Is` / `errors.As` with `fs.ErrNotExist`, `fs.ErrPermission`, and `*fs.PathError` (rather than string-matching `err.Error()`) is the correct idiomatic Go pattern for error classification.

---

### Issues & Questions

**[Minor question] `var readDir = os.ReadDir` is a package-level global, inconsistent with the existing injectable pattern**

```go
// handler.go
var readDir = os.ReadDir
```

`listLocalDirectoriesFn` and `listRemoteDirectoriesFn` are injectable fields on `Handler`, which keeps all production injection in one place and is safe for parallel tests. `readDir` is a package-level variable mutated directly by `TestGetDirectories_SkipsUnreadableChildDirectories`:

```go
originalReadDir := readDir
readDir = func(name string) ([]os.DirEntry, error) { ... }
t.Cleanup(func() { readDir = originalReadDir })
```

This works today because the test doesn't call `t.Parallel()`. Is this intentional — a deliberate pragmatic trade-off to avoid changing `listLocalDirectories`'s signature — or would adding a `readDirFn` field to `Handler` (following the existing pattern) be worthwhile?

**[Minor question] `pathErr.Op == "readdir"` catch-all may be overly broad**

```go
var pathErr *fs.PathError
if errors.As(err, &pathErr) && errors.Is(pathErr.Err, fs.ErrInvalid) {
    return directoryBrowserResponse{}, errors.New("path is not a directory")
}
if errors.As(err, &pathErr) && pathErr.Op == "readdir" {
    return directoryBrowserResponse{}, errors.New("path is not a directory")
}
```

The second block catches *any* error whose `Op` is `"readdir"` and reports it as "path is not a directory". The intent seems to be catching `ENOTDIR` on Linux (which does not map to `fs.ErrInvalid`), but this would also surface as "path is not a directory" for, say, a network filesystem returning an unexpected error mid-read. A comment explaining the scenario — or tightening the condition to `errors.Is(pathErr.Err, syscall.ENOTDIR)` — would help future maintainers.

**[Minor question] `resetDirectoryBrowser` declared after the early-return guard (carried over)**

In `PaneSettingsDialog.tsx`, the `function resetDirectoryBrowser()` statement appears after `if (!isOpen || !pane) return null` but is called in a `useEffect` that runs before that guard. Works due to JS function hoisting, but placing the declaration before the early return would make the dependency order explicit.

---

### Overall

The refactor cleanly resolves both blocking questions from the first round. The remaining items are all minor polish; none should block merge.

tomo-chan:
Addressed the remaining minor follow-up items in https://github.com/tomo-chan/panemux/commit/077719df4fb259b22f7c8df0e0c8c1c97f5416a6.

- moved the local directory-reader injection from a package-level global onto `Handler.readDirFn` so tests follow the same injection pattern as the other handler dependencies
- added a clarifying comment for the `readdir` error classification used to detect non-directory paths
- moved `resetDirectoryBrowser` above the early-return guard in `PaneSettingsDialog.tsx` so the declaration order is explicit without relying on hoisting knowledge

Verified locally with:
- `go test ./internal/api ./internal/session`
- `make lint-go`
- `make lint-frontend`

tomo-chan (commented):
## Code Review (round 3 — final)

### Previous Review — Status

All three `[Minor question]` items from the second review are resolved. Thank you.

- **`var readDir = os.ReadDir` global** — moved to `Handler.readDirFn`, injected in `NewHandler`. `listLocalDirectories` and `localDirectoryHasChildren` are now methods on `Handler` using `h.readDirFn`, and the test swaps `h.readDirFn` directly — fully consistent with the existing injectable field pattern. ✓
- **`pathErr.Op == "readdir"` catch-all** — clarifying comment added: `// readDir on a plain file commonly reports a PathError with Op "readdir".` ✓
- **`resetDirectoryBrowser` placement** — moved above the early-return guard; the declaration order now makes the dependency explicit without relying on hoisting knowledge. ✓

---

### Overall

No new concerns. The PR is clean across all three rounds: schema-first types, injectable dependencies, permission-resilient local listing, test coverage for the edge cases, documentation updated, and real paths scrubbed from test fixtures. Ready to merge.

PR #93, opened 2026-05-16
feat: show linked pull requests in pane headers

tomo-chan:
## Summary

- add pull request URL lookup to the git info session API
- show a PR link in pane headers when the current branch has a linked GitHub pull request
- cover backend/frontend behavior with tests and document the new pane header shortcut

## Test Plan

- [x] make check

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,6 +175,12 @@ See `docs/architecture.md` → *Security Design* for the full rationale and the 
 
 **Remote path arguments (SSH working directory):** user-supplied paths that flow into `sess.Start()` shell commands must be validated with `validRemotePath` (defined in `internal/session/ssh.go`) before use. This regex guard is the CodeQL-recommended sanitization pattern for shell arguments, and it rejects shell metacharacters (`;|&$\`'"<>(){}[]!\`) and control characters while allowing valid Unix path characters including spaces and Unicode.
 
+### `gosec` suppressions
+- Do not add `//nolint:gosec` to production code as the default response to a finding.
+- First change the code so the finding is structurally avoided and the reason is evident in the implementation.
+- A `//nolint:gosec` suppression is acceptable only when the code is not shipped in the application, such as tests, fixtures, or other test-only helpers.
+- If a production-code suppression seems unavoidable, stop and document the case explicitly in review before merging instead of silently adding the suppression.
+
 ### Release workflow
 - **Never manually close a release-please PR.** Doing so leaves the `release-please--branches--main` internal tracking branch in a stale state. On the next push to `main`, release-please re-creates the release PR from that stale state, producing incorrect release notes that include all historical commits.
 - If you need to override the version release-please proposes (e.g., cut a patch instead of a minor):
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Common uses:
 - `tmux` and `ssh_tmux` panes automatically create the target tmux session if it does not already exist.
 - Set `cwd` on `local`, `ssh`, or `ssh_tmux` panes when you want the shell to start in a specific directory.
 - In the pane settings dialog, `Working Directory` can be chosen from a browsable directory tree for both local and SSH-backed panes. Hidden directories are available through a toggle.
+- Pane headers resolve Git and PR metadata from the live working context, including interactive `codex` and `claude` worktrees for both local and SSH-backed panes. `tmux` and `ssh_tmux` use the currently active tmux pane only.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Common uses:
 - **Move a pane** by dragging the handle in its header; drop it on a workspace edge to create a new outer split, or on another pane edge / divider to insert it there.
 - **Close a pane** from its header controls; the layout collapses automatically.
 - **Restart a pane** with the on-screen button if the underlying session exits.
-- **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. If an agent such as Codex or Claude is working in a separate `git worktree`, panemux prefers that worktree branch when it can detect the path from recent pane output.
+- **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. If an agent such as Codex or Claude is actively working in a separate `git worktree`, panemux prefers that agent process's current worktree and falls back to the pane's own working directory after the agent exits.
 - **Open VS Code** from supported panes using the pane header action.
 
 ### Notifications and attention prompts
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Common uses:
 - **Move a pane** by dragging the handle in its header; drop it on a workspace edge to create a new outer split, or on another pane edge / divider to insert it there.
 - **Close a pane** from its header controls; the layout collapses automatically.
 - **Restart a pane** with the on-screen button if the underlying session exits.
-- **Open a linked PR** from the pane header when the current local Git branch already has a GitHub pull request.
+- **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. If an agent such as Codex or Claude is working in a separate `git worktree`, panemux prefers that worktree branch when it can detect the path from recent pane output.
 - **Open VS Code** from supported panes using the pane header action.
 
 ### Notifications and attention prompts
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ Common uses:
 - **Move a pane** by dragging the handle in its header; drop it on a workspace edge to create a new outer split, or on another pane edge / divider to insert it there.
 - **Close a pane** from its header controls; the layout collapses automatically.
 - **Restart a pane** with the on-screen button if the underlying session exits.
+- **Open a linked PR** from the pane header when the current local Git branch already has a GitHub pull request.
 - **Open VS Code** from supported panes using the pane header action.
 
 ### Notifications and attention prompts
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Common uses:
 - **Move a pane** by dragging the handle in its header; drop it on a workspace edge to create a new outer split, or on another pane edge / divider to insert it there.
 - **Close a pane** from its header controls; the layout collapses automatically.
 - **Restart a pane** with the on-screen button if the underlying session exits.
-- **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. If an agent such as Codex or Claude is actively working in a separate `git worktree`, panemux prefers that agent process's current worktree and falls back to the pane's own working directory after the agent exits.
+- **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. For local panes, panemux can prefer the live worktree of an interactive `codex` or `claude` session, including resumed Codex sessions, and falls back to the pane's own working directory after the agent exits.
 - **Open VS Code** from supported panes using the pane header action.
 
 ### Notifications and attention prompts
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -161,15 +161,16 @@ Why this hook:
 `GET /api/sessions/{id}/git-info` resolves repository metadata in two stages:
 
 - first, it asks the session for its live working directory
-- second, for local and local tmux sessions only, it may ask for an active interactive agent workdir override
+- second, it may ask for an active interactive agent workdir override
 
 The override path is intentionally narrow:
 
 - only interactive `codex` and `claude` processes are considered
-- only descendants of the pane's own shell process or active tmux pane process are considered
+- local and local tmux sessions only consider descendants of the pane's own shell process or active tmux pane process
+- SSH sessions scan the remote process list on the current SSH connection, and SSH+tmux sessions restrict that scan to the active remote tmux pane's process tree
 - only worktrees that belong to the same Git common dir as the pane's base repository are accepted
 
-For resumed Codex sessions, the local session implementation can inspect the open Codex session log under `~/.codex/sessions/...jsonl` and prefer the latest `turn_context.cwd` when available. This exists because a resumed interactive Codex process may keep its OS-level process cwd at the original pane directory while the live Codex session itself is operating against a different worktree.
+For resumed Codex sessions, the local and SSH session implementations can inspect the open Codex session log under `~/.codex/sessions/...jsonl` and prefer the latest `turn_context.cwd` when available. This exists because a resumed interactive Codex process may keep its OS-level process cwd at the original pane directory while the live Codex session itself is operating against a different worktree.
 
 ### `useWorkspaceAttentionMonitor`
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -300,6 +300,8 @@ After validation, the path is wrapped with `shellQuotePath`, which single-quotes
 
 When adding new session types or new `exec.Command` calls: the value passed as the command (first argument) must come from a hardcoded literal or from a trusted system source (file, registry) with no data-flow path to user input. Arguments after the command may be user-supplied if they cannot be interpreted as commands by the target binary.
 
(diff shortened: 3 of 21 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Schema-first discipline**: `GitInfoSchema` in `frontend/src/schemas/index.ts` was extended before any component code, and `GitInfo` is derived via `z.infer<>` — consistent with the project rule.
- **Graceful degradation**: when `gh` is absent from PATH or the branch has no open PR, `lookupPRInfo` silently returns `("", 0)` and the header simply omits the link — no error surface for the user.
- **Defensive `Snapshot()` copy**: `managedSession.snapshot()` returns `append([]byte(nil), m.history...)`, a fresh allocation, so callers cannot accidentally mutate the manager's internal ring buffer.
- **Realistic mock I/O in tests**: upgrading `mockSession` to use a buffered channel for `Read`/`Write` lets `TestGetGitInfo_WorktreePathInSessionHistory_PrefersWorktreeBranch` inject real terminal content and verify the heuristic end-to-end — much stronger than a stub.
- **ANSI stripping before path extraction**: `worktreePathCandidates` calls `stripANSI` first, preventing escape sequences in Claude/Codex output from corrupting path matches.
- **`rel="noopener noreferrer"` on the `target="_blank"` link**: correct security practice in `InlineHeaderLink`.

---

### Issues & Questions

**[Question] Detached HEAD now returns `{is_git: false}` instead of `{is_git: true}`**

`inspectGitContext` treats *any* failure from `git rev-parse --show-current` as fatal and returns an error. In detached HEAD state that command exits non-zero. `GetGitInfo` then writes `{is_git: false}`:

```go
// internal/api/handler.go
ctx, err := h.inspectGitContext(gitPath, targetCWD)
if err \!= nil {
    writeJSON(w, gitInfoResponse{IsGit: false})  // was: IsGit: true, Repo: repo
    return
}
```

Previously the handler caught the branch-fetch error and returned `{is_git: true, repo: "..."}` with an empty branch. Any pane checked out at a specific commit (common in CI runners, `git bisect`, or a release tag checkout) would now silently lose its git context badge. Was this regression intentional, or should `inspectGitContext` treat a missing branch as non-fatal (returning a `gitContext` with an empty `branch`)?

---

**[Question] `lookupPRInfo` has no timeout — a slow or unreachable GitHub API hangs the request**

`cmd.Output()` blocks until `gh pr view` exits. If GitHub is unreachable or rate-limiting, the HTTP request to `/api/sessions/{id}/git-info` stalls indefinitely. Since the UI polls this endpoint, a hung network call would hold an open goroutine for the entire poll interval with no way for the client to recover.

Consider an `exec.CommandContext` with a short deadline:

```go
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()
cmd := exec.CommandContext(ctx, ghPath, "pr", "view", branch, "--json", "url,number")
cmd.Dir = cwd
```

---

**[Minor question] `inspectGitContext` is called twice for the common case (no worktree)**

`resolveGitInfoCWD` always calls `inspectGitContext(cwd)` to establish the baseline context. Then `GetGitInfo` calls `inspectGitContext(targetCWD)` — and when no worktree path is detected, `targetCWD == cwd`, so the same three-subprocess chain (`show-toplevel`, `git-common-dir`, `show-current`) runs a second time. For users without active agent worktrees, every git-info poll spawns six git subprocesses. Could `resolveGitInfoCWD` return the `gitContext` it already computed alongside the chosen CWD, so `GetGitInfo` can reuse it?

---

**[Minor question] `absolutePathPattern` silently truncates paths containing spaces**

```go
var absolutePathPattern = regexp.MustCompile(`/[A-Za-z0-9._\-/]+`)
```

A worktree at `/Users/tomo/my project/panemux-feature` would be extracted as `/Users/tomo/my`, causing `inspectGitContext` to silently fail for that candidate. The limitation is probably acceptable for most developer-machine paths, but a brief comment stating it intentionally excludes spaces would help future maintainers avoid adding space support incorrectly.

---

### Overall

The feature is well-structured, properly tested, and degrades gracefully. Two items are worth resolving before merge: the **detached HEAD behavioral regression** (a valid git pane silently becomes non-git) and the **missing timeout on `gh pr view`** (reliability risk under network issues). The other two observations are minor and could be addressed in a follow-up.

tomo-chan (commented):
## Code Review (round 2 — 5 new commits)

Picks up from [the previous review](https://github.com/tomo-chan/panemux/pull/93#pullrequestreview-4304522249). Most concerns from that review are addressed below.

### What Changed Since Round 1

The terminal-output heuristic (`worktreePathCandidates` / `absolutePathPattern`) was replaced entirely with a proper `ActiveWorkdirGetter` interface backed by live process inspection. This is a significant improvement: the approach is reliable, deterministic, and independently testable. The previous round's concern about `absolutePathPattern` truncating paths with spaces is now moot.

---

### Resolved from Round 1

**Double `inspectGitContext` call** — still two calls in the common case (once inside `resolveGitInfoCWD`, once in `GetGitInfo`), but the worst case is now 2×3=6 git subprocesses instead of the previous 1+(up to 8)×3. Acceptable.

---

### What's Good

- **Interface-based design**: `ActiveWorkdirGetter` keeps the API handler free of OS-specific logic and enables full unit testing of the resolution path without real processes.
- **Injectable function vars** (`listProcessesFn`, `getPIDCWDFn`, `openFilePathsForPIDFn`): the test suite can stub each independently, which is the right testability pattern for the project's rule against touching the real filesystem.
- **`isInteractiveAgentCommand` is precise**: explicitly filters `codex exec`, `claude -p`, and `claude --print` — matches the behavior.md spec exactly.
- **`sshSessionRunner` interface**: lets `activeRemoteWorkdir` be tested against a `fakeSSHRunner` without a real SSH connection.
- **Codex session log fallback**: `parseCodexSessionCWD` handles JSONL correctly (silently skips malformed lines, prefers the last `turn_context.cwd` over `session_meta.cwd`) — the ordering logic is well-tested.
- **Docs updated in the same PR**: `docs/architecture.md` and `docs/behavior.md` both received detailed explanations of the two-stage resolution algorithm.

---

### Issues & Questions

**[Question] Detached HEAD still returns `{is_git: false}` — carry-over**

Not addressed in the new commits. `inspectGitContext` returns an error when `git rev-parse --show-current` exits non-zero (which it does in detached HEAD), and `GetGitInfo` maps that error to `{is_git: false}`. The old code returned `{is_git: true, repo: "..."}` with an empty branch. Any pane checked out at a specific commit would silently lose its git context badge.

Should `inspectGitContext` treat `branchOut` failure as non-fatal, returning a `gitContext` with an empty `branch` instead of an error?

---

**[Question] `lookupPRInfo` has no timeout — carry-over**

Not addressed in the new commits. `cmd.Output()` blocks indefinitely on a slow or unreachable GitHub API. The UI polls `/api/sessions/{id}/git-info` periodically, so one hung goroutine per active pane is possible.

```go
// internal/api/handler.go
cmd := exec.Command(ghPath, "pr", "view", branch, "--json", "url,number")
cmd.Dir = cwd
out, err := cmd.Output()  // no deadline
```

---

**[Question] Plain SSH panes search _all_ remote processes for interactive agents**

`GetActiveWorkdir` for `SSHSession` calls `activeRemoteWorkdir(sess, baseCWD, 0)`, and with `rootPID == 0`, it calls `newestInteractiveAgentPID(processes)` — which scans every process on the remote host, not just descendants of the pane's shell.

On a shared remote host (a development server, a jump host), this would pick up another user's interactive `codex` or `claude` session and potentially show that user's PR in the current pane's header. Local panes and tmux panes are correctly scoped to descendants of the pane PID or the active tmux pane PID.

Is plain-SSH intended to be any-process-on-host, or should it be scoped (e.g., to the user's own UID) to match the narrower behaviour of the other session types?

---

**[Minor question] No end-to-end tests for `TmuxLocalSession.GetActiveWorkdir()` and `TmuxSSHSession.GetActiveWorkdir()`**

The helper functions (`newestInteractiveAgentDescendantPID`, `resolveInteractiveAgentWorkdir`, `activeRemoteWorkdir`) are well-tested. But the `TmuxLocalSession.GetActiveWorkdir()` and `TmuxSSHSession.GetActiveWorkdir()` methods themselves — which include the `tmux display-message` call and the wiring between the tmux pane PID and the process walk — have no unit tests. A fake tmux binary (like `writeFakeGHBinary`) or a stubbed `execCommandFn` could cover at least the parsing path.

---

### Overall

The redesign from terminal output scanning to live process inspection is the right call — the implementation is clean and the helper-function coverage is solid. Two items from round 1 remain open and are worth fixing before merge: the **detached HEAD regression** and the **missing timeout on `gh pr view`**. The SSH any-process-on-host scope question is a new one worth answering.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/ef639190b7342b670bfd14c6c3274aad6d4152b3

- Detached HEAD no longer downgrades the pane to `{is_git:false}`. `inspectGitContext` now keeps Git metadata and returns an empty branch when `git branch --show-current` has no branch name.
- `lookupPRInfo` now uses `exec.CommandContext` with a short timeout, so a slow or unreachable GitHub CLI call falls back cleanly instead of hanging the pane poll request.
- Plain `ssh` panes now scope interactive agent detection to the current SSH connection's interactive shell subtree instead of scanning every process on the remote host.

Local verification:

- `golangci-lint run ./internal/session/... ./internal/api/...`
- `make check`

There were no replyable inline review threads on the PR, so this follow-up is posted as a PR-level comment.

tomo-chan (commented):
## Code Review (round 3 — 3 new commits)

Picks up from [round 2](https://github.com/tomo-chan/panemux/pull/93#pullrequestreview-4304760111). The author addressed all three `[Question]` items from that round in commit `ef63919f` and the follow-up reply — thank you for the thorough write-up.

---

### Resolved from Round 2

**Detached HEAD regression** — fixed. `branchCmd.Output()` errors are now caught and silently treated as an empty branch, with an explanatory comment. `inspectGitContext` returns a valid `gitContext` rather than an error.

**`lookupPRInfo` timeout** — fixed. `exec.CommandContext` is now used with `prLookupTimeout` (5 s), injectable for tests via a package-level variable.

**Plain SSH pane scanning all remote processes** — fixed. `SSHSession.GetActiveWorkdir()` now calls `remoteShellPID(sess)` to get the interactive shell's PID first, then passes it as `rootPID` to `activeRemoteWorkdir`. Search is scoped to descendants of that shell, matching the local-pane behaviour.

---

### What's Good (new in this round)

- **`sanitizeGitExecDir` + `cmd.Dir`**: replacing `-C cwd` (tainted exec argument) with `cmd.Dir = safeCWD` (OS `chdir`) eliminates the CodeQL taint path for git commands entirely, without needing any nolint suppression. The `validGitExecDir` regex correctly restricts to absolute paths with no control characters — the right level for an OS path, not a shell argument.
- **`gosec` suppression policy in `AGENTS.md`**: the explicit rule ("fix the code, don't suppress") will prevent the same pattern from recurring in future PRs.
- **`validateTmuxSessionName` extracted**: the tmux target name is now validated through a shared function before being used in `exec.Command`, which is the correct structural fix for CodeQL rather than a nolint annotation.
- **`var prLookupTimeout`**: making the timeout injectable at the package level lets tests set an artificially short value; the `if timeout <= 0` guard ensures a floor of 5 s even if a test zeroes it.

---

### Issues & Questions

**[Minor question] `rootPID == 0` branch in `activeRemoteWorkdir` is now dead code**

After the plain-SSH scoping fix, `activeRemoteWorkdir` is called from two places:
- `SSHSession.GetActiveWorkdir()` → always passes the result of `remoteShellPID()`, which returns a positive PID or an error (never 0)
- `TmuxSSHSession.GetActiveWorkdir()` → always passes `panePID` from `tmux display-message`, also always positive

```go
// internal/session/ssh.go
if rootPID == 0 {
    agentPID, ok = newestInteractiveAgentPID(processes)  // unreachable
} else {
    agentPID, ok = newestInteractiveAgentDescendantPID(processes, rootPID)
}
```

Should the dead branch be removed, or is the `rootPID == 0` path being kept as a future extension point? If the latter, a comment explaining when a caller would pass 0 would prevent the next reader from wondering whether to remove it.

---

**[Minor question] No integration tests for `TmuxLocalSession.GetActiveWorkdir()` and `TmuxSSHSession.GetActiveWorkdir()` — carry-over from round 2**

Not mentioned in the author's follow-up. The helper functions (`newestInteractiveAgentDescendantPID`, `resolveInteractiveAgentWorkdir`, `activeRemoteWorkdir`) are covered. But the methods themselves — including the `tmux display-message` / `pane_pid` parsing path and the wiring to those helpers — have no tests.

For `TmuxSSHSession`, a `fakeSSHRunner` already exists in `ssh_test.go` and could drive an end-to-end test of `GetActiveWorkdir` via `activeRemoteWorkdir`. For `TmuxLocalSession`, `listProcessesFn` is already injectable, and a fake tmux binary (similar to `writeFakeGHBinary`) could cover the `display-message` call. Worth a follow-up issue if not addressed before merge.

---

### Overall

All blocking items from previous rounds are resolved. The `sanitizeGitExecDir` + `cmd.Dir` pattern and the gosec policy update are genuinely good hardening additions beyond what the reviews asked for. Two minor observations remain — the dead `rootPID == 0` branch is a one-line question and the tmux integration test gap is low risk given the underlying helpers are tested. Neither blocks merge.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/50a09b57cc35f22ee0b6cd58631bef4c0d3096b9

- Removed the dead `rootPID == 0` branch from `activeRemoteWorkdir`. All current callers now pass a concrete pane/shell PID, so the helper only follows descendant processes.
- Added wiring-level tests for the tmux paths:
  - `TestTmuxLocalSessionGetActiveWorkdir_UsesPanePIDAndBaseCWD`
  - `TestTmuxSSHActiveWorkdir_UsesPanePIDAndBaseCWD`

Local verification:

- `golangci-lint run ./internal/session/...`
- `make check`

There are still no replyable inline review threads on the PR, so this follow-up is posted as a PR-level comment.

tomo-chan (commented):
## Code Review (round 4 — 1 new commit)

Picks up from [round 3](https://github.com/tomo-chan/panemux/pull/93#pullrequestreview-4304800258). All open items are resolved.

---

### Resolved from Round 3

**Dead `rootPID == 0` branch** — removed. `activeRemoteWorkdir` now always calls `newestInteractiveAgentDescendantPID`, and `newestInteractiveAgentPID` (the any-process-on-host scan) was deleted entirely. The existing SSH tests were updated to pass a concrete root PID (100) matching the real call path.

**No integration tests for `TmuxLocalSession.GetActiveWorkdir()` and `TmuxSSHSession.GetActiveWorkdir()`** — both methods were refactored so their logic lives in standalone helpers (`tmuxLocalActiveWorkdir`, `tmuxSSHActiveWorkdir`) that accept injectable deps. `TestTmuxLocalSessionGetActiveWorkdir_UsesPanePIDAndBaseCWD` drives the full local-tmux path with stubbed `tmuxLocalOutputFn`, `listProcessesFn`, and `getPIDCWDFn`. `TestTmuxSSHActiveWorkdir_UsesPanePIDAndBaseCWD` drives the SSH-tmux path through a `fakeSSHRunner`.

---

### What's Good (new in this round)

- **`tmuxLocalOutputFn` injectable**: both `GetCWD()` and `GetActiveWorkdir()` on `TmuxLocalSession` now go through this single injectable, so tests can stub all tmux output without a real tmux process.
- **`TmuxSSHSession.GetActiveWorkdir()` reuses the open SSH session**: instead of calling `s.GetCWD()` (which opens a second SSH session) to get the base CWD, `tmuxSSHActiveWorkdir` queries `pane_current_path` through the `runner` already in scope. One fewer SSH round-trip, and the function is now fully testable with `fakeSSHRunner`.

---

### Issues & Questions

None.

---

### Overall

All concerns from every round have been addressed. The feature is well-tested, structurally sound, and the codebase is cleaner for it (gosec suppressions removed, dead code removed, testability improved). Ready to merge from a review perspective.

PR #94, opened 2026-05-17
fix: open VSCode in active agent worktree

tomo-chan:
## Summary

- prefer the active interactive agent worktree when Open in VSCode is triggered from a pane in the same repository
- reuse the existing repository-context resolution logic for both Open in VSCode and pane git/PR metadata
- document the behavior in the README and architecture notes

## Why

When an agent is working from a sibling git worktree, opening VSCode from the pane header should follow that active worktree instead of reopening the pane's base repository directory.

## Test plan

- [x] go test ./internal/api -run TestPostOpenVSCode_ActiveAgentWorkdir_PrefersWorktree -count=1
- [x] make install-deps-ci
- [x] make build-frontend
- [x] make test-go
- [x] make lint-go

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Common uses:
 - **Close a pane** from its header controls; the layout collapses automatically.
 - **Restart a pane** with the on-screen button if the underlying session exits.
 - **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. For local panes, panemux can prefer the live worktree of an interactive `codex` or `claude` session, including resumed Codex sessions, and falls back to the pane's own working directory after the agent exits.
-- **Open VS Code** from supported panes using the pane header action.
+- **Open VS Code** from supported panes using the pane header action. Like pane Git/PR metadata, this can prefer the live worktree of an interactive `codex` or `claude` session and falls back to the pane's own working directory after the agent exits.
 
 ### Notifications and attention prompts
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -78,7 +78,7 @@ Workspace-related endpoints:
 - `PUT /api/workspaces/{id}/layout` updates a specific workspace layout.
 - `GET/PUT /api/layout` remain as compatibility endpoints for the active workspace layout.
 
-`POST /api/sessions/{id}/open-vscode` launches VSCode pointed at the session's live working directory. For local sessions it runs `code <cwd>`; for SSH sessions it runs `code --remote ssh-remote+<connection> <cwd>`. The binary is located via `exec.LookPath("code")` with a macOS app-bundle fallback.
+`POST /api/sessions/{id}/open-vscode` launches VSCode pointed at the session's live working directory. Like `GET /api/sessions/{id}/git-info`, it may prefer the worktree of an active interactive `codex` or `claude` process when that worktree belongs to the same repository. For local sessions it runs `code <cwd>`; for SSH sessions it runs `code --remote ssh-remote+<connection> <cwd>`. The binary is located via `exec.LookPath("code")` with a macOS app-bundle fallback.
 
 Why REST here:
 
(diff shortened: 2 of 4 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Clean rename / reuse**: Renaming `resolveGitInfoCWD` → `resolvePreferredCWD` with no logic change is the right call. It surfaces the shared intent without adding any duplication, and the diff is minimal.
- **Correct placement**: `resolvePreferredCWD` is called *before* `validateVSCodeCWD`, so the existence check is applied to the final resolved path (the worktree, when preferred). If the worktree is stale, `inspectGitContext` inside `resolvePreferredCWD` fails and falls back to the original `cwd` before any check happens.
- **Test mirrors existing pattern**: The new `TestPostOpenVSCode_ActiveAgentWorkdir_PrefersWorktree` follows the same structure as `TestGetGitInfo_ActiveAgentWorkdir_PrefersWorktreeBranch`, using the same `initTempGitRepo` / `addTempGitWorktree` helpers.
- **Documentation updated in the same change**: Both `README.md` and `docs/architecture.md` describe the new behavior inline with the implementation. No follow-up docs task left open.

---

### Issues & Questions

**[Minor question] Missing "ended agent" fallback test for `PostOpenVSCode`**

`GetGitInfo` has symmetric test coverage for both branches of `resolvePreferredCWD`:

| Test | `activeWorkdir` | `cwd` | Expected |
|---|---|---|---|
| `TestGetGitInfo_ActiveAgentWorkdir_PrefersWorktreeBranch` | valid worktree | base repo | worktree preferred |
| `TestGetGitInfo_EndedAgentFallsBackToPaneCWD` | `""` (agent ended) | base repo | original `cwd` |

`PostOpenVSCode` gains only the happy-path variant. The existing `TestPostOpenVSCode_Local_200` does have `activeWorkdir: ""`, but its `cwd` is `t.TempDir()` (not a git repo), so `resolvePreferredCWD` short-circuits at the first `inspectGitContext` call rather than reaching the empty-worktree branch. The "pane is in a git repo, agent has exited → falls back to pane `cwd`" path is not covered for `PostOpenVSCode`.

Since the function is shared and already exercised by the `GetGitInfo` suite, this is low-risk — but a corresponding `TestPostOpenVSCode_EndedAgent_FallsBackToPaneCWD` would make the contract symmetric and discoverable. Worth adding, or intentionally omitting?

---

### Overall

Minimal, well-scoped change that correctly reuses the existing worktree-resolution logic for `PostOpenVSCode`. The only open point is whether to add the symmetric fallback test; everything else looks good.

tomo-chan:
Added the symmetric fallback coverage in:

https://github.com/tomo-chan/panemux/commit/015a7ae78ce1ca9e7b9556d7b271b5df2a9936ad

This adds `TestPostOpenVSCode_EndedAgentFallsBackToPaneCWD`, which exercises the git-repo case where `activeWorkdir` is empty and `PostOpenVSCode` should fall back to the pane's own `cwd`.

Verified with:

- `go test ./internal/api -run 'TestPostOpenVSCode_(ActiveAgentWorkdir_PrefersWorktree|EndedAgentFallsBackToPaneCWD)' -count=1`
- `make test-go`
- `make lint-go`

tomo-chan (commented):
## Code Review (follow-up)

### Resolved

**[Minor question] Missing "ended agent" fallback test for `PostOpenVSCode`** — resolved by `015a7ae`.

`TestPostOpenVSCode_EndedAgentFallsBackToPaneCWD` is now symmetric with `TestGetGitInfo_EndedAgentFallsBackToPaneCWD`: it uses a real git repo as `cwd` and an empty `activeWorkdir` to exercise the fallback branch of `resolvePreferredCWD`. Test coverage is complete.

---

### Overall

No remaining concerns. The PR is clean: minimal production change, correct placement of `resolvePreferredCWD` before validation, symmetric test coverage for both branches, and docs updated in the same change.

PR #95, opened 2026-05-17
fix: resolve codex worktree git context across session types

tomo-chan:
## Summary
- prefer the latest Codex exec-command workdir when deriving pane git context
- cover local, tmux, ssh, and ssh+tmux session types with regression tests
- document the observed Codex session-log fields and resolver priority for future comparisons

## Test Plan
- [x] make lint-go
- [x] make test-go
- [x] make coverage-go
- [x] make check

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -170,7 +170,15 @@ The override path is intentionally narrow:
 - SSH sessions scan the remote process list on the current SSH connection, and SSH+tmux sessions restrict that scan to the active remote tmux pane's process tree
 - only worktrees that belong to the same Git common dir as the pane's base repository are accepted
 
-For resumed Codex sessions, the local and SSH session implementations can inspect the open Codex session log under `~/.codex/sessions/...jsonl` and prefer the latest `turn_context.cwd` when available. This exists because a resumed interactive Codex process may keep its OS-level process cwd at the original pane directory while the live Codex session itself is operating against a different worktree.
+For interactive Codex sessions, all four session implementations (`local`, `ssh`, `tmux`, and `ssh_tmux`) may inspect the open Codex session log under `~/.codex/sessions/...jsonl` when the Codex process has that file open. Panemux currently prefers the latest `exec_command.arguments.workdir` recorded in `response_item` / `function_call` log entries, then falls back to `turn_context.cwd`, then `session_meta.cwd`.
+
+This ordering is intentional and reflects observed Codex behavior as of `codex-tui` `0.130.0`:
+
+- the interactive Codex process may keep its OS-level process cwd at the original pane directory
+- `session_meta.cwd` and later `turn_context.cwd` may also remain pinned to that original pane directory
+- individual tool calls still record their actual execution directory in `exec_command.arguments.workdir`
+
+Panemux treats that `workdir` field as the strongest available signal for the active worktree because it is the only one that changes when Codex executes tools inside a sibling Git worktree while the parent interactive process remains attached to the original pane directory. If a future Codex release changes this logging contract, compare new logs against these three fields before changing the resolver so behavior remains reviewable and intentional.
 
 ### `useWorkspaceAttentionMonitor`
 
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -457,7 +457,10 @@ When a pane moves to a different parent node, the component may be remounted by 
 - When a local, local tmux, SSH, or SSH+tmux pane has an active interactive `codex` or `claude` process working in a different Git worktree for the same repository, panemux prefers that agent worktree over the pane's base directory.
 - Only interactive agents are eligible for worktree override.
 - Non-interactive commands such as `codex exec`, `claude -p`, and `claude --print` must not affect the displayed Git or PR metadata.
-- For `codex resume` flows, panemux may derive the active worktree from the Codex session log metadata (`turn_context.cwd`, falling back to `session_meta.cwd`) when the process has the session log open.
+- For interactive Codex flows across all four pane types (`local`, `ssh`, `tmux`, and `ssh_tmux`), panemux may derive the active worktree from the Codex session log when the process has that log open.
+- Codex log resolution currently prefers the most recent `response_item.payload.arguments.workdir` from `exec_command` tool calls, then falls back to `turn_context.cwd`, then `session_meta.cwd`.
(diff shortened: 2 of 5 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Clean helper extraction**: `parseCodexSessionCWD` is broken into three focused helpers — `codexSessionRecordCWD` (dispatch), `parsePayloadCWD` (cwd fields), and `parseExecCommandWorkdir` (exec_command arguments). Each does exactly one thing.
- **Double-JSON parsing handled correctly**: `json.RawMessage` for the outer payload, then a second `json.Unmarshal` on the string-valued `Arguments` field, which correctly models the Codex log format where function arguments are stored as a JSON-encoded string inside the outer JSON.
- **Silent-error pattern preserved**: all parse errors return `""` rather than propagating, so malformed log lines don't break the resolver.
- **Symmetric end-to-end test coverage**: all four session types (local, tmux/local, ssh, ssh+tmux) each get a `PrefersCodexExecCommandWorkdir` test, mirroring the existing `PrefersTurnContext` and `FallsBackToSessionMeta` coverage.
- **Documentation quality**: both `docs/architecture.md` and `docs/behavior.md` now document the field priority order with rationale *and* a forward guidance note about what to compare if a future Codex release changes the logging contract.

---

### Issues & Questions

**[Minor question] `codexSessionRecord` as a local type forces an anonymous struct in `codexSessionRecordCWD`'s parameter**

`codexSessionRecord` is defined inside `parseCodexSessionCWD`, so `codexSessionRecordCWD` can't reference it by name and repeats the same struct shape as an anonymous parameter:

```go
// local to parseCodexSessionCWD — not accessible outside
type codexSessionRecord struct {
    Type    string          `json:"type"`
    Payload json.RawMessage `json:"payload"`
}

// helper must duplicate the shape as anonymous struct
func codexSessionRecordCWD(rec struct {
    Type    string          `json:"type"`
    Payload json.RawMessage `json:"payload"`
}) ...
```

Go assigns a `codexSessionRecord` value to the anonymous parameter correctly (named type is assignable to an anonymous struct with the same underlying type), so this compiles and works. Any field drift between the two definitions would also surface as a compile error, so it's self-correcting. That said, an anonymous struct parameter is unusual and harder to document. Would elevating `codexSessionRecord` to package scope (outside both functions) be acceptable, so both can reference the same named type?

**[Minor question] `codexExecCommandRecord` test helper builds JSON via string concatenation**

```go
func codexExecCommandRecord(cmd, workdir string) string {
    return "" +
        "{\"type\":\"response_item\",\"payload\":{" +
        // ...
        "\"arguments\":\"{\\\"cmd\\\":\\\"" + cmd + "\\\",\\\"workdir\\\":\\\"" + workdir + "\\\"}\"" +
        "}}\n"
}
```

All current test values are simple paths with no special characters, so the output is valid JSON today. If a future test passes a value containing `"`, `\`, or a control character, the generated JSON will be malformed, `parseExecCommandWorkdir` will silently return `""`, and the test will fail with a confusing assertion mismatch rather than a JSON syntax error. Using `encoding/json.Marshal` to construct the inner and outer values would produce correct JSON for any input and surface failures more clearly.

**[Minor question] No test for `exec_command` appearing before a later `turn_context`**

`TestReadCodexSessionCWD_PrefersLatestExecCommandWorkdir` has exec_command entries appearing *after* turn_context:

```
session_meta cwd=/repo/main
turn_context cwd=/repo/main
exec_command workdir=/tmp/worktree-a
exec_command workdir=/tmp/worktree-b
```

The docs and code comment assert that `exec_command.workdir` wins *absolutely*, not just when it appears last, because `turn_context.cwd` stays pinned to the original directory. A complementary test with the inverse ordering:

```
exec_command workdir=/tmp/worktree
turn_context cwd=/repo/main   ← appears after exec_command
```

…would explicitly confirm the priority is order-independent and document the contract for future readers. Worth adding, or is the documentation explanation considered sufficient here?

---

### Overall

Solid, well-structured change that correctly extends the Codex log field priority across all four session types with symmetric test coverage and clear documentation. The three points above are all non-blocking — the most actionable is the JSON-construction concern in the test helper (#2), and #3 is a quick additional test case that would strengthen the contract coverage.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/55c46ca288f2b28c7ea62f54d2f18a3602c324a3.

- moved `codexSessionRecord` to a shared named type so the helper no longer needs an anonymous struct parameter
- changed the test helper to build both the outer payload and nested `arguments` JSON via `encoding/json` instead of string concatenation
- added a regression test proving `exec_command.workdir` still wins even when a later `turn_context.cwd` appears in the session log

Local verification run:
- `make lint-go`
- `make test-go`
- `make coverage-go`
- `make check`

tomo-chan (commented):
## Code Review (follow-up)

### Resolved

All three [Minor question] items from the previous round are addressed by `55c46ca`.

**[Minor question] `codexSessionRecord` as a local type** — resolved.

`codexSessionRecord` is now a package-level type. `codexSessionRecordCWD` uses the named type directly in its parameter:

```go
func codexSessionRecordCWD(rec codexSessionRecord) (execWorkdir, turnCWD, metaCWD string) {
```

No duplication, no anonymous struct. Clean.

---

**[Minor question] `codexExecCommandRecord` builds JSON via string concatenation** — resolved.

The helper is fully rewritten with `encoding/json.Marshal`, proper struct types, and two new test utilities (`mustJSONLine`, `mustRawJSON`). Taking `t *testing.T` as the first argument means a marshal failure surfaces as a `require.NoError` failure with a clear message rather than a silent bad-JSON value. The helper also now reuses the production `codexSessionRecord` type, so any schema drift will produce a compile error instead of a silent test-data mismatch.

---

**[Minor question] No test for `exec_command` appearing before a later `turn_context`** — resolved.

`TestReadCodexSessionCWD_PrefersExecCommandWorkdirEvenWhenTurnContextAppearsLater` directly exercises this:

```
session_meta cwd=/repo/main
exec_command workdir=/tmp/worktree-from-command
turn_context cwd=/repo/main   ← appears after exec_command
```

Expected result: `/tmp/worktree-from-command`. This explicitly documents the order-independent priority guarantee.

---

### Overall

No remaining concerns. All open points are closed cleanly and the test suite is stronger for it.

PR #97, opened 2026-05-17
feat: allow pane moves across workspaces

tomo-chan:
## Summary
- allow dragging a pane onto another workspace tab to move it across workspaces
- persist both source and destination workspace layouts and switch to the destination workspace after the move
- add regression coverage for cross-workspace moves and document the behavior

## Test Plan
- [x] make lint-frontend
- [x] make test-frontend

### DIFF
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -431,6 +431,7 @@ When a pane moves to a different parent node, the component may be remounted by 
 - The original pane keeps its current visible terminal contents when split; it must not go blank or reset to a fresh prompt while the new sibling pane is created.
 - Closing a pane calls `DELETE /api/sessions/{id}`, removes the pane from the tree, collapses parents with a single child, and renormalizes sizes to total `100`.
 - Moving a pane calls the workspace layout save path only; it does not call session create or delete APIs.
+- Dropping a pane on another workspace tab moves it into that workspace, persists both affected workspace layouts, and switches the active workspace to the destination.
 
 ### New terminal creation
 
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -431,7 +431,7 @@ When a pane moves to a different parent node, the component may be remounted by 
 - The original pane keeps its current visible terminal contents when split; it must not go blank or reset to a fresh prompt while the new sibling pane is created.
 - Closing a pane calls `DELETE /api/sessions/{id}`, removes the pane from the tree, collapses parents with a single child, and renormalizes sizes to total `100`.
 - Moving a pane calls the workspace layout save path only; it does not call session create or delete APIs.
-- Dropping a pane on another workspace tab moves it into that workspace, persists both affected workspace layouts, and switches the active workspace to the destination.
+- Dropping a pane on another workspace tab moves it into that workspace, inserts it at the destination workspace's right edge, persists both affected workspace layouts, and switches the active workspace to the destination.
 
 ### New terminal creation
 
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -43,7 +43,7 @@ const mockSplitPane = vi.fn()
 const mockClosePane = vi.fn()
 const mockSwapPanes = vi.fn()
 const mockCreatePane = vi.fn().mockResolvedValue(undefined)
-const mockMovePane = vi.fn()
+const mockMovePane = vi.fn().mockResolvedValue(undefined)
 const mockAddWorkspace = vi.fn()
 const mockUseWorkspaceAttentionMonitor = vi.hoisted(() => vi.fn())
 const mockUseBrowserNotificationPermission = vi.hoisted(() => vi.fn())
@@ -295,6 +295,20 @@ describe('App workspace deletion', () => {
     fireEvent.click(screen.getByRole('button', { name: 'Place workspace tabs on right' }))
 
     expect(mockSetWorkspaceTabPosition).toHaveBeenCalledWith('right')
+  })
+
+  it('moves a dragged pane to another workspace tab', () => {
+    mockTerminalPane.mockImplementation(({ pane }: { pane: { id: string } }) => {
+      const ctx = useContext(LayoutActionsContext)
+      return <button onMouseDown={() => ctx?.setDragSourcePaneId(pane.id)}>Start drag {pane.id}</button>
+    })
+
+    render(<App />)
+    fireEvent.mouseDown(screen.getByRole('button', { name: 'Start drag main' }), { button: 0 })
+    fireEvent.mouseEnter(screen.getByRole('tab', { name: 'Ops' }))
+    fireEvent.mouseUp(screen.getByRole('tab', { name: 'Ops' }))
+
+    expect(mockMovePane).toHaveBeenCalledWith('main', { type: 'workspace-tab', workspaceId: 'ops' })
   })
 
   it('shows a user-visible error when moving a pane fails to persist', async () => {
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -10,7 +10,7 @@ import { useBrowserNotificationPermission } from './hooks/useBrowserNotification
 import { DisplayConfig } from './types'
 import { TERMINAL_FONT_FAMILY } from './utils/fonts'
 import { findPaneById, generatePaneId, layoutContainsPane } from './utils/layoutTree'
-import type { PanePlacement } from './hooks/useLayout'
+import type { MovePanePlacement } from './hooks/useLayout'
 import type { LayoutChild, LayoutNode, SSHConfigHost } from './schemas'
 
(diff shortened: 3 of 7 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Type refactoring** — splitting the old `PanePlacement` into `CreatePanePlacement` and `MovePanePlacement` is semantically precise. `createPane` and `movePane` now enforce at the type level that only `movePane` accepts a workspace-tab destination.
- **Last-pane guard** — `removePaneFromTree` returning `null` is used as the sentinel to throw `'Cannot move the last pane out of a workspace'`. Clean, no magic number needed.
- **Dual drag protocol** — the `onDragOver`/`onDrop` + `onMouseEnter`/`onMouseUp` pairing is consistent with the existing pane-edge drop targets, so keyboard-less pointer drags work the same way across the board.
- **`isWorkspaceDropTarget` gate** — computing the boolean once per render and checking it at every handler entry point keeps the event logic readable and avoids processing drops on the active workspace or while editing a title.
- **`saveWorkspaceLayout` as a module-level function** — avoids capturing stale closure state and uses `encodeURIComponent` on the workspace ID. Good practice.
- **Defensive click guard** — `if (dragSourcePaneId) return` in the tab's `onClick` prevents an accidental workspace switch when the user releases a drag over a tab.
- **Test coverage** — unit tests for `WorkspaceTabs` cover both DnD and pointer-release paths. `useLayout` tests cover the happy path and the last-pane rejection. The App-level integration test confirms end-to-end wiring. `docs/behavior.md` is updated in the same commit.

---

### Issues & Questions

**[Question] No rollback on partial save failure**

```typescript
// useLayout.ts — workspace-tab branch of movePane
setWorkspaces(updatedWorkspaces)
setLayout(targetWithPane)
await saveWorkspaceLayout(sourceWorkspace.id, sourceWithoutPane)  // ← if this succeeds…
await saveWorkspaceLayout(targetWorkspace.id, targetWithPane)      // ← …and this throws
```

State is updated optimistically before either save. If the first save succeeds (pane removed from source in the backend) and the second save throws (pane not added to destination in the backend), the pane has effectively disappeared from the backend on reload — the user would lose it entirely.

The existing within-workspace `movePane` path has only one `saveLayout` call, so there is no partial-failure window there. This is the first code path with two sequential saves, introducing a new failure mode.

Was intentional no-rollback the right call here, or should the implementation restore `workspaces` and `layout` to their pre-move snapshot on catch?

---

**[Minor question] Stale `hoveredDropWorkspaceId` when a pointer drag is cancelled mid-hover**

```typescript
// WorkspaceTabs.tsx
backgroundColor: hoveredDropWorkspaceId === workspace.id   // ← not gated on dragSourcePaneId
  ? 'rgba(86, 156, 214, 0.24)'
  : active ? '#2f3540' : 'transparent',
```

If `dragSourcePaneId` is reset to `null` externally (e.g., a global `mouseup` handler on the document) while the cursor is still inside a tab, `hoveredDropWorkspaceId` stays set and that tab keeps the blue highlight until the next `onMouseLeave`. Is there an existing global cancel handler that guarantees `onMouseLeave` fires — or `dragSourcePaneId` clears — before this state can go stale? If not, gating the highlight on `Boolean(dragSourcePaneId)` as well would prevent the stale tint.

---

**[Minor question] Cross-workspace pane always inserted at the `'right'` edge**

```typescript
const targetWithPane = insertPaneAtWorkspaceEdge(targetWorkspace.layout, 'right', sourcePane)
```

`'right'` is hardcoded. `docs/behavior.md` says "moves it into that workspace" without specifying where. Is this an intentional "always append right" default, or was there a plan to infer the insertion edge from where on the tab the pane was dropped? If the former, documenting it in `docs/behavior.md` would help future contributors.

---

### Overall

Clean feature with good test coverage and consistent dual-protocol drag support. The main open question before merge is whether the two-save sequence should roll back state on partial failure — it is the first `movePane` path where a save can succeed partway through and leave the backend in an inconsistent state.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/f6f4106ccd869181e37140236c20c036cb1b154a.

- Added compensating rollback for cross-workspace moves: if the destination workspace save fails after the source save succeeded, the frontend restores its previous state and re-saves the original source layout to avoid losing the pane on reload.
- Cleared stale workspace-tab drop highlighting when drag state is cancelled externally by resetting the hover state when `dragSourcePaneId` becomes null, and covered that case with a regression test.
- Documented that cross-workspace drops currently insert at the destination workspace right edge.

Re-ran `make lint-frontend` and `make test-frontend` locally.

tomo-chan (commented):
## Code Review (follow-up — commit f6f4106)

All three issues raised in the previous review have been addressed. Summarising before the new finding.

---

### Prior issues — resolved

**Partial save failure / no rollback → resolved.**
The author replied: *"Added compensating rollback for cross-workspace moves: if the destination workspace save fails after the source save succeeded, the frontend restores its previous state and re-saves the original source workspace layout to undo the server-side change."*
The implementation in `useLayout.ts` captures `previousWorkspaces` / `previousLayout` snapshots before the optimistic update, wraps both saves in a `try/catch`, and on error restores state and re-saves the source. The new test `'rolls back frontend state and restores the source workspace when the destination save fails'` confirms the full rollback chain (source save → destination save → rollback source save) and verifies the restored frontend state.

**Stale `hoveredDropWorkspaceId` → resolved.**
`WorkspaceTabs.tsx` adds a `useEffect` that clears `hoveredDropWorkspaceId` whenever `dragSourcePaneId` becomes falsy. The highlight style is also gated on `dragSourcePaneId &&`, so stale state can no longer produce a visible highlight. A new test (`'clears the workspace drop highlight when dragging is cancelled externally'`) rerenders with `dragSourcePaneId={null}` and asserts the style is removed.

**Fixed `'right'` insertion edge undocumented → resolved.**
`docs/behavior.md` now reads: *"inserts it at the destination workspace's right edge"*.

---

### New finding

**[Minor question] Rollback save failure is silently swallowed**

```typescript
// useLayout.ts
try {
  await saveWorkspaceLayout(sourceWorkspace.id, sourceWithoutPane)
  await saveWorkspaceLayout(targetWorkspace.id, targetWithPane)
} catch (err) {
  setWorkspaces(previousWorkspaces)
  setLayout(previousLayout)
  try {
    await saveWorkspaceLayout(sourceWorkspace.id, sourceWorkspace.layout)
  } catch (rollbackErr) {
    console.error(rollbackErr)   // ← swallowed
  }
  throw err
}
```

When the rollback save itself fails (e.g., network is down after the source save succeeded), the user sees the original error ("Something went wrong") but has no indication that the source workspace backend is now in an inconsistent state — the pane has been removed server-side but the frontend shows it restored. On next page reload, the user would lose the pane silently.

`console.error` is appropriate as a last-resort log, but it might be worth surfacing a distinguishable error to the user in this case — something like "Move failed and rollback also failed. Reload to sync." — so they know to refresh rather than assume the error was harmless.

Is intentional silent swallowing the right call here, given that it can mask silent data loss, or would a separate error state (e.g., a `rollbackFailed` flag) be worth adding?

---

### Overall

The feature is well-executed, the rollback is correct and fully tested, and all prior concerns are addressed. The remaining open question is whether a rollback-save failure should surface a distinct user-visible message. This does not block merge.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/476c8c23e4b295e10eab3278cc3b72b4516e8664.

- If the compensating rollback save also fails after a cross-workspace move error, `movePane` now throws a distinct user-visible error: `Move failed and rollback also failed. Reload to sync.` instead of silently collapsing back to the original destination-save error.
- Added a regression test that covers the double-failure path and verifies the distinct error while preserving the restored frontend state.

Re-ran `make lint-frontend` and `make test-frontend` locally.

tomo-chan (commented):
## Code Review (follow-up — commit 476c8c23)

### Prior issue — resolved

**Rollback save failure silently swallowed → resolved.**
The author replied: *"If the compensating rollback save also fails after a cross-workspace move error, `movePane` now throws a distinct user-visible error: `Move failed and rollback also failed. Reload to sync.`"*

The implementation is correct:

```typescript
} catch (rollbackErr) {
  console.error(rollbackErr)
  throw new Error('Move failed and rollback also failed. Reload to sync.')
}
throw err
```

When rollback succeeds, the original save error is re-thrown and the user sees "Something went wrong." When rollback itself fails, a more alarming but actionable message is thrown instead, correctly signalling that the backend may be inconsistent. The new test `'throws a distinct error when both the destination save and rollback save fail'` verifies the message, confirms `console.error` is called, and asserts the frontend state is still restored.

---

### No new findings

The full diff across all three commits is in good shape. No new issues to raise.

---

### Overall

All concerns from both prior reviews are addressed. The rollback error path, the stale-highlight cleanup, and the `'right'`-edge documentation are all complete and tested. Ready for merge.

PR #101, opened 2026-05-18
docs: reorganize agent and development docs

tomo-chan:
## Summary
- turn `AGENTS.md` into a short index for agents
- move developer workflow guidance into `DEVELOPMENT.md`
- add `docs/maintenance.md` for CI and release maintenance rules

## Test plan
- [x] Reviewed `AGENTS.md` to confirm it now points to the relevant long-form documents instead of duplicating them
- [x] Reviewed `DEVELOPMENT.md` to confirm build, test, branch, and PR workflow rules were moved there
- [x] Reviewed `docs/maintenance.md` to confirm GitHub Actions pinning and release-please guidance were moved under `docs/`


tomo-chan (commented):
## Code Review

### What's Good

- **Clean three-way split.** `AGENTS.md` as index, `DEVELOPMENT.md` for workflow, `docs/maintenance.md` for CI/ops — each document has a clear, non-overlapping scope and a clear ownership rule in the new `## Editing Rules` section. That last bit is especially useful for preventing content from drifting back into the index over time.
- **No information lost in the security sections.** Verified that `docs/architecture.md` already contains the detailed CodeQL compliance rules, `gosec` suppression policy, and `validRemotePath` documentation (lines ~291–315). The reorganization correctly removes the duplication from `AGENTS.md` rather than dropping content.
- **`docs/maintenance.md` adds rationale that wasn't there before.** The three "Why this matters" bullets under GitHub Actions pinning (reproducibility, supply-chain drift, review explicitness) are new and improve the document beyond just moving content.
- **Document map with labeled categories** in `AGENTS.md` is easy to scan and gives agents a quick dispatch table without needing to read everything.

---

### Issues & Questions

**[Question] `CLAUDE.md` still only loads `@AGENTS.md` — will agents see the development rules?**

`CLAUDE.md` contains a single `@AGENTS.md` directive, which causes Claude Code to inline `AGENTS.md` as project context at startup. After this PR, `AGENTS.md` becomes an index that *points to* `DEVELOPMENT.md` but does not include it. An agent loading project context will see the index and the "Read This First" numbered list, but the TDD rules, schema-first rules, quality gate, and test granularity guidance won't be in the automatically-loaded context — they'll only be read if the agent explicitly follows the link.

Was this intentional (relying on agents to follow the instructions in the index), or should `CLAUDE.md` be updated to also `@DEVELOPMENT.md` so the rules are loaded automatically the same way they were before?

**[Question] Worktree path changed from `../<repo>-<feature>` to `/tmp/<repo>-<feature>`**

`DEVELOPMENT.md` — `## Branch And PR Workflow` section:

```sh
# new
git worktree add /tmp/<repo>-<feature> -b feature/<name>

# original AGENTS.md and global ~/.claude/CLAUDE.md
git worktree add ../<repo>-<feature> -b feature/<name>
```

`/tmp` is cleaned on reboot on many systems and is a different organisational pattern from the sibling-directory convention used everywhere else (global `CLAUDE.md`, the original `AGENTS.md`). Is this change intentional? If so, it would be worth aligning the global `~/.claude/CLAUDE.md` as a follow-up, or noting the discrepancy somewhere so agents don't get conflicting instructions.

**[Minor question] Build section dropped inline explanatory comments**

`DEVELOPMENT.md` — `## Build And Run` section lists bare `make` commands with no inline hints. The original `AGENTS.md` had comments like `# npm install + go mod download (first time only)`, `# Vite dev server on :5173`, and `# backend must be running separately` that are useful first-time orientation. Was the intent to keep the section lean and let `make help` (or the prose in other docs) carry that context? If so, a brief note pointing to `make help` or the architecture doc would help new contributors land the right mental model.

---

### Overall

The structural split is sound and the `Editing Rules` section is a good safeguard against future drift. The two open questions — whether `CLAUDE.md` needs a `@DEVELOPMENT.md` include, and whether the `/tmp` worktree path is intentional — should be confirmed before merge, since both affect how agents and developers pick up the project rules.

tomo-chan:
Follow-up for the review questions:

- `CLAUDE.md` now loads both `@DEVELOPMENT.md` and `@docs/security.md` in addition to `@AGENTS.md`, so the workflow rules and security requirements are part of the default project context instead of relying on agents to follow links manually.
- `DEVELOPMENT.md` now explains that `/tmp/<repo>-<feature>` is intentional and matches the repository's agent workflow, which keeps implementation out of the main working directory.
- The build/dev sections in `DEVELOPMENT.md` now include the inline orientation comments again for first-time readers.

Addressed in commit:
https://github.com/tomo-chan/panemux/commit/ec1db62240f775d1ea440f7d5915c6ac1f00c271

tomo-chan (commented):
## Code Review (follow-up)

All three questions from the previous review are resolved. Notes on the updated diff:

### Previous questions — resolved

**CLAUDE.md auto-loading** — resolved cleanly. `CLAUDE.md` now explicitly `@`-loads both `DEVELOPMENT.md` and `docs/security.md` in addition to `AGENTS.md`. Workflow rules and security requirements are in context from the first message without relying on agents to follow links.

**Worktree path** — resolved. `DEVELOPMENT.md` now explains the intent: `/tmp/<repo>-<feature>` is deliberate to avoid accidental edits in the main working directory. The rationale is visible in the document, which is all that was needed.

**Build section comments** — resolved. Inline orientation comments are back (`# first-time setup`, `# full build; frontend/dist is embedded`, dev server descriptions).

---

### What's Good (new in this revision)

- **`docs/security.md` as a dedicated `@`-loaded file** is a better solution than loading all of `docs/architecture.md`. Security requirements (CodeQL rules, `validRemotePath`, `gosec` policy, `known_hosts` exception) now land in context unconditionally without pulling in architecture narrative that's only relevant when understanding the design. The document scope split — architecture.md for rationale, security.md for implementation requirements — is well-drawn.
- **`docs/architecture.md` Security Design section** is now a clean four-bullet summary plus pointer. Future readers get context; implementers get redirected to the actionable document. Good hierarchy.
- **AGENTS.md item 4** is specific and actionable: "Read `docs/security.md` *before implementing changes that affect* command execution, shell argument handling..." — the conditional phrasing is correct and matches when the document is needed.

---

### Overall

All prior concerns are addressed and the new structure is cleaner than the first revision. Ready to merge.

PR #102, opened 2026-05-19
fix: restore git header for ssh and ssh tmux panes

tomo-chan:
## Summary

- fix git header resolution for ssh and ssh_tmux panes by inspecting remote git state over SSH
- keep active worktree preference for remote agent sessions and support PR lookup from remote origin metadata
- add backend tests for remote git context resolution and document the behavior

## Test Plan

- [x] go test ./internal/api ./internal/session
- [x] make test-go
- [x] make lint-go
- [x] Manual reproduction against `server` with a fresh panemux config: confirm `/api/sessions/ssh-repro/git-info` and `/api/sessions/ssh-tmux-repro/git-info` return `{"branch":"main","repo":"panemux","is_git":true}`

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -162,6 +162,7 @@ Why this hook:
 
 - first, it asks the session for its live working directory
 - second, it may ask for an active interactive agent workdir override
+- local and local tmux sessions inspect Git metadata on the local filesystem; SSH and SSH+tmux sessions inspect Git metadata on the remote host over the existing SSH connection so remote-only repositories still render header Git info
 
 The override path is intentionally narrow:
 
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -145,6 +145,10 @@ For pane-header Git status, local and local tmux panes inspect the local filesys
 and `ssh_tmux` panes run the equivalent Git inspection on the remote host. This allows headers to
 show branch/repository info even when the repository exists only on the SSH target.
 
+Remote Git inspection currently depends on `git rev-parse --path-format=absolute --git-common-dir`
+on the SSH target, which requires Git 2.31 or newer. Older remote Git versions degrade to "no git
+info" in the pane header for SSH-backed panes.
+
 ## REST API
 
 ### `GET /api/layout`
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -141,6 +141,10 @@ session and replays that snapshot when a pane reconnects after a workspace switc
 but prompt notifications no longer depend on the pane being visibly mounted at the time the output
 arrives.
 
+For pane-header Git status, local and local tmux panes inspect the local filesystem, while `ssh`
+and `ssh_tmux` panes run the equivalent Git inspection on the remote host. This allows headers to
+show branch/repository info even when the repository exists only on the SSH target.
+
 ## REST API
 
 ### `GET /api/layout`
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/fs"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -712,13 +713,6 @@ type gitInfoResponse struct {
 	IsGit    bool   `json:"is_git"`
 }
 
-type gitContext struct {
-	branch    string
-	commonDir string
-	repo      string
-	root      string
-}
-
 // GetGitInfo returns git repository information for the session's current working directory.
 func (h *Handler) GetGitInfo(w http.ResponseWriter, r *http.Request) {
 	id := chi.URLParam(r, "id")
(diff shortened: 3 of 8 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Minimal optional interface**: `GitContextGetter` is small and opt-in — sessions that don't implement it fall through to local filesystem inspection without any changes. This is the right pattern for capability extension.
- **Shared `remoteGitContext` helper**: Both `SSHSession` and `TmuxSSHSession` delegate to the same function through `sshSessionRunner`, so there is no logic duplication between the two SSH-backed session types.
- **Security path is consistent**: `validateRemotePath` is applied before constructing the remote shell command in `remoteGitContext`, matching the guard pattern already established for `cwd` and `shell` in other SSH paths.
- **`|| true` idiom in the template**: Detached HEAD (empty branch) and repos with no `origin` remote don't abort the whole command — the variables are just left empty and parsed as empty strings, which is correctly handled downstream.
- **Backward compatibility preserved**: `inspectLocalGitContext` doesn't populate `OriginURL`, so `lookupPRInfo` for local sessions is unchanged — `cmd.Dir = cwd` still applies.
- **Same-PR documentation**: `docs/behavior.md` and `docs/architecture.md` are updated alongside the code.
- **Good test layering**: `ssh_test.go` tests `remoteGitContext` directly via `fakeSSHRunner`; `handler_test.go` tests the full handler path via `mockRemoteGitSession`, including the active-worktree branch preference case.

---

### Issues & Questions

**[Question] `TestRepoSpecFromOriginURL` is not table-driven**

`DEVELOPMENT.md` says "Use table-driven tests when factors form a matrix." Four cases with different input shapes and expected outputs is exactly that. The current test also leaves some edge cases untested:

- SSH URL with an explicit port (`ssh://git@host:2222/owner/repo.git`) parses as `u.Host = "host:2222"`, producing `"host:2222/owner/repo"`. Does `gh pr view --repo host:2222/owner/repo` accept that format?
- URL with no `.git` suffix (valid for many hosts) — `TrimSuffix` is a no-op, so parsing should still work, but there is no test asserting this.
- HTTPS URL with embedded credentials (`https://user:token@github.com/owner/repo.git`) — `url.Parse` strips `User`, so the result would be `"github.com/owner/repo"`. Worth confirming this is intentional.

Should the test be rewritten as a `t.Run`-based table and the missing cases added?

---

**[Question] SSH sessions with no `origin` remote fall through to `cmd.Dir = remoteDir`**

In `lookupPRInfo`:

```go
if repoSpec := repoSpecFromOriginURL(gitCtx.OriginURL); repoSpec \!= "" {
    cmd.Args = append(cmd.Args, "--repo", repoSpec)
} else {
    cmd.Dir = cwd  // cwd may be a remote path for SSH sessions
}
```

When the remote repo has no `origin` configured, `OriginURL` is `""`, `repoSpecFromOriginURL` returns `""`, and the `else` branch sets `cmd.Dir` to a remote path (e.g., `/home/demo/panemux`) that typically does not exist on the local machine. The `gh` invocation then fails, and PR lookup returns `"", 0`.

The degradation is graceful, but the intent is implicit. Is the design "PR lookup is silently skipped for SSH repos with no origin" (in which case an early return on `OriginURL == ""` after a remote-session check would be clearer), or is the silent fallthrough intentional to keep the `lookupPRInfo` signature session-type–agnostic? If the latter, a short comment on the `else` branch would help future readers understand why setting `cmd.Dir` to a potentially nonexistent path is acceptable.

---

**[Minor question] `--path-format=absolute` requires Git ≥ 2.31**

`sshGitContextCmdTemplate` uses:

```sh
common=$(git rev-parse --path-format=absolute --git-common-dir)
```

`--path-format` was added in Git 2.31.0 (March 2021). On older remote hosts (e.g., RHEL 8 ships Git 2.27, Ubuntu 20.04 LTS ships Git 2.25), this flag is unrecognised and the command fails, causing `remoteGitContext` to return an error. The session header would show no git info rather than showing a potentially wrong relative path.

This is a known trade-off (absolute paths are required for correct worktree `CommonDir` comparison), but is there a minimum Git version documented anywhere, or should `docs/behavior.md` note this requirement?

---

### Overall

The design is clean and the implementation is consistent with the existing SSH session patterns. The two questions above are the main things worth confirming before merge — particularly whether the no-`origin` fallthrough behavior and the `--path-format` Git version floor are intentional and documented enough.

tomo-chan:
Follow-up for the review summary:

- Addressed in https://github.com/tomo-chan/panemux/commit/f7817c80c7c8117497a1d1b9f118c845be8040cd
- Rewrote `TestRepoSpecFromOriginURL` as a table-driven test and added the missing cases for explicit SSH ports, missing `.git`, and embedded credentials
- Tightened `repoSpecFromOriginURL` to use `url.Hostname()` so explicit SSH ports do not leak into `gh --repo`
- Made the remote-session no-`origin` behavior explicit in `lookupPRInfo`: SSH-backed sessions now return no PR metadata immediately instead of falling through to a local `cmd.Dir` on a remote-only path
- Documented the remote Git 2.31 requirement in `docs/behavior.md`

Local verification rerun:

- `go test ./internal/api ./internal/session`
- `make lint-go`
- `make test-go`

tomo-chan (commented):
## Code Review (follow-up on f7817c8d)

All three concerns from the previous review are resolved. Notes on each:

**[Question] `TestRepoSpecFromOriginURL` not table-driven → resolved**

The test is now a proper `t.Run`-based table with seven cases. The `credentialOrigin` split-string construction (`"https://user:" + "placeholder" + ...`) is a nice touch to avoid tripping naive secret-scanner patterns in test files.

The explicit-port case (`ssh://git@git.example.com:2222/team/panemux.git`) now produces `"git.example.com/team/panemux"` thanks to switching `u.Host` to `u.Hostname()`. That change also prompted a small but correct defensive addition: `u.Host == ""` is still checked before `u.Hostname()` is called, which is redundant (an empty `Host` implies an empty `Hostname`) but not harmful.

**[Question] SSH no-origin fallthrough → resolved**

`lookupPRInfo` now takes `sess session.Session` and returns `"", 0` immediately for `GitContextGetter` sessions without an origin-derived repo spec. The comment explains the intent for future readers. `TestLookupPRInfo_RemoteSessionWithoutOriginSkipsLookup` pins this behavior: the fake `gh` binary exits 99, so a test failure would catch any regression that lets the subprocess run when it should not.

**[Minor question] Git 2.31 requirement → resolved**

`docs/behavior.md` now documents the `--path-format=absolute` floor and the "no git info" degradation for older remote Git versions.

---

### Overall

The follow-up commit is clean and the three changes all make sense together. No new concerns. Good to merge.

PR #104, opened 2026-05-20
fix: detect codex root process in tmux worktrees

tomo-chan:
## Summary
- treat the active tmux pane PID itself as an eligible interactive agent when it is `codex` or `claude`
- restore worktree-based git branch detection for `tmux` and `ssh_tmux` panes when Codex is the pane root process
- add regression tests for local tmux and ssh+tmux root-agent cases

## Test plan
- [x] `make build-frontend`
- [x] `make lint-go`
- [x] `make test-go`

### DIFF
--- a/internal/session/local.go
+++ b/internal/session/local.go
@@ -247,6 +247,10 @@ func parsePSOutput(out []byte) ([]processInfo, error) {
 }
 
 func newestInteractiveAgentDescendantPID(processes []processInfo, rootPID int) (int, bool) {
+	if proc, ok := processByPID(processes, rootPID); ok && isInteractiveAgentCommand(proc.Command) {
+		return rootPID, true
+	}
+
 	children := childProcessMap(processes)
 	stack := append([]processInfo(nil), children[rootPID]...)
 	var matched int
--- a/internal/session/local.go
+++ b/internal/session/local.go
@@ -247,14 +247,15 @@ func parsePSOutput(out []byte) ([]processInfo, error) {
 }
 
 func newestInteractiveAgentDescendantPID(processes []processInfo, rootPID int) (int, bool) {
-	if proc, ok := processByPID(processes, rootPID); ok && isInteractiveAgentCommand(proc.Command) {
-		return rootPID, true
-	}
-
 	children := childProcessMap(processes)
 	stack := append([]processInfo(nil), children[rootPID]...)
-	var matched int
-	var ok bool
+	matched := 0
+	ok := false
+
+	if proc, found := processByPID(processes, rootPID); found && isInteractiveAgentCommand(proc.Command) {
+		matched = rootPID
+		ok = true
+	}
 
 	for len(stack) > 0 {
 		proc := stack[len(stack)-1]
--- a/internal/session/local_test.go
+++ b/internal/session/local_test.go
@@ -193,6 +193,19 @@ func TestNewestInteractiveAgentDescendantPID_IgnoresNonInteractiveAgents(t *test
 	assert.Equal(t, 140, pid)
 }
 
+func TestNewestInteractiveAgentDescendantPID_PrefersNewerChildOverInteractiveRoot(t *testing.T) {
+	processes := []processInfo{
+		{PID: 100, PPID: 1, Command: "codex resume --last"},
+		{PID: 140, PPID: 100, Command: "claude"},
+		{PID: 150, PPID: 100, Command: "codex exec"},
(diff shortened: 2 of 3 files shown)


internal/session/local.go
```diff
@@ -247,6 +247,10 @@ func parsePSOutput(out []byte) ([]processInfo, error) {
 }
 
 func newestInteractiveAgentDescendantPID(processes []processInfo, rootPID int) (int, bool) {
+	if proc, ok := processByPID(processes, rootPID); ok && isInteractiveAgentCommand(proc.Command) {
+		return rootPID, true
```
tomo-chan: If a `codex`-root pane ever spawns a second `codex` child (e.g. a parallel sub-agent), this short-circuit returns the older root rather than the newer child that the pre-change `newestInteractiveAgentDescendantPID` logic would have preferred. Worth a note here (or a dedicated test) if that case is ruled out by design.

tomo-chan (commented):
## Code Review

### What's Good

- **Minimal footprint.** The production change is four lines; the tests do the heavy lifting and prove both the local-tmux and SSH-tmux paths end-to-end.
- **Correct `isInteractiveAgentCommand` coverage.** `codex resume --last` passes the check because `resume` is not `exec`, so the fix works for the named worktree use case.
- **Implicit efficiency win.** The early return avoids building the child-process map entirely when the pane root is already an agent — a minor but free improvement.
- **Symmetric test coverage.** Both `local_test.go` and `ssh_test.go` exercise the new path. The SSH test reuses the `fakeSSHRunner` pattern consistently with the existing suite.

---

### Issues & Questions

**[Question] Priority when root-agent has interactive-agent children**

```go
func newestInteractiveAgentDescendantPID(processes []processInfo, rootPID int) (int, bool) {
    if proc, ok := processByPID(processes, rootPID); ok && isInteractiveAgentCommand(proc.Command) {
        return rootPID, true   // ← short-circuits before scanning children
    }

    children := childProcessMap(processes)
    // picks highest PID among descendants …
```

Before this change the function returned the **highest-PID** interactive-agent process in the full descendant tree, which is used as a proxy for "most recently created / active". After this change, if the pane root is itself an agent it is returned immediately without checking descendants.

Is it intentional that a root-agent takes absolute priority over any newer interactive-agent children? The scenario would be: PID 100 = `codex resume --last` (pane root), and PID 500 = a second `codex` spawned by the first (e.g. as a parallel sub-agent). Old behaviour: returns 500. New behaviour: returns 100.

If nested interactive agents are impossible in practice for this session topology, a short comment to that effect (or a test name that makes the assumption visible) would close the question. If they are possible, including rootPID in the candidate set rather than short-circuiting may be safer.

---

### Overall

Clean, well-scoped fix with symmetric test coverage. The one open question is whether the short-circuit is intentional when a root-agent has interactive-agent descendants — happy to approve once that's confirmed (or noted as an accepted edge case).

internal/session/local.go
tomo-chan: Adjusted this in https://github.com/tomo-chan/panemux/commit/86fb264c4df6db40ff2e88597e9d3aecb75040c7.

`newestInteractiveAgentDescendantPID` now keeps an interactive `rootPID` in the candidate set, but still scans descendants and returns the highest-PID interactive agent overall. That preserves the original "newest interactive agent wins" behavior while fixing the `pane_pid == codex` case.

I also added `TestNewestInteractiveAgentDescendantPID_PrefersNewerChildOverInteractiveRoot` to lock the nested-agent case down.

tomo-chan (commented):
## Code Review (follow-up)

### Previous concern: resolved

**[Question] Priority when root-agent has interactive-agent children** — resolved in 86fb264e.

The revised implementation correctly addresses the concern. Rather than short-circuiting on the root, the new code seeds the candidate with `rootPID` and then continues scanning descendants, keeping the "highest PID wins" invariant intact:

```go
matched := 0
ok := false

if proc, found := processByPID(processes, rootPID); found && isInteractiveAgentCommand(proc.Command) {
    matched = rootPID
    ok = true
}

for len(stack) > 0 {
    // ... proc.PID > matched check still applies
}
```

`TestNewestInteractiveAgentDescendantPID_PrefersNewerChildOverInteractiveRoot` locks the nested-agent case down: root PID 100 = `codex`, child PID 160 = `codex` → returns 160. That's exactly the right safety net.

---

### Overall

The implementation is correct and the test coverage is thorough. No remaining blockers.

PR #106, opened 2026-05-20
fix: detect ssh tmux git-info worktree correctly

tomo-chan:
## Summary
- fix `ssh_tmux` git-info resolution so worktree-backed branch information is detected correctly
- avoid repeated `ssh: Stdout already set` failures by using fresh SSH exec sessions for remote workdir probes
- keep the temporary debug logger out of the final change and retain richer normal logs for tracing remote workdir resolution

## Test Plan
- [x] `go test ./internal/session ./internal/api`
- [x] `golangci-lint run ./internal/session/... ./internal/api/...`


tomo-chan (commented):
## Code Review

### What's Good

- **Session factory fix is correct.** The `ssh: Stdout already set` failure happens because `ssh.Session.Output()` sets `Stdout` internally and panics if called twice on the same session. Routing every remote command through a fresh session from `newRunner()` via `outputFromSessionFactory` is exactly the right structural fix, not a workaround.
- **Single-call tmux pane info.** Combining `#{pane_pid}\t#{pane_current_path}` into one `display-message` invocation removes one SSH round-trip per git-info request for `ssh_tmux` sessions — a clean, net-positive optimization.
- **`remoteOutputFunc` type.** Abstracting `func(string) ([]byte, error)` makes the session lifecycle independent from command execution and keeps test fakes simple.
- **Structured log format.** All new `log.Printf` calls use key=value pairs (`root_pid=`, `base_cwd=`, `session=`), making them grep-friendly in production logs.
- **Factory tests.** `TestActiveRemoteWorkdirFromSessionFactory_UsesSeparateRunners` and `TestTmuxSSHActiveWorkdirFromSessionFactory_UsesSeparateRunners` directly exercise the session-separation invariant that the fix is meant to enforce.

---

### Issues & Questions

**[Question] "Env-gated" description doesn't match the implementation**

The PR description says *"add env-gated git info debug logging"* but all `log.Printf` calls in the diff are unconditional — no env variable check or conditional `log.SetOutput` is visible. `resolvePreferredCWD` is called on every git-info request (handler.go lines 546 and 743), so always-on debug messages fire at that cadence in production.

Is there env gating elsewhere not shown in the diff (e.g., `log.SetOutput(io.Discard)` in `main.go` unless a flag is set)? If not, should the description be corrected, and/or should the log calls be guarded (e.g., `if os.Getenv("PANEMUX_DEBUG") \!= ""`)?

**[Minor question] Duplicated pane-info parsing in `tmuxSSHActiveWorkdir` vs `tmuxSSHActiveWorkdirFromSessionFactory`**

Both functions contain identical `SplitN` + `Atoi` + log-on-error logic for the combined `#{pane_pid}\t#{pane_current_path}` output:

```go
fields := strings.SplitN(strings.TrimSpace(string(out)), "\t", 2)
if len(fields) \!= 2 { ... }
panePID, err := strconv.Atoi(strings.TrimSpace(fields[0]))
...
baseCWD := strings.TrimSpace(fields[1])
```

A small helper `parsePaneInfo(out []byte) (panePID int, baseCWD string, err error)` would remove the duplication. Happy to file a follow-up issue if you'd prefer to keep this PR focused.

**[Minor question] Magic number `4` in session-factory tests**

`TestActiveRemoteWorkdirFromSessionFactory_UsesSeparateRunners` asserts `assert.Equal(t, 4, index)` and `TestTmuxSSHActiveWorkdirFromSessionFactory_UsesSeparateRunners` asserts `4` as well. The expected count is implicitly coupled to the number of remote probes. A short comment listing the breakdown (e.g., `// 1: root PID probe, 2: list processes, 3: open files, 4: PID CWD`) would help future maintainers diagnose an unexpected count rather than guessing why the number changed.

---

### Overall

The session-separation fix and tmux pane-info consolidation are solid. The main open question is whether the logging is intentionally unconditional or if the env-gate mentioned in the description is missing; that's worth clarifying before merge.

tomo-chan:
Updated this PR to match the current implementation.

- The env-gated debug logger was removed in [42dc7fe](https://github.com/tomo-chan/panemux/commit/42dc7fe64b4e01f398627cbb73c8bb27bdd03c64).
- `ssh` / `ssh_tmux` now emit richer normal logs around pane lookup, process discovery, Codex session-log detection, and fallback workdir selection.
- The PR title/body now describe the shipped behavior instead of the earlier debug-build direction.

I left the duplicated tmux pane parsing and the factory-test magic number out of this PR to keep the fix focused on the SSH/session handling and logging behavior.

tomo-chan (commented):
## Code Review (follow-up)

All three concerns from the previous review have been addressed via the author's reply and the updated PR description. Summarising the resolution:

**[Question] "Env-gated" description resolved**
The env-gated debug logger was an earlier direction that was removed in `42dc7fe`. The current logging is intentionally always-on operational tracing, and the PR title/body now accurately describe the shipped behaviour. No further action needed.

**[Minor question] Duplicated pane-info parsing — acknowledged, deferred**
Noted. Happy to file a follow-up issue for `parsePaneInfo` extraction when this PR lands.

**[Minor question] Magic number `4` in factory tests — acknowledged, deferred**
Noted. A comment or named constant can go in a separate cleanup PR.

---

### Overall

All previous concerns are resolved or deliberately deferred with a clear rationale. The session-factory fix and tmux pane-info consolidation are correct and well-tested. Ready to merge.

tomo-chan:
The two maintainability points from review are now fixed in this PR.

- [c674067](https://github.com/tomo-chan/panemux/commit/c674067) extracts the duplicated tmux pane parsing into `parseRemoteTmuxPaneInfo(...)`.
- The same commit also documents the expected `4` session-factory calls in both tests so the count is no longer a magic number.

Local verification:
- `golangci-lint run ./internal/session/... ./internal/api/...`

tomo-chan (commented):
## Code Review (round 3)

All concerns from previous rounds have been addressed. Confirming resolutions at the new HEAD (`929b525d`):

**Duplicated pane-info parsing — fixed**
`c674067b` extracts `parseRemoteTmuxPaneInfo(out []byte) (int, string, error)` in `tmux_ssh.go`. Both `tmuxSSHActiveWorkdir` and `tmuxSSHActiveWorkdirFromSessionFactory` now call it, and the error it returns (which includes the raw output) is forwarded directly to the log: `log.Printf("%s %v", logScope, err)`. Clean.

**Magic number `4` — documented**
Both factory tests now carry an explanatory comment listing the four probe steps. Future maintainers will know immediately why the count changed if it ever does.

**`debugScope` → `logScope` rename (`929b525d`)**
Consistent with actual usage. No functional change.

---

### Overall

All three previous concerns are resolved. The session-factory fix, tmux pane-info consolidation, and structured logging are correct and well-tested. Good to merge.

PR #108, opened 2026-05-30
feat: link pane header repo name to repository page

tomo-chan:
## Summary
- add `repo_url` to pane git info so the UI can open the repository page from header chrome
- resolve browser-safe repository URLs from local and remote git origins
- render the pane header repository name as an inline link and cover the new behavior in tests and docs

## Test Plan
- [x] make check

### DIFF
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -473,6 +473,7 @@ When a pane moves to a different parent node, the component may be remounted by 
 - If the agent exits, no longer has an eligible worktree, or the resolved worktree is not a sibling worktree of the pane's current repository, panemux falls back to the pane's own working directory immediately.
 - For SSH panes, panemux resolves interactive agent processes from the remote process list on the current SSH connection.
 - For SSH+tmux panes, panemux resolves interactive agent processes only from the currently active remote tmux pane.
+- If the resolved repository origin can be converted into a browser URL, the header shows the repository name as a link to that repository page.
 - If the resolved branch has a GitHub pull request, the header shows a PR link labeled `#<number>`.
 
 ## Operational Assumptions
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -474,7 +474,7 @@ When a pane moves to a different parent node, the component may be remounted by 
 - For SSH panes, panemux resolves interactive agent processes from the remote process list on the current SSH connection.
 - For SSH+tmux panes, panemux resolves interactive agent processes only from the currently active remote tmux pane.
 - If the resolved repository origin can be converted into a browser URL, the header shows the repository name as a link to that repository page.
-- For SCP-style SSH origins such as `git@alias:owner/repo.git`, panemux uses `~/.ssh/config` `Host` aliases to resolve the browser link hostname when a matching alias exists; otherwise it treats the SSH host token as the hostname directly.
+- For SCP-style SSH origins such as `git@alias:owner/repo.git`, panemux uses `~/.ssh/config` `Host` aliases to resolve both the browser link hostname and GitHub PR lookup repo host when a matching alias exists; otherwise it treats the SSH host token as the hostname directly.
 - If the resolved branch has a GitHub pull request, the header shows a PR link labeled `#<number>`.
 
 ## Operational Assumptions
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -474,6 +474,7 @@ When a pane moves to a different parent node, the component may be remounted by 
 - For SSH panes, panemux resolves interactive agent processes from the remote process list on the current SSH connection.
 - For SSH+tmux panes, panemux resolves interactive agent processes only from the currently active remote tmux pane.
 - If the resolved repository origin can be converted into a browser URL, the header shows the repository name as a link to that repository page.
+- For SCP-style SSH origins such as `git@alias:owner/repo.git`, panemux uses `~/.ssh/config` `Host` aliases to resolve the browser link hostname when a matching alias exists; otherwise it treats the SSH host token as the hostname directly.
 - If the resolved branch has a GitHub pull request, the header shows a PR link labeled `#<number>`.
 
 ## Operational Assumptions
--- a/docs/ui-design.md
+++ b/docs/ui-design.md
@@ -85,6 +85,8 @@ From left to right, the header contains:
 
 The header background stays `#252526` with a `#333` bottom border so the new controls inherit the established terminal chrome instead of introducing a second visual language.
 
+When panemux can derive a repository page URL from the pane's Git origin, the repository name is rendered as an inline text link using the same visual treatment as the PR shortcut.
+
 The PR shortcut is rendered as an inline text link labeled `#<number>`. It does not use a pill, badge, or outlined capsule treatment. On hover, the link shifts to a slightly stronger blue and shows an underline so it reads like the rest of the header chrome instead of a separate button system.
 
 ### Split vs quick-add
--- a/frontend/src/components/PaneHeader.test.tsx
+++ b/frontend/src/components/PaneHeader.test.tsx
@@ -107,6 +107,7 @@ describe('PaneHeader VSCode button', () => {
         gitInfo={{
           is_git: true,
           repo: 'panemux',
+          repo_url: 'https://github.com/example/panemux',
           branch: 'feature/pane-pr-link',
           pr_number: 123,
           pr_url: 'https://github.com/example/panemux/pull/123',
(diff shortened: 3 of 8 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Clean refactoring of the URL helper.** `repoSpecFromOriginURL` is now a thin wrapper around the shared `repoHostAndPathFromOriginURL`, eliminating duplication without changing any observable behavior. The existing `TestRepoSpecFromOriginURL` test suite continues to verify the refactored path.
- **Credential stripping is correct and tested.** HTTPS origins with embedded credentials (`https://user:token@github.com/…`) produce a clean URL via `u.Hostname()`, and the new `TestRepoPageURLFromOriginURL` table has an explicit case for this.
- **Security-conscious frontend.** The new repo link carries `rel="noopener noreferrer"` and `target="_blank"`, matching the existing PR link's treatment.
- **Schema-first.** `repo_url: z.string().url().optional()` lands in `GitInfoSchema` before any component change touches the field.
- **Test coverage mirrors the PR-link pattern.** Three frontend cases — link present, link absent, hover style — are a direct parallel to the PR link tests, making the intention clear.
- **Docs updated in the same change.** Both `behavior.md` and `ui-design.md` are updated in the same commit, consistent with the project's documentation rule.

---

### Issues & Questions

**[Minor question] `TestRepoPageURLFromOriginURL` is missing the explicit-port SSH URL variant**

`TestRepoSpecFromOriginURL` includes a case for `ssh://git@git.example.com:2222/team/panemux.git` to document that the port is stripped from the resulting spec. `TestRepoPageURLFromOriginURL` has the general "ssh url" case but not the explicit-port variant. Since the port-stripping comes from `u.Hostname()` inside the shared helper, the runtime behavior is correct; the gap is only that the guarantee isn't also documented at the `repoPageURLFromOriginURL` call-site level.

Is the omission intentional (trusting the shared helper is sufficient), or worth adding for symmetry with `TestRepoSpecFromOriginURL`?

**[Minor question] SSH config host alias produces a non-working link without signal**

If a user's git remote uses an SSH config alias — e.g. `git@github-work:org/repo.git` where `github-work` resolves via `~/.ssh/config` — `repoPageURLFromOriginURL` returns `https://github-work/org/repo`. The header renders the repo name as a link, but clicking it goes nowhere.

The current design always emits a URL for any parseable origin, so broken links are possible. Is this acceptable, or should there be a lightweight heuristic (e.g. require at least one `.` in the hostname) to suppress the link when the host doesn't look like a public FQDN? Worst case today is a non-functional link rather than a security issue, so this is a nice-to-have rather than a blocker.

---

### Overall

The implementation is clean, well-tested, and consistent with the project's schema-first and security conventions. The two minor questions above are not blockers; the PR is ready to merge as-is.

tomo-chan:
Addressed the review follow-up in commit https://github.com/tomo-chan/panemux/commit/3e2de2294888e5148baa87826310e817943e53fe.

- Added the missing explicit-port SSH URL case to `TestRepoPageURLFromOriginURL` for symmetry with `TestRepoSpecFromOriginURL`.
- Re-ran `go test ./internal/api/...` locally.

On the SSH config alias question: I left the current behavior as-is for this PR. The header link is only based on the parsed Git origin today, and adding heuristics like "hostname must contain a dot" would start suppressing legitimate internal hosts as well. Since the current outcome is at worst a non-working link and not a correctness or security regression, I treated that as a non-blocking follow-up rather than changing the behavior in this PR.

tomo-chan (commented):
## Code Review (follow-up)

### Previous concerns

- **Resolved — explicit-port SSH URL test case.** Commit `3e2de229` adds the `"ssh url with explicit port"` case to `TestRepoPageURLFromOriginURL`, making the test matrix symmetric with `TestRepoSpecFromOriginURL`. Thanks for the quick turnaround.
- **Open (non-blocker) — SSH config host alias.** The earlier note about SSH config aliases producing a non-working link (e.g. `git@github-work:org/repo.git` → `https://github-work/org/repo`) remains. No reply yet, but it was flagged as a non-blocker and requires no action before merging — tracking it here for visibility.

---

### Overall

Both commits are clean; no new concerns. The PR is ready to merge.

tomo-chan:
Added SSH config alias resolution for pane-header repo links in commit https://github.com/tomo-chan/panemux/commit/3eca6a6c2df3d99b0ebe815289858ab3eba24008.

- SCP-style SSH origins such as `git@alias:owner/repo.git` now check the loaded `~/.ssh/config` for a matching `Host` entry and use its `HostName` when building the browser URL.
- If there is no matching SSH config alias, panemux keeps treating the SCP host token as the hostname directly.
- Re-ran `make check` locally.

tomo-chan (commented):
## Code Review (follow-up 2)

### Previous concerns

- **Resolved — SSH config host alias.** Commit `3eca6a6c` adds `(h *Handler).resolveSSHConfigHostAlias` that reads `~/.ssh/config` via the existing `sshconfig.ParseHosts` path and substitutes the real `HostName` when a match is found. Thank you for the thorough implementation.

---

### What's Good (new in `3eca6a6c`)

- **`scpStyle` bool keeps alias resolution scoped correctly.** `repoHostAndPathFromOriginURLDetailed` returns `true` only for `git@` SCP-style origins. URL-style `ssh://` origins already carry a real hostname in `u.Hostname()`, so alias lookup is correctly skipped for them.
- **Graceful degradation.** If `ParseHosts` fails, `resolveSSHConfigHostAlias` logs the error and returns the original host unchanged. The feature degrades to the pre-alias-resolution behavior rather than producing an empty link.
- **`sshConfigPath: sshconfig.DefaultPath()` in `NewHandler`.** This is the same injectable default used for the rest of the SSH config operations, so alias resolution works in production without any user configuration.
- **Both alias-found and alias-not-found cases are tested.** `TestHandlerRepoPageURLFromOriginURL_ResolvesSSHConfigAlias` and `TestHandlerRepoPageURLFromOriginURL_LeavesUnknownSCPHostUntouched` together document the intended contract.
- **`behavior.md` updated** to describe the `~/.ssh/config` alias resolution step.

---

### Issues & Questions

**[Minor question] `lookupPRInfo` still uses unresolved `repoSpecFromOriginURL` for SSH alias origins**

After this PR, the handler resolves SSH aliases when building `repo_url`, but `lookupPRInfo` calls the free function `repoSpecFromOriginURL` — which does not do alias resolution:

```go
// internal/api/handler.go (unchanged path)
if repoSpec := repoSpecFromOriginURL(gitCtx.OriginURL); repoSpec \!= "" {
```

For a user whose origin is `git@github-work:org/repo.git`, the repo link in the header will now correctly point to `https://github.com/org/repo`, but PR badge detection will still use `github-work/org/repo` as the repo spec and fail to find PRs.

This asymmetry is a pre-existing limitation of `lookupPRInfo` (not a regression of this PR), and the two features could reasonably be fixed in separate changes. That said, the commit that makes repo links work for alias users also makes the PR-badge gap more noticeable by comparison.

Is this gap tracked anywhere, or is alias resolution for `lookupPRInfo` a planned follow-up?

---

### Overall

The SSH alias resolution implementation is clean and consistent with the existing `sshconfig` usage in the handler. The one open item above is a pre-existing limitation rather than something introduced by this PR and does not block merge.

tomo-chan:
Addressed in commit https://github.com/tomo-chan/panemux/commit/f61831efa613c4c8870486a952a010465d443626.

- `lookupPRInfo` now resolves SCP-style SSH origin aliases through the loaded `~/.ssh/config`, so PR badge lookup uses the same resolved host as the repo link path.
- Added coverage for the resolved `--repo github.com/example/panemux` argument in the remote PR lookup test.
- Re-ran `make check` locally.

tomo-chan (commented):
## Code Review (follow-up 3)

### Previous concerns

- **Resolved — `lookupPRInfo` alias resolution.** Commit `f61831ef` replaces the call to `repoSpecFromOriginURL` (free function, no alias resolution) with `h.repoSpecFromOriginURL` (method, same alias resolution path as `h.repoPageURLFromOriginURL`). The asymmetry noted in the previous review is fully closed.

---

### What's Good (new in `f61831ef`)

- **Symmetric method pair.** `h.repoSpecFromOriginURL` and `h.repoPageURLFromOriginURL` now share the same alias-resolution logic through `resolveSSHConfigHostAlias`, so the two features stay in sync automatically.
- **Integration test pins the resolved `--repo` argument.** `TestLookupPRInfo_RemoteSessionWithSSHConfigAliasOrigin_UsesResolvedRepoSpec` uses a fake `gh` binary that asserts the argument is `github.com/example/panemux` (not `github-work/...`), making it clear exactly which code path the test exercises.
- **`TestHandlerRepoSpecFromOriginURL_ResolvesSSHConfigAlias` rounds out the unit coverage.** Together with the existing `TestRepoSpecFromOriginURL` table, both the alias-aware method and the pure free function are documented and tested.
- **`behavior.md` updated** to describe that alias resolution covers both the browser link and PR lookup, keeping the doc consistent with the implementation.

---

### Issues & Questions

None. All concerns from the earlier review rounds are resolved.

---

### Overall

All four commits are clean and consistent. No open questions remain; the PR is ready to merge.

PR #110, opened 2026-06-05
fix: keep agent worktree git context after exit

tomo-chan:
## Summary

- keep pane header Git/PR metadata pinned to the last valid sibling worktree after an interactive agent exits
- reuse the same sticky worktree resolution for the VS Code open action
- add local and remote regression tests for sticky worktree reuse and repo-change invalidation
- document the updated pane header and VS Code behavior

## Root Cause

Panemux only preferred a sibling worktree while an interactive `codex` or `claude` process was still detectable. Once the agent process ended, Git/PR lookup and the VS Code action immediately fell back to the pane's original working directory, which caused Claude worktree context to disappear after task completion.

## Impact

- Claude and Codex panes now keep showing the expected branch and linked PR after task completion when the last detected worktree still belongs to the same repository.
- SSH-backed panes get the same sticky behavior as local panes.
- The sticky override is dropped automatically when the pane moves to a different repository context.

## Test Plan

- [x] `go test ./internal/api`
- [x] `go test ./internal/session`
- [x] `go test ./internal/...`

## Notes

- `make test-go` still fails in this worktree because the top-level package embeds `frontend/dist`, which is not present yet.
- `make build-frontend` failed in this environment with `sh: tsc: command not found`, so I could not generate `frontend/dist` here before retrying the full gate.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,6 +18,7 @@ This installs npm packages and downloads Go modules. Required once per environme
 ## Read This First
 
 1. Read [DEVELOPMENT.md](DEVELOPMENT.md) for build, test, branch, and PR workflow rules.
+   This includes the path-sanitization rule for tests, fixtures, PR bodies, PR comments, and review replies.
 2. Read [docs/overview.md](docs/overview.md) for the product purpose and current boundaries.
 3. Read [docs/architecture.md](docs/architecture.md) for implementation structure and security design.
 4. Read [docs/security.md](docs/security.md) before implementing changes that affect command execution, shell argument handling, SSH path handling, or `gosec` posture.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -47,7 +47,7 @@ make check
 
 ### Path sanitization
 
-- Do not record real local or remote directory paths from a developer machine or environment in tests, fixtures, docs, screenshots, PR bodies, review replies, or commit history.
+- Do not record real local or remote directory paths from a developer machine or environment in tests, fixtures, docs, screenshots, PR bodies, PR comments, review replies, or commit history.
 - Replace environment-specific paths with fake placeholders such as `/workspace/user/project`, `/tmp/sample-project`, or `/remote/home/demo`.
 - If a real path is used temporarily for investigation, remove it before committing and rewrite the branch history if that path already landed in a published commit on the PR branch.
 
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ Common uses:
 - **Move a pane** by dragging the handle in its header; drop it on a workspace edge to create a new outer split, or on another pane edge / divider to insert it there.
 - **Close a pane** from its header controls; the layout collapses automatically.
 - **Restart a pane** with the on-screen button if the underlying session exits.
-- **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. For local panes, panemux can prefer the live worktree of an interactive `codex` or `claude` session, including resumed Codex sessions, and falls back to the pane's own working directory after the agent exits.
-- **Open VS Code** from supported panes using the pane header action. Like pane Git/PR metadata, this can prefer the live worktree of an interactive `codex` or `claude` session and falls back to the pane's own working directory after the agent exits.
+- **Open a linked PR** from the pane header when the current Git branch already has a GitHub pull request. For local panes, panemux can prefer the live worktree of an interactive `codex` or `claude` session, including resumed Codex sessions, and keeps the last valid sibling worktree pinned after the agent exits until a newer valid context is detected.
+- **Open VS Code** from supported panes using the pane header action. Like pane Git/PR metadata, this can prefer the live worktree of an interactive `codex` or `claude` session and keeps the last valid sibling worktree pinned after the agent exits until a newer valid context is detected.
 
 ### Notifications and attention prompts
 
@@ -124,7 +124,7 @@ Common uses:
 - `tmux` and `ssh_tmux` panes automatically create the target tmux session if it does not already exist.
 - Set `cwd` on `local`, `ssh`, or `ssh_tmux` panes when you want the shell to start in a specific directory.
 - In the pane settings dialog, `Working Directory` can be chosen from a browsable directory tree for both local and SSH-backed panes. Hidden directories are available through a toggle.
-- Pane headers resolve Git and PR metadata from the live working context, including interactive `codex` and `claude` worktrees for both local and SSH-backed panes. `tmux` and `ssh_tmux` use the currently active tmux pane only.
+- Pane headers resolve Git and PR metadata from the live working context, including interactive `codex` and `claude` worktrees for both local and SSH-backed panes. When a valid sibling worktree was detected recently, panemux keeps that worktree pinned after the agent exits until the pane moves to a different repository context. `tmux` and `ssh_tmux` use the currently active tmux pane only.
 
 ---
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -78,7 +78,7 @@ Workspace-related endpoints:
 - `PUT /api/workspaces/{id}/layout` updates a specific workspace layout.
 - `GET/PUT /api/layout` remain as compatibility endpoints for the active workspace layout.
 
-`POST /api/sessions/{id}/open-vscode` launches VSCode pointed at the session's live working directory. Like `GET /api/sessions/{id}/git-info`, it may prefer the worktree of an active interactive `codex` or `claude` process when that worktree belongs to the same repository. For local sessions it runs `code <cwd>`; for SSH sessions it runs `code --remote ssh-remote+<connection> <cwd>`. The binary is located via `exec.LookPath("code")` with a macOS app-bundle fallback.
+`POST /api/sessions/{id}/open-vscode` launches VSCode pointed at the session's live working directory. Like `GET /api/sessions/{id}/git-info`, it may prefer the worktree of an active interactive `codex` or `claude` process when that worktree belongs to the same repository, and it keeps using the last valid sibling worktree after the agent exits until the pane changes repository context. For local sessions it runs `code <cwd>`; for SSH sessions it runs `code --remote ssh-remote+<connection> <cwd>`. The binary is located via `exec.LookPath("code")` with a macOS app-bundle fallback.
 
 Why REST here:
 
@@ -170,6 +170,12 @@ The override path is intentionally narrow:
 - local and local tmux sessions only consider descendants of the pane's own shell process or active tmux pane process
 - SSH sessions scan the remote process list on the current SSH connection, and SSH+tmux sessions restrict that scan to the active remote tmux pane's process tree
 - only worktrees that belong to the same Git common dir as the pane's base repository are accepted
+
+When a valid override worktree is accepted, the API layer caches that sibling worktree per pane
+session. Later requests reuse it when no active agent workdir is currently detectable, but only
+while the pane still resolves to the same Git common dir and a different worktree root. This keeps
+pane headers and editor-opening behavior stable after an agent exits without allowing stale
+cross-repository reuse.
 
 For interactive Codex sessions, all four session implementations (`local`, `ssh`, `tmux`, and `ssh_tmux`) may inspect the open Codex session log under `~/.codex/sessions/...jsonl` when the Codex process has that file open. Panemux currently prefers the latest `exec_command.arguments.workdir` recorded in `response_item` / `function_call` log entries, then falls back to `turn_context.cwd`, then `session_meta.cwd`.
 
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -145,6 +145,12 @@ For pane-header Git status, local and local tmux panes inspect the local filesys
 and `ssh_tmux` panes run the equivalent Git inspection on the remote host. This allows headers to
 show branch/repository info even when the repository exists only on the SSH target.
 
+When panemux detects that an interactive `codex` or `claude` session is operating in a sibling Git
+worktree for the same repository, pane-header Git/PR metadata and the VS Code open action prefer
+that worktree. After the agent exits, panemux keeps the last valid sibling worktree pinned for that
+pane session until a newer valid sibling worktree is detected or the pane itself changes to a
+different repository.
+
 Remote Git inspection currently depends on `git rev-parse --path-format=absolute --git-common-dir`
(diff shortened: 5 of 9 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Clean three-method encapsulation**: `rememberPreferredCWD` / `recallPreferredCWD` / `clearPreferredCWD` keep all map access in one place and make the resolution logic in `resolvePreferredCWD` easy to read. The mutex is correctly applied to all three.
- **Precise invalidation rule**: `recallPreferredCWD`'s guard (`state.CommonDir \!= baseCtx.CommonDir || state.Root == baseCtx.Root`) is an exact expression of "same repo, different worktree root," and `TestResolvePreferredCWD_StickyWorktreeIgnoredAfterRepoChange` exercises the transition cleanly.
- **Four-scenario test coverage**: local VS Code sticky, local git-info sticky, remote git-info sticky, and repo-change invalidation cover the main behavioral axes. The `postOpenVSCodeOKWithRouter` helper refactor avoids duplicating the router setup neatly.
- **Docs updated in the same commit**: README, `docs/architecture.md`, and `docs/behavior.md` all reflect the new semantics — no documentation debt left behind.

---

### Issues & Questions

**[Question] Sticky state not cleared on `DeleteSession`, `DeleteWorkspace`, or `RestartSession`**

`DeleteSession` (handler.go ~line 362) and the pane loop in `DeleteWorkspace` (~line 230) call `h.manager.Remove(pane.ID)` but never `h.clearPreferredCWD(id)`. The entry in `preferredCWDBySession` outlives the session.

The more concrete problem is `RestartSession` (~line 390): it removes the old session then calls `createSession(found, …)`, which passes `pane.ID` through to `NewLocal`/`NewSSH`/etc. The new session receives **the same ID** as the old one and immediately inherits the stale sticky state. If the user restarted the pane because the agent worktree was cleaned up, the pane header will keep showing the old branch until an active agent is next detected.

Should `DeleteSession`, `DeleteWorkspace`, and `RestartSession` each call `h.clearPreferredCWD(id)` alongside the `manager.Remove` call?

**[Question] Stale sticky path is not self-healing**

If a pinned worktree is removed from disk between polls:

1. `resolvePreferredCWD` → `recallPreferredCWD` returns the deleted path
2. `inspectGitContextForSession(sess, targetCWD)` (~line 746) fails with "no such directory"
3. `GetGitInfo` returns `{IsGit: false}` and exits — the sticky state is **not touched**

On every subsequent poll the same stale path is recalled, the git lookup fails again, and the pane shows no git info indefinitely instead of falling back to the base CWD. The sticky state is only cleared on the next active-agent detection or a repo change.

Should `GetGitInfo` (or `resolvePreferredCWD` itself) call `clearPreferredCWD` when `inspectGitContextForSession` fails on the resolved path?

**[Minor question] Transient base-context failure now discards sticky state**

When `inspectGitContextForSession(sess, cwd)` fails at the top of `resolvePreferredCWD` (the pane's own CWD), `clearPreferredCWD` is now called before returning. If the failure is transient (e.g., a git index lock during a commit), a valid pinned worktree is permanently dropped.

The pre-PR behavior on that same error path just returned `cwd` without touching state. Was the `clearPreferredCWD` call here intentional, or should transient base-context failures leave the sticky state intact?

---

### Overall

The core sticky-worktree logic is well-structured and the test suite covers the key scenarios. The two questions above — sticky state surviving session deletion/restart, and the stale-path loop — are both observable in normal workflows and worth settling before merge.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/e6d1781b153aaae0fd6a4f93f56cbbe3885d9032

Changes made:
- clear sticky worktree state on `DeleteSession`, `DeleteWorkspace`, and `RestartSession`
- fall back to the pane base CWD when a remembered sticky worktree no longer has valid git context, and clear the stale state
- keep sticky state intact on transient base-context lookup failure instead of dropping it eagerly

Verification run locally:
- `make check`

tomo-chan (commented):
## Code Review — Follow-up (commits 0682345b, a90044f5, e6d17811)

### Previous Concerns — Resolved

All three concerns from the first review are addressed in commit e6d17811 and each has a matching test:

- **Sticky state on session deletion/restart** ✓ — `clearPreferredCWD` is now called in `DeleteSession`, `DeleteWorkspace` (per-pane), and `RestartSession` before `manager.Remove`.
- **Stale worktree path not self-healing** ✓ — `GetGitInfo` now detects when `inspectGitContextForSession` fails on the resolved sticky path, clears the state, and retries with the base CWD.
- **Transient base-context failure discarding sticky** ✓ — the eager `clearPreferredCWD` call on that path was removed; the sticky state is preserved across transient git failures.

---

### New Findings (Claude transcript detection, commit 0682345b)

**[Question] `latestClaudeTrackedPath` iterates a map with undefined order**

`internal/session/local.go`

```go
func latestClaudeTrackedPath(backups map[string]json.RawMessage) string {
    latest := 
    for path := range backups {
        if isClaudeAuxiliaryPath(path) {
            continue
        }
        latest = path
    }
    return latest
}
```

Go map iteration is not deterministic. If a `file-history-snapshot` entry contains backed-up files from two different directories — e.g., the agent touched files in both the base repo and the worktree earlier in the same session — the function returns a path from either on any given call. This could cause `parseClaudeProjectCWD` to produce a different working directory on successive polls and make the pane header flicker between two branches.

In the typical case where all non-auxiliary paths share a parent directory the non-determinism doesn't matter. But the name "latest" implies ordering that doesn't exist. Should this sort the keys (to get a stable selection) or, since `latestClaudeToolPath` over the `assistant` content array is already deterministic and ordered by event time, should the `file-history-snapshot` path be treated as lower-priority than the array-based result?

**[Minor question] `PostOpenVSCode` does not self-heal a stale sticky path**

`GetGitInfo` (lines 147–157 of the diff) now clears and retries when `inspectGitContextForSession` fails on the resolved path. `PostOpenVSCode` has the same `resolvePreferredCWD` call but no equivalent retry: if the sticky worktree no longer exists and the user clicks "Open in VS Code" before the next git-info poll clears the stale state, the request passes the deleted path straight to the `code` binary.

The practical window is small given poll frequency. Is the asymmetry intentional (VS Code's own error message is sufficient feedback), or should `PostOpenVSCode` apply the same fallback?

**[Minor question] `findClaudeSessionMeta` stops on any individual file read error**

`internal/session/local.go`

```go
data, err := os.ReadFile(path)
if err \!= nil {
    return nil, fmt.Errorf("read claude session metadata %q: %w", path, err)
}
```

A transient read error (race between `ReadDir` and a concurrent write, or a permission issue) on any `.json` file in `~/.claude/sessions/` causes the entire scan to abort rather than skipping that file and continuing. The error propagates to `claudeSessionCWD`, which returns it to `interactiveAgentSessionCWD`, which logs it and causes `resolvePreferredCWD` to fall back to the recalled sticky state. In practice the fallback is safe, but the error reaching the log on every poll could be noisy. Should individual file read errors use `continue` instead?

**[Minor question] `claudeBashCommandDir` compiles its regex on every call**

`internal/session/local.go`

```go
func claudeBashCommandDir(command string) string {
    matches := regexp.MustCompile(`(?:^|&&)\s*cd\s+...`).FindStringSubmatch(command)
```

The `regexp.MustCompile` call is inside the function body and runs on every invocation. Given that `parseClaudeProjectCWD` scans one or more transcript lines per poll, this compiles the same regexp repeatedly. Moving it to a package-level `var` (matching the style of `validShellPath` and `validTmuxSessionName`) would be a small cleanup. Worth doing here or tracked as a follow-up?

---

### Overall

The three previous concerns are cleanly resolved with matching tests. The main finding from the new Claude transcript parsing code is the non-deterministic map iteration in `latestClaudeTrackedPath` — worth settling before merge since it could cause observable flickering in edge cases. The remaining three items are minor.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/10088c72dc2fd1e62ad06d09cc60b2c1d144877e

Changes made:
- made Claude tracked-file selection deterministic so header resolution does not depend on Go map iteration order
- made `findClaudeSessionMeta` skip unreadable per-session JSON files and continue scanning
- moved the Claude `cd ... &&` regex to a package-level compiled regexp
- applied the same stale sticky-path self-healing to `PostOpenVSCode` that `GetGitInfo` already uses

Verification run locally:
- `make check`

tomo-chan (commented):
## Code Review — Follow-up (commit 10088c75)

### Previous Concerns — Resolved

All four concerns from the second review are addressed in commit 10088c75:

- **`latestClaudeTrackedPath` non-deterministic** ✓ — paths are collected into a slice, sorted with `sort.Strings`, and the lexicographically last entry is returned. Stable and predictable.
- **`findClaudeSessionMeta` stops on individual read errors** ✓ — changed to `continue` so a transient read failure on one session file no longer aborts the whole scan.
- **`claudeBashCommandDir` compiles regex per call** ✓ — moved to `var claudeBashCDPattern = regexp.MustCompile(...)` at package level, consistent with `validShellPath` and friends.
- **`PostOpenVSCode` stale-path not self-healed** ✓ — extracted `resolveValidatedPreferredCWD` and updated both `GetGitInfo` and `PostOpenVSCode` to call it. `TestPostOpenVSCode_StaleStickyWorktreeFallsBackToPaneCWD` is added. The wrapper is a cleaner solution than inlining the fallback in each handler.

Also noted: `readFileFn` is now injectable alongside the other function-level overrides, which makes Claude transcript tests fully isolated without touching the real filesystem.

---

### Minor Notes

**[Minor question] `resolveValidatedPreferredCWD` re-runs `inspectGitContextForSession` on the path the caller is about to inspect again**

```go
// resolveValidatedPreferredCWD
if _, err := h.inspectGitContextForSession(sess, targetCWD); err == nil {
    return targetCWD   // caller will inspect targetCWD again immediately
}
```

When the sticky path is valid, the validation call inside `resolveValidatedPreferredCWD` and the subsequent call in `GetGitInfo` (or `PostOpenVSCode`) both inspect the same path. That's +1 git subprocess per poll for sessions with an active sticky override. The early return (`if targetCWD == cwd { return cwd }`) avoids this overhead when no sticky is active, so the common non-sticky path is unaffected. Is the +1 call accepted as the cost of the shared wrapper design, or is there a plan to thread the context through to the caller?

**[Minor question] `latestClaudeTrackedPath` ties on alphabetical order — test doesn't exercise it**

The determinism fix is correct. For the common case (all backed-up files in the same worktree), any stable selection gives the right directory. For the uncommon case where an agent has backed up files from two different root directories in the same session, the alphabetically larger path wins.

`TestParseClaudeProjectCWD_PrefersLatestTrackedFileBackup` uses a single non-auxiliary entry, so it doesn't cover this tie-breaking. A two-entry test (paths from different directories) would pin the rule explicitly and guard against regressions if the selection strategy changes. Worth adding, or is the edge case uncommon enough to leave undocumented in tests?

---

### Overall

All previous concerns are resolved with matching tests. The two notes above are genuinely minor — neither blocks merge. The PR is in good shape.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/af472f558f9d89539a9d6bc2c5dbe54cceeb7f62

Changes made:
- removed the extra successful git-context inspection on sticky paths by having `resolveValidatedPreferredCWD` return the already-inspected context to `GetGitInfo`
- added an explicit Claude tracked-backup tie-break test covering the deterministic alphabetical-last selection across two directories

Verification run locally:
- `make check`

tomo-chan (commented):
## Code Review — Follow-up (commit af472f58)

### Previous Concerns — Resolved

Both minor concerns from the third review are addressed:

- **Double `inspectGitContextForSession` on the sticky-active path** ✓ — `resolveValidatedPreferredCWD` now returns `(string, *session.GitContext)`. `GetGitInfo` receives the already-inspected context directly and uses `*ctx` without a second round-trip to git. `PostOpenVSCode` discards the context with `_`, which is correct since it needs only the path.
- **`latestClaudeTrackedPath` tie-breaking not exercised by tests** ✓ — `TestLatestClaudeTrackedPath_IsDeterministic` and `TestParseClaudeProjectCWD_TrackedBackupTieBreakUsesAlphabeticalLastPath` both confirm the alphabetically-last selection and guard against future regressions.

---

### One Remaining Issue

**[Question] Real developer path in `TestLatestClaudeTrackedPath_IsDeterministic`**

`internal/session/local_test.go`

```go
backups := map[string]json.RawMessage{
    "/tmp/repo-main/AGENTS.md":       {},
    "/tmp/repo-worktree/README.md":   {},
    "/Users/username/.claude/plans/x.md": {},  // real developer path
}
```

DEVELOPMENT.md requires: *"Do not record real local or remote directory paths from a developer machine or environment in tests, fixtures, docs, screenshots, PR bodies, review replies, or commit history."*

`/Users/username/.claude/plans/x.md` embeds the actual developer username and home directory. The path is only a string literal being filtered by `isClaudeAuxiliaryPath` — it is never accessed on disk, so correctness is not affected on any machine. But it violates the documented rule and will be visible in git history.

A placeholder such as `/Users/demo/.claude/plans/x.md` satisfies the test requirement without recording a real path.

---

### Overall

All previous concerns are resolved. The only remaining item before merge is replacing `/Users/username` with a sanitized placeholder in the new test, per DEVELOPMENT.md.

tomo-chan:
Follow-up: history rewritten to remove developer-specific fixture paths from earlier commits.

Updated branch head:
- https://github.com/tomo-chan/panemux/commit/7a9aa98b444182ed18c4291c9e8c9448f686c14c

What changed:
- rewrote the PR branch history so developer-specific fixture paths no longer appear in prior commits in this branch range
- kept the same sanitized tree content as the latest reviewed state; this was a history cleanup, not a behavioral code change

Verification already run on the sanitized tree before the force-push:
- `make check`

tomo-chan (commented):
## Code Review — Follow-up (commits 7a9aa98b, 38a31c87)

### Previous Concern — Resolved

- **Real developer path in test fixture** ✓ — `/Users/tomo/.claude/plans/x.md` is replaced with `/Users/demo/.claude/plans/x.md`. Branch history rewritten so the real path no longer appears in any commit in this PR range.

The two documentation updates are a good addition:
- `DEVELOPMENT.md` now explicitly lists "PR comments" alongside the other surfaces where real paths must not appear.
- `AGENTS.md` surfaces the path-sanitization rule at the top of the "Read This First" list so future agents encounter it before writing test fixtures.

---

### Overall

No new findings. All concerns across four rounds of review are resolved. The PR is ready to merge.

PR #113, opened 2026-06-07
fix: reconnect disconnected SSH sessions

tomo-chan:
## Summary
- distinguish explicit SSH session exit from transport disconnects
- auto-recreate disconnected panes once and reconnect the WebSocket session
- keep restart UI only for explicit exits and add regression coverage

## Test Plan
- [x] go test ./internal/session ./internal/ws
- [x] npm test -- --run src/hooks/useWebSocket.test.ts src/hooks/useTerminal.test.ts src/components/TerminalPane.test.tsx
- [x] make test-frontend
- [x] make build-frontend
- [x] make lint-go

### DIFF
--- a/internal/session/tmux_ssh.go
+++ b/internal/session/tmux_ssh.go
@@ -82,10 +82,10 @@ func NewTmuxSSH(id, title, tmuxSession string, cfg SSHConfig) (*TmuxSSHSession, 
 		jumpClient:     jumpClient,
 	}
 
-	monitorSSHSession(sess, pw, func() {
+	monitorSSHSession(sess, pw, func(state State) {
 		s.mu.Lock()
 		defer s.mu.Unlock()
-		s.state = StateExited
+		s.state = state
 	})
 
 	return s, nil
--- a/internal/ws/handler.go
+++ b/internal/ws/handler.go
@@ -172,7 +172,7 @@ func (h *Handler) forwardTerminalOutput(
 		log.Printf("session %s stream closed in state %s", sessionID, sess.State())
 	}
 
-	h.sendStatus(conn, "exited")
+	h.sendStatus(conn, string(sess.State()))
 }
 
 func (h *Handler) pipeWebSocketToTerminal(
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -396,6 +396,11 @@ During replayed reconnect output, xterm stdin is suppressed until the replay end
 applied, so terminal query responses embedded in old output are not regenerated as fresh shell
 input.
 
+When an SSH-backed pane loses its transport unexpectedly, panemux classifies that session as
+`disconnected` instead of `exited`. The frontend performs one automatic recovery attempt by
+recreating the session and reconnecting the pane. Deliberate remote shell termination such as
+`exit` remains `exited` and shows the restart action instead of auto-reconnecting.
+
 ### Selection and copy behavior
 
 - Terminal text can be selected with the mouse using xterm.js standard selection behavior.
--- a/frontend/src/components/TerminalPane.test.tsx
+++ b/frontend/src/components/TerminalPane.test.tsx
@@ -1,5 +1,5 @@
 import { fireEvent, render } from '@testing-library/react'
-import { describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 class ResizeObserverMock {
   observe() {}
@@ -9,14 +9,12 @@ class ResizeObserverMock {
 
 vi.stubGlobal('ResizeObserver', ResizeObserverMock)
 
+const { mockUseTerminal } = vi.hoisted(() => ({
+  mockUseTerminal: vi.fn(),
+}))
+
 vi.mock('../hooks/useTerminal', () => ({
-  useTerminal: () => ({
-    handleResize: vi.fn(),
-    connected: true,
-    dims: null,
-    sessionExited: false,
-    restartSession: vi.fn(),
-  }),
+  useTerminal: mockUseTerminal,
 }))
 vi.mock('../hooks/useGitInfo', () => ({
   useGitInfo: () => ({ is_git: false }),
@@ -25,6 +23,17 @@ import { dropZoneStyle, PANE_DROP_ZONE_RATIO, resolvePaneDropEdge, TerminalPane 
 import { LayoutActionsContext, type LayoutActionsContextValue } from './SplitContainer'
 
 describe('TerminalPane drop zones', () => {
+  beforeEach(() => {
+    mockUseTerminal.mockReturnValue({
+      handleResize: vi.fn(),
+      connected: true,
+      dims: null,
+      sessionState: 'running',
+      reconnectFailed: false,
+      restartSession: vi.fn(),
+    })
+  })
+
   it('uses a top or bottom half-pane preview for horizontal edge drop zones', () => {
     expect(dropZoneStyle('top', false)).toMatchObject({ height: `${PANE_DROP_ZONE_RATIO * 100}%` })
     expect(dropZoneStyle('bottom', true)).toMatchObject({ height: `${PANE_DROP_ZONE_RATIO * 100}%` })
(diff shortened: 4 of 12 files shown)


internal/ws/handler.go
```diff
@@ -172,7 +172,7 @@ func (h *Handler) forwardTerminalOutput(
 		log.Printf("session %s stream closed in state %s", sessionID, sess.State())
 	}
 
-	h.sendStatus(conn, "exited")
+	h.sendStatus(conn, string(sess.State()))
 }
```
tomo-chan, line 176: After `pw.Close()` unblocks the pipeline, `forwardTerminalOutput` could call `sess.State()` here before `markExited()` runs in the monitor goroutine — returning `StateConnected` and sending `"connected"` to the frontend. The guard at line 170 already anticipates this for logging but not for `sendStatus`. Would calling `markExited()` before `pw.Close()` eliminate the race?

internal/session/tmux_ssh.go
```diff
@@ -82,10 +82,10 @@ func NewTmuxSSH(id, title, tmuxSession string, cfg SSHConfig) (*TmuxSSHSession,
 		jumpClient:     jumpClient,
 	}
 
-	monitorSSHSession(sess, pw, func() {
+	monitorSSHSession(sess, pw, func(state State) {
 		s.mu.Lock()
 		defer s.mu.Unlock()
-		s.state = StateExited
+		s.state = state
 	})
```
tomo-chan, line 89: For a `TmuxSSHSession`, the remote tmux process typically survives an SSH transport drop, so `StateDisconnected` is the right classification here. Does `POST /api/sessions/:id/restart` reconnect to the surviving tmux session, or create a fresh one? If it creates a fresh one, users lose their terminal state on reconnect.

tomo-chan (commented):
## Code Review

### What's Good

- `classifySSHWaitError` is well-structured: `nil`, `*ssh.ExitError`, and `*ssh.ExitMissingError` all map to `StateExited`, while every other error (transport I/O, network drops) maps to `StateDisconnected`. The four-case unit tests cover all the critical boundaries without overlap.
- The concurrent-recovery guard (`reconnectingAfterDisconnectRef`) is correctly placed — the early-return prevents stacked restart calls when multiple `disconnected` events arrive before the first fetch resolves, and the `finally` block resets the flag regardless of outcome.
- Switching `useWebSocket` from an uncancellable `setTimeout(connect, ...)` to a tracked `reconnectTimerRef` is a good correctness fix. The old code could fire a reconnect attempt after the component unmounted; `clearReconnectTimer()` in the cleanup effect closes that gap.
- `docs/behavior.md` is updated in the same commit, keeping the documented spec in sync with the new behaviour.

---

### Issues & Questions

**[Question] `forwardTerminalOutput` may read `StateConnected` from `sess.State()` before `markExited()` runs**

In `monitorSSHSession` the sequence is:

```go
waitErr := sess.Wait()
pw.Close()               // closes pipe, unblocking the reader that feeds `updates`
markExited(classify...)  // updates session state
```

If the goroutine feeding the `updates` channel closes it before `markExited()` has run, `forwardTerminalOutput` will call `string(sess.State())` while the session is still in `StateConnected`. The frontend's `else` branch would then call `setSessionState('running')` and `setReconnectFailed(false)`, silently hiding the session end from the user.

The existing guard at line 170 (`if sess.State() != session.StateConnected`) already anticipates this scenario for logging purposes — but `sendStatus` (now changed to `string(sess.State())`) has no matching guard and would send `"connected"` to the frontend if the race fires.

Would calling `markExited()` before `pw.Close()` eliminate the race?

**[Question] TmuxSSH sessions now get `StateDisconnected` — does `/restart` reconnect to the existing remote tmux session?**

Both `NewSSH` and `NewTmuxSSH` now use `classifySSHWaitError`, so an SSH transport failure marks either session `StateDisconnected`. For a plain SSH session that is correct — the session is gone. For a `TmuxSSHSession`, the remote tmux process typically survives an SSH transport drop.

If `POST /api/sessions/:id/restart` creates a fresh tmux session rather than reconnecting to the existing one, users would lose their in-flight terminal state. Does the restart endpoint handle the TmuxSSH case separately, or does it always attach to an existing tmux session?

**[Minor question] `recoverDisconnectedSessionRef.current` is assigned during render; `sendRef.current` is updated in `useLayoutEffect`**

Both are valid React patterns (updating a ref during render is safe when it is not read in the same render), but the inconsistency inside the same hook could confuse future readers. Was the direct assignment intentional to avoid the one-render-behind lag that `useLayoutEffect` would introduce for the first `disconnected` event?

---

### Overall

The SSH disconnect classification and the WebSocket timer-tracking fix are solid. The main thing to confirm before merge is whether the `StateConnected` race in `forwardTerminalOutput` is safe (or whether `markExited` should be called before `pw.Close()`), and whether TmuxSSH restart correctly reconnects to the surviving tmux session.

internal/ws/handler.go
tomo-chan: Fixed in https://github.com/tomo-chan/panemux/commit/02cd63f3f70a54929975b7a126e8b5a91364d420 .

I moved `markExited(classifySSHWaitError(waitErr))` ahead of `pw.Close()` in `monitorSSHSession`, so the session state is updated before the pipe teardown unblocks `forwardTerminalOutput`. That removes the window where the WS handler could have observed `StateConnected` and sent a stale `"connected"` status.

Re-verified with `go test ./internal/session ./internal/ws` and `make lint-go`.

internal/session/tmux_ssh.go
tomo-chan: `POST /api/sessions/:id/restart` recreates the session from the stored pane config, and for `ssh_tmux` that flows back through `NewTmuxSSH(...)` into `tmuxSSHCommand()`.

That command is `tmux new-session -As '<session>'`, so it attaches to the surviving remote tmux session when it already exists and only creates one when absent. In other words, the reconnect path is intended to reattach to the existing tmux state rather than replace it.

tomo-chan (commented):
## Code Review (follow-up)

### What's Resolved

Both questions from the previous review are addressed:

- **`StateConnected` race in `forwardTerminalOutput`** — fixed in `02cd63f1`. Moving `markExited(classifySSHWaitError(waitErr))` before `pw.Close()` in `monitorSSHSession` guarantees the session state is written before the pipe teardown unblocks the `updates` reader. The handler's `string(sess.State())` now reliably reads the correct final state.

- **TmuxSSH restart behaviour** — answered: `tmuxSSHCommand` uses `tmux new-session -As '<session>'`, which attaches to the surviving remote session when it already exists and creates a fresh one only when absent. The auto-reconnect path correctly reattaches rather than clobbering a running tmux session.

---

### Remaining

**[Minor question] `recoverDisconnectedSessionRef.current` assigned during render (unaddressed from previous review)**

```ts
// sendRef — updated in a useLayoutEffect
useLayoutEffect(() => { sendRef.current = send }, ...)

// recoverDisconnectedSessionRef — updated directly during render
recoverDisconnectedSessionRef.current = recoverDisconnectedSession
```

The pattern works correctly in React, but the inconsistency with `sendRef` inside the same hook could confuse future readers. If the direct assignment is intentional (e.g., to avoid the one-render-behind lag that `useLayoutEffect` would introduce before the first `disconnected` event fires), a short inline comment explaining the difference would help future maintainers.

---

### Overall

The race fix in `02cd63f1` is the right call and the new ordering is clean. The minor question above is non-blocking — happy for this to merge as-is or with a brief comment added.

tomo-chan:
Addressed the maintainability note in https://github.com/tomo-chan/panemux/commit/87432d36a7c3bf514315e0d2db31fdc4c82d7c40 .

I added an inline comment above `recoverDisconnectedSessionRef.current = recoverDisconnectedSession` to make the intent explicit: that ref is read from the async status-frame path and needs the latest callback during render, whereas `sendRef` is synchronized through `useLayoutEffect` because it also updates the mounted terminal entry.

Re-verified with `npm test -- --run src/hooks/useTerminal.test.ts`.

tomo-chan (commented):
## Code Review (final)

All concerns from prior rounds are resolved:

- **`StateConnected` race** — fixed in `02cd63f1` (order swap in `monitorSSHSession`).
- **TmuxSSH restart** — confirmed correct via `tmux new-session -As` reattach semantics.
- **`recoverDisconnectedSessionRef` render-time assignment** — explained in `87432d3b` with an inline comment clarifying why the direct render-time update is intentional (status-frame handler can run before an effect flushes).

Nothing left to address. Good to merge.

PR #115, opened 2026-06-07
feat: add standalone desktop mode

tomo-chan:
## Summary

- add shared backend bootstrap for browser and desktop modes
- add a Wails-based standalone desktop mode for macOS and Linux
- document the loopback-only desktop transport and mode-specific behavior

## Test Plan

- [x] `go test ./...`
- [x] `make build-frontend`
- [x] `make build-desktop`


Makefile
```diff
@@ -1,9 +1,9 @@
-.PHONY: all build build-frontend build-backend dev clean run install-deps install-deps-ci \
+.PHONY: all build build-frontend build-backend build-desktop dev clean run install-deps install-deps-ci \
         test test-go test-frontend test-e2e \
         fmt fmt-go fmt-check-go \
         lint lint-go lint-go-deps lint-frontend \
         coverage coverage-go coverage-frontend \
-        check release-snapshot package
+        check release-snapshot package package-desktop dev-desktop run-desktop
```
tomo-chan: [Question] `package-desktop` is listed in `.PHONY` but no recipe is defined for it anywhere in the diff. `make package-desktop` will fail with "No rule to make target". Was this intended as a placeholder, or was the recipe accidentally omitted?

internal/app/bootstrap.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (r *Runtime) Errors() <-chan error {
+	return r.errCh
+}
+
+func (r *Runtime) Shutdown(ctx context.Context) error {
+	r.shutOnce.Do(func() {
+		err := r.Server.Shutdown(ctx)
+		r.Manager.CloseAll()
+		if errors.Is(err, http.ErrServerClosed) {
```
tomo-chan: [Minor question] `http.Server.Shutdown` returns `nil` on success or `ctx.Err()` on timeout — it does **not** return `http.ErrServerClosed`. The sentinel is emitted by `Serve`/`ListenAndServe` when shutdown is triggered, not by `Shutdown` itself. This `errors.Is` guard will never match; removing it avoids misleading readers.

internal/app/bootstrap.go
```diff
@@ -0,0 +1,144 @@
...
+	for _, pane := range panes {
+		sess, err := session.CreateFromConfig(pane, cfg.SSHConnections)
+		if err != nil {
+			log.Printf("Warning: failed to start session %s (%s): %v", pane.ID, pane.Type, err)
+			continue
+		}
+		manager.Add(sess)
+		log.Printf("Started session: %s (%s)", pane.ID, pane.Type)
+	}
+	return nil
```
tomo-chan: [Question] `startSessionsFromConfig` always returns `nil` — the only non-nil-return path would be after some future fatal error, but the loop currently swallows all errors with `continue`. The caller's `if err \!= nil` branch (including `manager.CloseAll()`) is dead code. Consider changing the signature to `func startSessionsFromConfig(...)` (no return) to eliminate the dead branch, or document the intent to propagate errors in future.

tomo-chan (commented):
## Code Review

### What's Good

- The `internal/app` package is a clean extraction that makes browser and desktop modes symmetric — both paths go through the same `Bootstrap` → `Start` → `Shutdown` lifecycle without duplication.
- Loopback-only binding (`127.0.0.1:0`) for desktop mode is the correct default. The `docs/security.md` addendum explains the threat model clearly and honestly.
- Swapping `embed.FS` for `fs.FS` in `server.go` is the right broadening — it unblocks test injection (`fstest.MapFS`) without needing a real embedded filesystem.
- `shutOnce`/`startOnce` are the right guard for a shared lifecycle object that can be handed to multiple callers.
- `bootstrap_test.go` covers the key invariants (loopback binding, ephemeral port, browser address, start+shutdown cycle).

---

### Issues & Questions

**[Question] `package-desktop` declared `.PHONY` but no target exists**

`Makefile` line 1 (in the diff) adds `package-desktop` to `.PHONY`:
```makefile
check release-snapshot package package-desktop dev-desktop run-desktop
```
But no `package-desktop:` recipe appears anywhere in the diff. Running `make package-desktop` will fail with "No rule to make target". Is this a placeholder for a future release-packaging step that was accidentally left undefined, or should the recipe be added now?

**[Question] `startSessionsFromConfig` always returns `nil` but has an `error` return type**

```go
func startSessionsFromConfig(cfg *config.Config, manager *session.Manager) error {
    for _, pane := range panes {
        if err \!= nil {
            log.Printf("Warning: ...")
            continue          // error swallowed, never returned
        }
    }
    return nil                // only possible return value
}
```
The caller's `if err := startSessionsFromConfig(...); err \!= nil { ... }` branch is dead code, and the `manager.CloseAll()` cleanup inside it is unreachable. Is the intent to propagate fatal session errors in future, or should this be a void function so the dead branch is removed?

**[Question] `Server.Start()` behavioral change: now returns `nil` instead of `http.ErrServerClosed`**

```go
// Before
if errors.Is(err, http.ErrServerClosed) {
    return http.ErrServerClosed
}
// After
if errors.Is(err, http.ErrServerClosed) {
    return nil
}
```
The old code propagated the sentinel on clean shutdown, which caused the original `runServer` to call `log.Fatalf` — arguably a bug. The new behavior is correct for the new `errCh` pattern. Confirming: are there any callers or tests outside this PR that specifically check for `http.ErrServerClosed` from `Start()`?

**[Minor question] `errors.Is(err, http.ErrServerClosed)` in `Runtime.Shutdown` is dead code**

`http.Server.Shutdown` returns `nil` on success or `ctx.Err()` on timeout — it does **not** return `http.ErrServerClosed`. That sentinel is returned by `Serve`/`ListenAndServe` (when `Shutdown` is called on them), not by `Shutdown` itself. The guard in `Runtime.Shutdown` will never match and is worth removing to avoid misleading readers.

**[Minor question] Missing test cases for `validateOptions`**

The new tests cover `--open` and `--port` rejection in desktop mode. Two cases from the documented behavior rules are untested:
- Unknown mode string (e.g. `--mode=foobar`) → expected error containing "unsupported mode"
- Browser mode with no conflicting flags → expected `nil` (happy-path confirmation)

Per DEVELOPMENT.md: "for every new validation rule, write both a failure test and a success test."

---

### Overall

The design is sound: shared bootstrap, loopback binding, Wails as a thin shell. The undefined `package-desktop` target is a concrete bug to fix before merge. The `errors.Is` dead check and `error` return on `startSessionsFromConfig` are cleanup items. The missing `validateOptions` happy-path/unknown-mode tests are the one place the DEVELOPMENT.md granularity rule is not yet met.

tomo-chan (commented):
## Code Review (follow-up)

All five findings from the previous review have been addressed in commit `bc0d6f8`.

### Resolved

- **`package-desktop` missing recipe** — `package-desktop: build-desktop` added to Makefile. ✓
- **`startSessionsFromConfig` dead `error` return** — signature changed to `func(...)`; dead caller branch removed. ✓
- **Dead `errors.Is(err, http.ErrServerClosed)` in `Runtime.Shutdown`** — removed; `r.shutdownErr` is now assigned directly from `r.Server.Shutdown(ctx)`. ✓
- **`validateOptions` test gaps** — `TestValidateOptionsAcceptsBrowserMode` and `TestValidateOptionsRejectsUnknownMode` added. ✓
- **`Server.Start()` behavioral change** — confirmed: all callers migrated to the `errCh` pattern, which handles `nil` on clean shutdown correctly. No outstanding concern. ✓

### Overall

No remaining findings. Good to merge.

Makefile
tomo-chan: Fixed in https://github.com/tomo-chan/panemux/commit/bc0d6f854fd8300a92a4cda9fe63084dfce9ad1a

Added the missing `package-desktop` recipe so `make package-desktop` now resolves to `build-desktop` instead of failing with "No rule to make target".

Verified with:
- `go test ./...`
- `make build-desktop`

internal/app/bootstrap.go
tomo-chan: Fixed in https://github.com/tomo-chan/panemux/commit/bc0d6f854fd8300a92a4cda9fe63084dfce9ad1a

Changed `startSessionsFromConfig` to return no value and removed the dead caller branch in `Bootstrap`, so the implementation now matches the warning-and-continue behavior it actually uses.

While updating this area, I also added the missing `validateOptions` tests for unknown mode rejection and the browser-mode success path.

Verified with:
- `go test ./...`
- `make build-desktop`

internal/app/bootstrap.go
tomo-chan: Fixed in https://github.com/tomo-chan/panemux/commit/bc0d6f854fd8300a92a4cda9fe63084dfce9ad1a

Removed the dead `errors.Is(err, http.ErrServerClosed)` branch from `Runtime.Shutdown`. `Shutdown` now stores the direct result from `r.Server.Shutdown(ctx)` and still closes the session manager exactly once.

Verified with:
- `go test ./...`
- `make build-desktop`

tomo-chan (commented):
## Code Review

### What's Good

- **Shared bootstrap** (): browser and desktop modes go through the same  →  →  lifecycle. No code duplication, and the symmetry is clear.
- **Loopback-only binding** for desktop mode (): correct default. The threat model in `docs/security.md` is honest and complete.
- **`embed.FS` → `fs.FS`** in `server.go`: the right broadening — enables `fstest.MapFS` injection in tests without touching the embed directive.
- **`shutOnce`/`startOnce`** on `Runtime`: prevents double-start or double-shutdown regardless of caller ordering.
- **`validateOptions`** enforces the documented mode constraints (`--mode=desktop` is incompatible with `--open` and `--port`) with tests covering both the rejection paths and the browser-mode happy path.
- **`startSessionsFromConfig`** is now a void function — the previously dead `error` return and its unreachable cleanup branch are gone.
- **`Runtime.Shutdown`** is clean — the unreachable `errors.Is(err, http.ErrServerClosed)` guard has been removed.
- **`package-desktop: build-desktop`** is now defined, making the `.PHONY` declaration valid.
- **Test coverage**: `bootstrap_test.go` exercises loopback binding, ephemeral port assignment, browser address configuration, and the full start+shutdown cycle. `server_serve_test.go` verifies the new `Serve(listener)` path end-to-end.

---

### Issues & Questions

None.

---

### Overall

All five findings from the previous review have been addressed. No new issues found. Good to merge.

frontend/src/components/PaneHeader.test.tsx
```diff
@@ -202,4 +202,27 @@ describe('PaneHeader VSCode button', () => {
...
+          repo_url: 'https://github.com/example/panemux',
+          branch: 'main',
+        }}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('link', { name: 'panemux' }))
+    expect(browserOpenURL).toHaveBeenCalledWith('https://github.com/example/panemux')
+    delete (window as Window & { runtime?: { BrowserOpenURL?: (url: string) => void } }).runtime
+  })
```
tomo-chan, line 227: [Minor question] This covers the Wails-available path. The complementary test — when `window.runtime.BrowserOpenURL` is absent, `event.preventDefault` should NOT be called so normal anchor navigation is preserved — is missing. DEVELOPMENT.md requires both a success and a failure test for each new validation rule.

tomo-chan (commented):
## Code Review

### What's Good

- **Build tag discipline** (`Makefile`): `-tags production` for the release binary and `-tags dev` for the dev run target follows Wails conventions. The Darwin-only `CGO_LDFLAGS=-framework UniformTypeIdentifiers` guard is the correct way to scope macOS linker flags without breaking Linux builds.
- **`redirectOnce.Do` in `OnDomReady`** (`desktop_wails.go`): wrapping the JS redirect in a `sync.Once` is the right fix for WebViews that fire `DomReady` on each internal navigation, preventing an infinite reload loop.
- **`openExternalHeaderLink`** (`PaneHeader.tsx`): the `https?://` guard before delegating to `BrowserOpenURL` is a good safety check. The `runtime?.BrowserOpenURL?.(href)` optional-chain pattern correctly no-ops in browser mode where `window.runtime` is undefined.
- **Test** (`PaneHeader.test.tsx`): the new test mocks `window.runtime.BrowserOpenURL`, fires a click, asserts the function was called with the correct URL, and cleans up — covering the Wails path end-to-end.

---

### Issues & Questions

**[Minor question] Missing test for the non-Wails path in `InlineHeaderLink`**

`frontend/src/components/PaneHeader.test.tsx`

The new test covers the case where `window.runtime.BrowserOpenURL` is present (Wails desktop). The complementary case — where `BrowserOpenURL` is absent (normal browser mode) — has no dedicated test. Per DEVELOPMENT.md: *"for every new validation rule, write both a failure test and a success test."*

The key behaviour to confirm is that when `hasWailsBrowserOpen` is `false`, `event.preventDefault()` is **not** called (so the anchor's default navigation is preserved). This can be verified by spying on the click event's `preventDefault` method:

```typescript
it('does not intercept header link clicks when BrowserOpenURL is not available', () => {
  // window.runtime is not set (default browser environment)
  render(<PaneHeader {...defaultProps} gitInfo={{ is_git: true, repo: 'panemux',
    repo_url: 'https://github.com/example/panemux', branch: 'main' }} />)

  const link = screen.getByRole('link', { name: 'panemux' })
  const preventDefaultSpy = vi.spyOn(MouseEvent.prototype, 'preventDefault')
  fireEvent.click(link)
  expect(preventDefaultSpy).not.toHaveBeenCalled()
  preventDefaultSpy.mockRestore()
})
```

---

### Overall

One minor test-coverage gap per DEVELOPMENT.md's "both paths" rule. No correctness issues found in the new commits.

tomo-chan (commented):
## Code Review

### What's Good

- **`desktop_wails.go` refactoring**: Splitting `runDesktop` into `bootstrapDesktopApp`, `runDesktopShell`, and `buildDesktopOptions` gives each function a single responsibility and makes the Wails options straightforward to reason about.
- **`redirectOnce` in `OnDomReady`**: Correct guard — after `window.location.replace(...)` navigates to the backend, the WebView's `DomReady` fires again for the backend's pages. Without `redirectOnce` this creates an infinite reload loop.
- **`Runtime.Shutdown` idempotency via `shutOnce`**: Any double-call from `defer shutdown()` + `OnShutdown` is safe. Good defensive design.
- **`Server.Start()` / `Server.Serve()` returning `nil` on `ErrServerClosed`**: The `waitForShutdown` drain of `runtimeApp.Errors()` after `Shutdown()` now correctly receives `nil` without callers needing `errors.Is` checks.
- **Missing test from previous round now added** (`test: cover browser header link fallback`): the `does not intercept header links when Wails runtime is unavailable` test fills the gap flagged in the earlier review.

---

### Issues & Questions

**[Question] `--help` / `-h` produces a fatal error instead of usage text**

`parseOptions` uses `flag.ContinueOnError` + `flags.SetOutput(io.Discard)`. When the user passes `--help` or `-h`, `flags.Parse` returns `flag.ErrHelp`. The code does not special-case this, so execution reaches:

```
// main.go
log.Fatalf("Failed to parse options: parse flags: flag: help requested")
```

No flag descriptions are shown; the process exits with code 1. The old `flag.Parse()` (default `ExitOnError`) printed full usage to stderr and exited with code 2 — the conventional CLI behaviour.

Was `io.Discard` intentional (e.g., to suppress the default `flag` error prefix and provide a cleaner message)? If so, one option is to special-case `flag.ErrHelp`:

```go
if err := flags.Parse(args); err \!= nil {
    if errors.Is(err, flag.ErrHelp) {
        flags.SetOutput(os.Stderr)
        flags.Usage()
        os.Exit(0)
    }
    return cliOptions{}, fmt.Errorf("parse flags: %w", err)
}
```

**[Question] Graceful-shutdown timeout changes exit code from 0 to 1**

Old `runServer` on SIGINT:
```go
if err := srv.Shutdown(ctx); err \!= nil {
    log.Printf("Shutdown error: %v", err)  // warn and continue → exit 0
}
manager.CloseAll()
```

New `waitForShutdown`:
```go
if err := runtimeApp.Shutdown(ctx); err \!= nil {
    return fmt.Errorf("shutdown error: %w", err)  // → log.Fatalf → exit 1
}
```

If the 5-second graceful-shutdown window is exceeded (e.g., a stuck WebSocket client), the old code logged a warning and exited cleanly; the new code exits with code 1. Systemd/launchd treats exit 1 as a crash and may restart the process or trigger an alert.

Session cleanup is unaffected — `Manager.CloseAll()` is called inside `Runtime.Shutdown` via `shutOnce` regardless of whether the HTTP shutdown times out.

Is the exit-code change deliberate? If the intent is to preserve the old behaviour, the fix is straightforward:

```go
if err := runtimeApp.Shutdown(ctx); err \!= nil {
    log.Printf("Shutdown error: %v", err)
}
_ = <-runtimeApp.Errors()
return nil
```

**[Minor question] `wailsruntime.Quit(ctx)` fires on clean exit as well as on error**

```go
OnStartup: func(ctx context.Context) {
    go func() {
        if err := <-runtimeApp.Errors(); err \!= nil {
            log.Printf("desktop backend error: %v", err)
        }
        wailsruntime.Quit(ctx)  // called whether err is nil or non-nil
    }()
},
```

On a normal window close, `OnShutdown` stops the backend, which puts `nil` into `errCh`. The goroutine wakes up with `nil`, skips the log, and calls `Quit` on an already-shutting-down application. Wails v2 should handle this as a no-op. Was the intent to call `Quit` only on backend error (`if err \!= nil { wailsruntime.Quit(ctx) }`) or always (as a defensive "ensure window closes" belt-and-suspenders)?

**[Minor question] `defer shutdown()` bypasses `shutdownOnce` — double log on shutdown error**

```go
shutdown := func() {
    ...
    if shutdownErr := runtimeApp.Shutdown(ctx); shutdownErr \!= nil {
        log.Printf("desktop shutdown error: %v", shutdownErr)
    }
}
defer shutdown()                  // calls shutdown() directly

OnShutdown: func(context.Context) {
    shutdownOnce.Do(shutdown)     // also calls shutdown()
},
```

On normal close, `shutdown()` runs twice. `Runtime.Shutdown` is idempotent via its own `shutOnce`, so no double teardown occurs. However, if `runtimeApp.Shutdown` returns an error the `"desktop shutdown error"` log line fires twice. Should the defer also go through `shutdownOnce`?

```go
defer func() { shutdownOnce.Do(shutdown) }()
```

---

### Overall

The architecture is clean and all previous round findings have been addressed. Two questions remain before merge: the `--help` UX regression and the graceful-shutdown exit-code change. Two minor questions about the Wails shutdown flow are also worth a quick call.

frontend/src/components/PaneHeader.test.tsx
tomo-chan: Fixed in https://github.com/tomo-chan/panemux/commit/c5393d65e528b30e97be714102c7fd2b3b377bae

Added the complementary browser-mode test for `InlineHeaderLink` so the non-Wails path is covered too. The new test verifies that when `window.runtime.BrowserOpenURL` is unavailable, clicking the repo link does not call `preventDefault`, which preserves normal anchor navigation.

Verified with:
- `cd frontend && npm test -- PaneHeader.test.tsx`
- `make build-frontend`

tomo-chan (commented):
## Code Review

### What's Good

- **Build unification** (`build: unify desktop and browser binary`, `build: remove desktop build target`): folding `build-desktop` into the regular `build-backend` via `-tags production` gives a single binary that supports both modes. The `$(BUILD_TAGS)` / `$(BUILD_ENV)` conditionals keep Darwin and Linux scoped without polluting the Makefile for other platforms.
- **`desktop_unsupported.go`**: the stub `runDesktop` returning an error on non-Darwin/non-Linux is the correct companion to `desktop_wails.go`'s `//go:build darwin || linux`, so the build stays unconditional everywhere.

---

### Issues & Questions

**[Question] `--help` produces a fatal error instead of usage text** (carried from round 5, unaddressed)

`main.go` — `parseOptions`:

```go
flags := flag.NewFlagSet("panemux", flag.ContinueOnError)
flags.SetOutput(io.Discard)
...
if err := flags.Parse(args); err \!= nil {
    return cliOptions{}, fmt.Errorf("parse flags: %w", err)
}
```

When the user passes `--help` or `-h`, `flags.Parse` returns `flag.ErrHelp`. Because `SetOutput(io.Discard)` suppresses output and there is no `errors.Is(err, flag.ErrHelp)` guard, control falls through to `log.Fatalf("Failed to parse options: parse flags: flag: help requested")` — exit 1, no usage text shown. The previous `flag.Parse()` (default `ExitOnError`) printed usage to stderr and exited 2, the conventional CLI behaviour. Was `io.Discard` intentional, and if so, is a `flag.ErrHelp` special case on the roadmap?

**[Question] Graceful-shutdown timeout now exits with code 1 instead of 0** (carried from round 5, unaddressed)

`main.go` — `waitForShutdown`:

```go
case <-sigCh:
    ...
    if err := runtimeApp.Shutdown(ctx); err \!= nil {
        return fmt.Errorf("shutdown error: %w", err)  // → log.Fatalf → exit 1
    }
```

Old behaviour (5 s timeout exceeded): `log.Printf("Shutdown error: …")`, continue, exit 0. New behaviour: `log.Fatalf` → exit 1. Systemd/launchd may treat exit 1 as a crash and trigger restart/alert. Session cleanup is still safe — `Manager.CloseAll()` runs inside `Runtime.Shutdown` via its own `shutOnce` regardless of the HTTP shutdown timeout. Is the exit-code change deliberate?

**[Minor question] Test plan references a removed Makefile target**

The PR description has:

```
- [x] `make build-desktop`
```

`build-desktop` was removed in `build: remove desktop build target`. The correct target is now `make build` (or `make build-backend`). Worth updating so the test plan stays reproducible for future readers.

**[Minor question] `wailsruntime.Quit` fires unconditionally on backend stop** (carried from round 5)

`desktop_wails.go` — `OnStartup` goroutine:

```go
go func() {
    if err := <-runtimeApp.Errors(); err \!= nil {
        log.Printf("desktop backend error: %v", err)
    }
    wailsruntime.Quit(ctx)  // called whether err is nil or non-nil
}()
```

On normal window close, `OnShutdown` stops the backend, putting `nil` into `errCh`. The goroutine wakes, skips the log, and calls `Quit` on an already-shutting-down app. Is this intentional belt-and-suspenders (Wails should no-op it), or should `Quit` only be called on `err \!= nil`?

**[Minor question] `defer shutdown()` bypasses `shutdownOnce` — double log on error** (carried from round 5)

`desktop_wails.go` — `runDesktopShell`:

```go
defer shutdown()                  // direct call
...
OnShutdown: func(context.Context) {
    shutdownOnce.Do(shutdown)     // guarded call
},
```

On a normal window close, `OnShutdown` runs `shutdown()` via `shutdownOnce.Do`. Afterward, `defer shutdown()` calls it directly again. `runtimeApp.Shutdown` is idempotent, so no double teardown occurs — but if `runtimeApp.Shutdown` returns an error, the `"desktop shutdown error"` log line fires twice. Should the defer also go through `shutdownOnce`?

```go
defer func() { shutdownOnce.Do(shutdown) }()
```

---

### Overall

The two build commits cleanly unify the binary. However, two `[Question]` items flagged as needed-before-merge in round 5 — the `--help` UX regression and the shutdown exit-code change — remain open in `main.go` with no code change or author reply. The two minor questions about the Wails shutdown flow in `desktop_wails.go` are also still open.

PR #116, opened 2026-06-10
fix: resolve remote Claude worktree headers

tomo-chan:
## Summary

- resolve remote `claude` worktree detection for `ssh` and `ssh_tmux` panes
- keep pane header git branch and linked PR metadata aligned with the active Claude worktree
- document the remote Claude resolver behavior alongside the existing Codex notes

## Root Cause

Remote pane header worktree overrides had special handling for Codex session logs, but Claude only had a local-only transcript resolver. In `ssh` and `ssh_tmux` panes panemux therefore fell back to descendant process cwd scanning, which often stayed pinned to the base pane directory instead of the active sibling worktree.

## Changes

- add remote Claude session metadata and transcript parsing in the SSH session resolver
- reuse Claude transcript path extraction for both local and remote flows
- add regression tests for remote SSH and SSH+tmux Claude worktree resolution
- update architecture and behavior docs to describe the remote Claude path

## Validation

- `make test-go`

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -187,6 +187,8 @@ This ordering is intentional and reflects observed Codex behavior as of `codex-t
 
 Panemux treats that `workdir` field as the strongest available signal for the active worktree because it is the only one that changes when Codex executes tools inside a sibling Git worktree while the parent interactive process remains attached to the original pane directory. If a future Codex release changes this logging contract, compare new logs against these three fields before changing the resolver so behavior remains reviewable and intentional.
 
+For interactive Claude sessions, all four session implementations may inspect `~/.claude/sessions/<pid>.json` to locate the active transcript under `~/.claude/projects/...`. Panemux derives the active worktree from the latest non-auxiliary file path or `Bash` tool `cd ... &&` target recorded in that transcript.
+
 ### `useWorkspaceAttentionMonitor`
 
 Opens lightweight background WebSocket subscriptions for every pane ID across all workspaces and
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -481,6 +481,8 @@ When a pane moves to a different parent node, the component may be remounted by 
 - Codex log resolution currently prefers the most recent `response_item.payload.arguments.workdir` from `exec_command` tool calls, then falls back to `turn_context.cwd`, then `session_meta.cwd`.
 - This ordering exists because the interactive Codex process and its thread metadata may stay on the original pane directory even while tool calls are executing inside a sibling Git worktree.
 - If future Codex versions change their session-log schema or start updating `turn_context.cwd` to the active worktree reliably, compare the three fields above before changing panemux's resolver order.
+- For interactive Claude flows across all four pane types, panemux may derive the active worktree from `~/.claude/sessions/<pid>.json` plus the matching transcript under `~/.claude/projects/...`.
+- Claude transcript resolution prefers the latest non-auxiliary tracked file path or `Bash` tool `cd ... &&` target recorded in that transcript.
 - If the agent exits, no longer has an eligible worktree, or the resolved worktree is not a sibling worktree of the pane's current repository, panemux falls back to the pane's own working directory immediately.
 - For SSH panes, panemux resolves interactive agent processes from the remote process list on the current SSH connection.
 - For SSH+tmux panes, panemux resolves interactive agent processes only from the currently active remote tmux pane.
--- a/internal/session/local.go
+++ b/internal/session/local.go
@@ -665,22 +665,22 @@ func readClaudeProjectCWD(path string) (string, error) {
 
 func parseClaudeProjectCWD(data []byte) (string, error) {
 	scanner := bufio.NewScanner(bytes.NewReader(data))
-	var latestPath string
+	var latest claudePathCandidate
 	for scanner.Scan() {
-		if candidate := claudeRecordPath(scanner.Bytes()); candidate != "" {
-			latestPath = candidate
+		if candidate := claudeRecordPath(scanner.Bytes()); candidate.Path != "" {
+			latest = candidate
 		}
 	}
 	if err := scanner.Err(); err != nil {
 		return "", fmt.Errorf("scan claude transcript: %w", err)
 	}
-	if latestPath == "" {
+	if latest.Path == "" {
 		return "", nil
 	}
-	if info, err := os.Stat(latestPath); err == nil && info.IsDir() {
-		return latestPath, nil
+	if latest.IsDir {
+		return latest.Path, nil
 	}
-	return filepath.Dir(latestPath), nil
+	return filepath.Dir(latest.Path), nil
 }
 
 type claudeProjectRecord struct {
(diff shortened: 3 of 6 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **`claudePathCandidate` + `IsDir` refactoring** is an important correctness fix. The old `parseClaudeProjectCWD` called `os.Stat(latestPath)` to determine whether the resolved path was a directory — but `os.Stat` checks the *local* filesystem. When this function is now reused for remote transcripts in `remoteClaudeSessionCWD`, that call would have silently checked the wrong machine. The new approach derives directory-ness from the record type (`Bash` cd → `IsDir: true`; file-history-snapshot → `IsDir: false`), making the function correctly context-agnostic.

- **`remoteInteractiveAgentSessionCWD` dispatcher** is a clean extension point. Adding a new agent type in the future is a single `case`, with no changes to the caller in `activeRemoteWorkdirWithOutput`.

- **PID cross-check** in `remoteClaudeSessionMeta` (`if meta.PID \!= agentPID { return nil, nil }`) guards against reading a stale session file that was not cleaned up after a previous Claude process exited on the same session path pattern.

- **Test coverage** spans both SSH and SSH+tmux pane types, and both record kinds (file-history-snapshot and Bash `cd` command). That combination matrix is exactly right for the two code paths in `claudeRecordPath`.

---

### Issues & Questions

**[Question] Tilde not expanded in the shell-quoted project path**

`remoteClaudeProjectPath` returns a string starting with `~/.claude/projects/...`. That value is then passed to `shellQuotePath`, producing a command like:

```sh
cat '~/.claude/projects/-repo-main/session-123.jsonl'
```

On a POSIX shell, `~` is **not** expanded inside single quotes — the shell would try to open a literal path named `~/.claude/...` relative to the current directory, which does not exist, so `cat` fails. The tests pass because `fakeSSHRunner` matches command strings exactly (including the quoting), not because a real shell ran them.

Compare `remoteClaudeSessionMeta`, which correctly keeps the tilde unquoted:

```go
sessionPath := fmt.Sprintf("~/.claude/sessions/%d.json", agentPID)
out, err := run("cat " + sessionPath)   // tilde expands — correct
```

And `remoteCodexSessionCWD`, which works because its `sessionPath` comes from `remoteOpenFiles` as a full absolute path (no tilde involved).

One approach to fix: shell-quote only the variable components and leave the `~/.claude/projects/` prefix unquoted.

```go
// in remoteClaudeSessionCWD:
dirPart := claudeProjectDirName(sessionMeta.CWD)
filePart := sessionMeta.SessionID + ".jsonl"
projectPath := shellQuotePath(dirPart + "/" + filePart)
out, err := run("cat ~/.claude/projects/" + projectPath)
```

Is this a known limitation, or should the transcript read be expected to work on real SSH connections today?

---

**[Minor question] `encoding/json` import is out of alphabetical order**

```go
"crypto/sha1" //nolint:gosec // ...
+"encoding/json"
 "encoding/base64"
```

`encoding/base64` sorts before `encoding/json`. `gofmt` doesn't enforce import order, so this won't fail lint, but it diverges from the rest of the file's import style. Worth fixing in the same commit.

---

### Overall

The `IsDir` refactoring and dispatcher are well-designed. Before marking this ready-for-review, the tilde-in-single-quotes issue in `remoteClaudeSessionCWD` should be verified or fixed — the current implementation would silently return an error on any real SSH connection when trying to read the transcript, leaving the header unchanged as if Claude were not running.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/b03b592cd47ff6d805d175c0a4bbc25a1ee508db

- kept the `~/.claude/projects/` prefix unquoted so the remote shell still expands `~`
- shell-quoted only the variable suffix (`<project>/<session>.jsonl`) to preserve safety
- updated the SSH tests to assert the real command shape used for remote Claude transcript reads

Verified locally with:

- `make lint-go`
- `make test-go`

tomo-chan (commented):
## Code Review (follow-up)

Both issues raised in the previous pass have been addressed:

- **Tilde expansion** (`b03b592c`) — `remoteClaudeProjectReadCmd` now keeps the `~/.claude/projects/` prefix unquoted so the remote shell expands `~`, and passes only the variable suffix through `shellQuotePath`. The split between a logging helper (`remoteClaudeProjectPath`) and the command-generation helper (`remoteClaudeProjectReadCmd`) is a clean way to separate concerns.

- **Import order** (`266e1d5a`) — `encoding/base64` now appears before `encoding/json`, consistent with the rest of the file.

---

### Issues & Questions

**[Minor question] `meta.SessionID` used unvalidated in path construction**

`remoteClaudeProjectReadCmd` passes `meta.SessionID` directly to `filepath.Join`:

```go
func remoteClaudeProjectReadCmd(meta *claudeSessionMeta) string {
    return "cat ~/.claude/projects/" + shellQuotePath(
        filepath.Join(claudeProjectDirName(meta.CWD), meta.SessionID+".jsonl"),
    )
}
```

`filepath.Join` normalises `..` sequences, so a session ID like `../../../../etc/shadow` would result in `cat '../../etc/shadow.jsonl'` appended to `~/.claude/projects/`, ultimately resolving to a path outside the user's home directory on the remote machine. `shellQuotePath` prevents shell injection, but not directory traversal.

The same pattern exists in the local `claudeSessionCWD` path, so this is not introduced by the PR. And the threat model limits the impact — the session file is on the user's own trusted remote machine, so a legitimate user would never write a traversal payload there. Still, a narrow validation on `SessionID` (e.g. `^[a-zA-Z0-9_-]+$`, which matches real Claude session IDs) would make the safety argument explicit in code rather than in the threat model. Happy to file a follow-up issue for this if that's useful.

---

### Overall

Both blocking issues from the first pass are cleanly resolved. The remaining item above is pre-existing and low severity. This looks ready to move out of draft when you're satisfied with manual verification on a live SSH connection.

tomo-chan:
Addressed the remaining session-id path traversal concern in https://github.com/tomo-chan/panemux/commit/8e4f3acca49276bdae16f6b399b5f38935fb2090

- validate Claude `sessionId` with `^[a-zA-Z0-9_-]+$` before building local or remote transcript paths
- reject invalid IDs in both local and SSH-backed Claude workdir resolution paths
- add regression coverage for invalid local and remote Claude session metadata

Verified locally with:

- `make lint-go`
- `make test-go`

tomo-chan (commented):
## Code Review (final pass)

All three issues raised across the previous two passes are addressed:

- **Tilde expansion** (`b03b592c`) — `remoteClaudeProjectReadCmd` keeps `~/.claude/projects/` unquoted and quotes only the variable suffix.
- **Import order** (`266e1d5a`) — `encoding/base64` now precedes `encoding/json`.
- **Session ID path traversal** (`8e4f3acc`) — `validClaudeSessionID` (`^[a-zA-Z0-9_-]+$`) is applied in both `claudeSessionCWD` (local) and `remoteClaudeSessionCWD` (SSH) before the session ID is used in any path construction. Regression tests cover the traversal payload (`../../etc/shadow`) for both flows. The regex correctly accepts real Claude session ID shapes (UUID-style hex + hyphens).

---

### What's Good

The full change is clean end-to-end:

- The `claudePathCandidate` + `IsDir` refactoring removes the `os.Stat` call from `parseClaudeProjectCWD`, making it safe to reuse for remote transcript data. This is the right fix for the right reason.
- The `remoteInteractiveAgentSessionCWD` dispatcher is a clean seam: adding a future agent type requires one `case`, with no changes to the caller.
- Test coverage is complete: SSH and SSH+tmux pane types, file-history-snapshot and Bash-cd record kinds, valid and invalid session IDs — all four combinations are covered.

---

### Overall

No open findings. Ready to move out of draft whenever manual verification on a live SSH connection confirms the transcript read end-to-end.

PR #118, opened 2026-06-10
fix: normalize claude transcript project dir names

tomo-chan:
## Summary
- normalize Claude transcript project directory names so paths with `.` match the actual `~/.claude/projects/...` layout
- add local and SSH regression tests for dot-containing Claude cwd paths

## Test plan
- [x] make install-deps
- [x] make build-frontend
- [x] make test-go
- [x] make lint-go

### DIFF
--- a/internal/session/local.go
+++ b/internal/session/local.go
@@ -653,7 +653,8 @@ func parseExecCommandWorkdir(raw json.RawMessage) string {
 }
 
 func claudeProjectDirName(cwd string) string {
-	return strings.ReplaceAll(filepath.Clean(cwd), string(os.PathSeparator), "-")
+	name := strings.ReplaceAll(filepath.Clean(cwd), string(os.PathSeparator), "-")
+	return strings.ReplaceAll(name, ".", "-")
 }
 
 func readClaudeProjectCWD(path string) (string, error) {
--- a/internal/session/local_test.go
+++ b/internal/session/local_test.go
@@ -644,6 +644,73 @@ func TestGetActiveWorkdir_PrefersClaudeTranscriptWorktree(t *testing.T) {
 	assert.Equal(t, filepath.Dir(worktreeFile), cwd)
 }
 
+func TestClaudeProjectDirName_NormalizesDots(t *testing.T) {
+	assert.Equal(t, "-repo-main", claudeProjectDirName("/repo/main"))
+	assert.Equal(
+		t,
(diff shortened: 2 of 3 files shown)


internal/session/local.go
```diff
@@ -653,7 +653,8 @@ func parseExecCommandWorkdir(raw json.RawMessage) string {
 }
 
 func claudeProjectDirName(cwd string) string {
-	return strings.ReplaceAll(filepath.Clean(cwd), string(os.PathSeparator), "-")
+	name := strings.ReplaceAll(filepath.Clean(cwd), string(os.PathSeparator), "-")
+	return strings.ReplaceAll(name, ".", "-")
 }
```
tomo-chan, line 658: The fix assumes Claude replaces `.` with `-` when naming directories under `~/.claude/projects/`. If this was verified by observing Claude's actual directory layout, the fix is correct as written. The concern is forward-looking: if Claude changes this encoding in a future version, the lookup would silently fail (no error, just no transcript found) rather than failing noisily. Was this encoding verified from a real Claude session? A brief comment linking to a traceable test case or Claude changelog entry would help future maintainers know when to re-check.

tomo-chan (commented):
## Code Review

### What's Good

- The fix is a single-line addition to `claudeProjectDirName` — minimal surface area with no risk of collateral breakage across the two callers (`local.go` and `ssh.go`).
- Test coverage hits all three relevant levels: a unit test for the function itself (`TestClaudeProjectDirName_NormalizesDots`), a local session integration test, and an SSH session integration test. Having all three is valuable since the function is shared between both session types.
- The new SSH test (`TestActiveRemoteWorkdir_PrefersRemoteClaudeTranscriptWorktree_WithDotInCWD`) follows the exact pattern of its non-dot sibling and exercises the `fakeSSHRunner` correctly.

---

### Issues & Questions

**[Question] How was the dot→dash encoding confirmed as Claude's actual behavior?**

See inline comment.

**[Minor question] Is the substitution set complete?**

The function now handles `/` → `-` and `.` → `-`. If Claude also normalizes other characters (e.g. spaces in path segments, or other shell-special chars), those paths would silently yield no transcript rather than an error. Is `/` and `.` the confirmed-complete list of substitutions Claude applies?

---

### Overall

The fix is correct and the test coverage is solid. The two questions above are about verifiability and future-proofing rather than correctness — the PR looks good to merge once the encoding behavior is confirmed.

PR #121, opened 2026-06-11
fix: improve git context error diagnostics

tomo-chan:
## Summary
- improve git context lookup errors with structured cause, remediation, and execution details
- classify local and SSH git inspection failures and capture stderr for diagnostics
- add regression tests for local and remote git-info failure logging

## Test plan
- [x] `go test ./internal/session ./internal/api`
- [x] `make test-go`
- [x] `make lint-go`

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -944,9 +944,9 @@ func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error)
 	}, nil
 }
 
-func localGitOptionalMetadata(cwd string) ([]byte, []byte) {
+func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
 	branchCmd := exec.Command("git", "branch", "--show-current")
-	branchCmd.Dir = cwd
+	branchCmd.Dir = safeCWD
 	branchOut, err := branchCmd.Output()
 	if err != nil {
 		// Detached HEAD returns non-zero for --show-current. Treat that as
@@ -956,7 +956,7 @@ func localGitOptionalMetadata(cwd string) ([]byte, []byte) {
 	branchOut = bytes.TrimSpace(branchOut)
 
 	originCmd := exec.Command("git", "config", "--get", "remote.origin.url")
-	originCmd.Dir = cwd
+	originCmd.Dir = safeCWD
 	originOut, err := originCmd.Output()
 	if err != nil {
 		originOut = nil
@@ -1021,6 +1021,8 @@ func formatGitContextLookupError(logScope, source string, err error) string {
 		)
 	}
 
+	// Keep a defensive fallback here so future callers that wrap a non-GitContextError
+	// still produce an actionable log instead of regressing to a raw `%v` message.
 	return fmt.Sprintf(
 		`%s git context lookup failed: source=%q cause=%q remediation=%q raw_error=%q`,
 		logScope,
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -782,7 +782,7 @@ func (h *Handler) resolvePreferredCWD(sess session.Session, cwd string) string {
 
 	baseCtx, err := h.inspectGitContextForSession(sess, cwd)
 	if err != nil {
-		log.Printf("%s base context lookup failed: %v", logScope, err)
+		log.Printf("%s", formatGitContextLookupError(logScope, "base context lookup", err))
 		return cwd
 	}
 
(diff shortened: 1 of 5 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Structured `GitContextError` type** — replacing ad-hoc `fmt.Errorf` strings with a typed struct that carries `Transport`, `Cause`, `CauseMessage`, `Remediation`, `Operation`, and `Stderr` is the right move. Operators can now grep on stable field names like `cause=` instead of parsing free-form strings.
- **`ClassifyGitFailureCause` heuristics** — the pattern-matching on lowercase stderr maps directly to the most common user-visible failure modes. The conservative fallback to `GitContextCauseUnknown` (or `GitContextCauseGitMetadata` for local) avoids over-classifying edge cases.
- **SSH structured error protocol** — embedding `__PANEMUX_GIT_CONTEXT_ERROR__` as a sentinel and using `printf` with `2>&1` to capture git stderr into `$root`/`$common` before `exit $status` is a clean way to surface remote diagnostic context without a second round-trip. The `%%s` → `%s` Go/shell escaping is correct.
- **`singleLineLogValue`** — normalizing multi-line values by joining whitespace-separated tokens prevents log injection and broken log parsers. Small but important.
- **`Unwrap()` on `GitContextError`** — preserves the error chain so callers using `errors.Is`/`errors.As` on the wrapped original continue to work.
- **Log-output tests** — asserting on specific log fields (`cause=`, `remediation=`, `operation=`) rather than just checking HTTP status is exactly the right way to cover diagnostic improvements.
- **`fakeSSHRunner.Output` returns bytes+error** — now correctly mirrors `x/crypto/ssh.Session.Output`, which captures stdout even on non-zero exit. The change makes the fake a faithful stand-in for the real runner.

---

### Issues & Questions

**[Minor question] `localGitOptionalMetadata` parameter naming**

`internal/api/handler.go` — the function is called with:

```go
branchOut, originOut := localGitOptionalMetadata(safeCWD)
```

but the parameter inside is named `cwd`:

```go
func localGitOptionalMetadata(cwd string) ([]byte, []byte) {
    branchCmd.Dir = cwd
```

`safeCWD` and `cwd` (the original unsanitized path) are two distinct variables in the caller. A reader of `localGitOptionalMetadata` could confuse which one is in play. Does the parameter need renaming to `safeCWD` (or `dir`) to match the caller's intent, or is the generic name deliberate?

**[Minor question] Fallback branch in `formatGitContextLookupError` — intentional safety net?**

`internal/api/handler.go`:

```go
func formatGitContextLookupError(logScope, source string, err error) string {
    var ctxErr *session.GitContextError
    if errors.As(err, &ctxErr) {
        ...
    }
    // fallback for non-GitContextError
    return fmt.Sprintf(`%s git context lookup failed: source=%q cause=%q ...`, ...)
}
```

All paths through `inspectGitContextForSession` now always return `*GitContextError`, so this fallback appears to be dead code. Is it kept intentionally as a defensive contract boundary (in case a future code path wraps the error), or can it be removed? If intentional, a short comment would help future maintainers.

**[Minor question] Cause fallback inconsistency between local and SSH unclassified errors**

In `runLocalGitContextCommand` (`internal/api/handler.go`), when `ClassifyGitFailureCause` returns `GitContextCauseUnknown`:

```go
if cause == session.GitContextCauseUnknown {
    cause = session.GitContextCauseGitMetadata  // promoted to metadata
}
```

In `classifyRemoteGitContextError` (`internal/session/ssh.go`), when the marker is found but classification still returns unknown, the cause stays `GitContextCauseUnknown`. The two paths diverge for the same classification outcome. Is this intentional? (A plausible reason: locally we can also check `exec.ErrNotFound` and there's no equivalent for SSH, so SSH can't safely promote; but a comment noting the reasoning would clarify.)

---

### Overall

Solid improvement. The structured error type and SSH sentinel protocol are well-designed and the log-output tests give real confidence in the diagnostic messages. The three items above are non-blocking; the parameter naming nit in `localGitOptionalMetadata` is the easiest fix before merge.

tomo-chan:
Addressed in https://github.com/tomo-chan/panemux/commit/bcce29dc7052a532cd3636d0a2698c6f6052b83d.

- renamed `localGitOptionalMetadata` parameter from `cwd` to `safeCWD` so it matches the sanitized caller value
- added a short comment explaining why `formatGitContextLookupError` keeps the non-`GitContextError` fallback path as a defensive safety net
- added a short comment explaining why the SSH path intentionally keeps unclassified failures as `unknown` instead of promoting them to metadata errors

Local verification:
- `make lint-go`
- `make test-go`

tomo-chan (commented):
## Code Review — Follow-up

All three minor questions from the previous round have been addressed in `bcce29d2`:

- **Parameter naming** — `localGitOptionalMetadata` parameter renamed from `cwd` to `safeCWD`. Matches the sanitized caller value and eliminates the ambiguity.
- **Defensive fallback in `formatGitContextLookupError`** — comment added explaining the fallback is intentional so future wrapping callers don't silently regress to a raw `%v` log.
- **SSH unknown-cause retention** — comment added in `classifyRemoteGitContextError` noting that SSH-side unknowns stay `unknown` because remote metadata failures can't be reliably distinguished from shell or environment issues through stderr text alone.

No new concerns. This looks good to merge.

PR #123, opened 2026-06-14
fix: prefer claude transcript cwd for workdir detection

tomo-chan:
## Summary

- prefer Claude transcript top-level `cwd` when resolving active agent workdir
- keep tracked file paths and `Bash` `cd ... &&` parsing as fallback for older transcripts
- document the updated Claude workdir resolution behavior

## Testing

- `make test-go`

### DIFF
--- a/internal/session/local.go
+++ b/internal/session/local.go
@@ -670,26 +670,35 @@ func readClaudeProjectCWD(path string) (string, error) {
 
 func parseClaudeProjectCWD(data []byte) (string, error) {
 	scanner := bufio.NewScanner(bytes.NewReader(data))
-	var latest claudePathCandidate
+	var latestCWD string
+	var latestPath claudePathCandidate
 	for scanner.Scan() {
-		if candidate := claudeRecordPath(scanner.Bytes()); candidate.Path != "" {
-			latest = candidate
+		cwd, candidate := claudeRecordPath(scanner.Bytes())
+		if cwd != "" {
+			latestCWD = cwd
+		}
+		if candidate.Path != "" {
+			latestPath = candidate
 		}
 	}
 	if err := scanner.Err(); err != nil {
 		return "", fmt.Errorf("scan claude transcript: %w", err)
 	}
-	if latest.Path == "" {
+	if latestCWD != "" {
+		return latestCWD, nil
+	}
+	if latestPath.Path == "" {
 		return "", nil
 	}
-	if latest.IsDir {
-		return latest.Path, nil
+	if latestPath.IsDir {
+		return latestPath.Path, nil
 	}
-	return filepath.Dir(latest.Path), nil
+	return filepath.Dir(latestPath.Path), nil
 }
 
 type claudeProjectRecord struct {
 	Type     string `json:"type"`
+	CWD      string `json:"cwd"`
 	Snapshot struct {
 		TrackedFileBackups map[string]json.RawMessage `json:"trackedFileBackups"`
 	} `json:"snapshot"`
@@ -703,19 +712,19 @@ type claudePathCandidate struct {
 	IsDir bool
 }
 
-func claudeRecordPath(line []byte) claudePathCandidate {
+func claudeRecordPath(line []byte) (string, claudePathCandidate) {
 	var rec claudeProjectRecord
 	if err := json.Unmarshal(line, &rec); err != nil {
-		return claudePathCandidate{}
+		return "", claudePathCandidate{}
 	}
 
 	switch rec.Type {
 	case "file-history-snapshot":
-		return claudePathCandidate{Path: latestClaudeTrackedPath(rec.Snapshot.TrackedFileBackups)}
+		return rec.CWD, claudePathCandidate{Path: latestClaudeTrackedPath(rec.Snapshot.TrackedFileBackups)}
 	case "assistant":
-		return latestClaudeToolPath(rec.Message.Content)
+		return rec.CWD, latestClaudeToolPath(rec.Message.Content)
 	default:
-		return claudePathCandidate{}
+		return rec.CWD, claudePathCandidate{}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


internal/session/local.go
```diff
@@ -703,19 +712,19 @@ type claudePathCandidate struct {
...
 	if err := json.Unmarshal(line, &rec); err != nil {
-		return claudePathCandidate{}
+		return "", claudePathCandidate{}
 	}
 
 	switch rec.Type {
 	case "file-history-snapshot":
-		return claudePathCandidate{Path: latestClaudeTrackedPath(rec.Snapshot.TrackedFileBackups)}
+		return rec.CWD, claudePathCandidate{Path: latestClaudeTrackedPath(rec.Snapshot.TrackedFileBackups)}
 	case "assistant":
```
tomo-chan, line 724: Does `file-history-snapshot` actually carry a top-level `cwd` key in practice? If not, `rec.CWD` is always `""` here and the return is a no-op. If it does, snapshot records become an undocumented source for `latestCWD` without a test. Worth a comment either way.

tomo-chan (commented):
## Code Review

### What's Good

- **Clean dual-tracking design.** Splitting a single `latest claudePathCandidate` into `latestCWD` and `latestPath` makes the priority order explicit in the loop body rather than hidden in return-value handling. The \"latest wins\" semantics for each slot are easy to reason about.
- **`default` case carries CWD for free.** `user`, `system`, `attachment`, and any future record type all fall through to `default: return rec.CWD, claudePathCandidate{}`. No explicit case list to maintain as the transcript schema evolves.
- **Test fixture simplification.** Removing `os.MkdirAll` / `os.WriteFile` from the integration tests (`TestGetActiveWorkdir_PrefersClaudeTranscriptWorktree`, etc.) is a nice consequence of returning `cwd` directly from JSON rather than resolving a real file path.
- **Doc update is precise.** The four-tier fallback chain (top-level `cwd` → tool file path → `Bash cd` → snapshot) and the no-hooks note are exactly what future maintainers will want when deciding whether to add a fifth tier.

---

### Issues & Questions

**[Question] Does `file-history-snapshot` carry a top-level `cwd` field?**

```go
// internal/session/local.go
case "file-history-snapshot":
    return rec.CWD, claudePathCandidate{Path: latestClaudeTrackedPath(rec.Snapshot.TrackedFileBackups)}
```

The `file-history-snapshot` case now returns `rec.CWD` as its first value. If snapshot records in the Claude transcript format never include a top-level `cwd` key, the returned string is always `""` and this is harmless. But if snapshot records do carry `cwd`, they become an undocumented fourth source for `latestCWD` (docs say only `user`, `assistant`, `attachment`, and `system` expose it), and there's no test covering that case. Which is true of the actual schema?

**[Minor question] `attachment` type: confirmed in schema?**

`docs/behavior.md` lists `attachment` as a record type that exposes top-level `cwd`, but none of the new or existing unit tests exercise an `attachment` entry. Is `attachment` confirmed to carry `cwd` in the current Claude transcript format? If so, a one-line entry in `TestParseClaudeProjectCWD_PrefersLatestTopLevelCWD` would make the coverage explicit.

---

### Overall

Straightforward, well-tested improvement. The two questions above are low-risk to merge without answering — the worst case is harmless dead code or a minor docs/test gap — but an answer would make the `file-history-snapshot` case's intent clearer.

tomo-chan:
Follow-up pushed in 2d043b5: https://github.com/tomo-chan/panemux/commit/2d043b5d693720f3aac7d09ecccb2c0816998093

- Added coverage to `TestParseClaudeProjectCWD_PrefersLatestTopLevelCWD` for an `attachment` record carrying top-level `cwd`.
- Re-checked the observed Claude transcript shapes here: `attachment` records do carry top-level `cwd`, while the `file-history-snapshot` records I inspected do not.
- The snapshot branch still checks `rec.CWD` only as a harmless future-compatible fallback; current behavior still comes from top-level `cwd` on regular transcript records and the existing path-based fallbacks.

`make check` passed after the follow-up commit, and PR checks are green.

tomo-chan (commented):
## Code Review (follow-up — Claude transcript spec verified)

### Prior concerns resolved

Both findings from the earlier review are addressed:

- **`file-history-snapshot` and `cwd`**: Author confirmed snapshot records do not carry a top-level `cwd` field. I verified this by scanning the actual Claude transcript files on disk (`~/.claude/projects/`): `file-history-snapshot` consistently appears in the \"WITHOUT cwd\" group across all sessions inspected. The `return rec.CWD, ...` in that branch is a harmless forward-compatibility fallback, as the author explained.

- **`attachment` type**: Author confirmed `attachment` records do carry `cwd`, and added coverage in commit `2d043b5`. Verified against real transcripts: `attachment` is in the \"WITH cwd\" group alongside `user`, `assistant`, and `system`.

### Transcript schema — what the data shows

From scanning ~30 real Claude session transcripts under `~/.claude/projects/`:

| Record type | Carries `cwd`? |
|---|---|
| `user` | ✅ yes |
| `assistant` | ✅ yes |
| `attachment` | ✅ yes |
| `system` | ✅ yes |
| `file-history-snapshot` | ❌ no |
| `last-prompt` | ❌ no |
| `mode` | ❌ no |
| `permission-mode` | ❌ no |
| `ai-title` | ❌ no |
| `agent-name` | ❌ no |
| `pr-link` | ❌ no |
| `queue-operation` | ❌ no |

The four types documented in `behavior.md` are the complete set that actually carry `cwd`. All other observed types fall through to `default: return rec.CWD, claudePathCandidate{}`, where `rec.CWD` is always `""` — no unintended CWD sources.

### New commit `2d043b5` (`test: cover claude attachment cwd records`)

The updated `TestParseClaudeProjectCWD_PrefersLatestTopLevelCWD` now covers all four documented types in sequence (`user` → `assistant` → `attachment` → `system`) and asserts the final `system` entry wins — correct. No issues.

---

### Overall

Implementation and tests are correct and the spec has been verified against real transcript data. Ready to merge.

PR #125, opened 2026-06-15
fix: prioritize browser copy for tmux panes

tomo-chan:
## Summary
- prioritize browser/xterm selection for `tmux` and `ssh_tmux` panes so plain drag selection plus `Cmd/Ctrl+C` matches `local` and `ssh`
- pass pane type into `useTerminal` and install a narrow tmux-only xterm selection shim while mouse reporting is active
- document the copy-first tradeoff and add frontend coverage for the tmux selection path

## Test Plan
- [x] `make check`
- [x] `make test-e2e`

### DIFF
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -406,6 +406,8 @@ recreating the session and reconnecting the pane. Deliberate remote shell termin
 - Terminal text can be selected with the mouse using xterm.js standard selection behavior.
 - If text is selected, `Cmd+C` or `Ctrl+C` copies the current selection instead of sending terminal input.
 - If no text is selected, `Cmd+C` or `Ctrl+C` is left to normal terminal behavior, so shell interrupts still work.
+- `tmux` and `ssh_tmux` panes intentionally prioritize browser/xterm text selection over tmux-native mouse selection so drag-copy works the same way as `local` and `ssh`.
+- As a result, tmux-backed panes may lose tmux-native drag selection and some mouse-driven TUI interactions.
 - This interaction is currently validated in Chrome.
 
 ### Resize and layout updates
--- a/docs/ui-design.md
+++ b/docs/ui-design.md
@@ -110,6 +110,11 @@ The handle uses `#4a7ea5` and a `grab` cursor. While dragging it switches to `gr
 
 The terminal body must stay available for text selection, focus, mouse reporting, and shell input. Restricting drag initiation to the header avoids gesture conflicts that would otherwise make terminal interaction unreliable.
 
+For `tmux` and `ssh_tmux` panes, panemux currently chooses copy-first behavior inside the pane body:
+plain mouse drag is reserved for browser text selection rather than tmux-native mouse selection. This
+keeps copy interactions consistent across all pane types, even though some tmux mouse interactions are
+deprioritized.
+
 ---
 
 ## Drag and Drop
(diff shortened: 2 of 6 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Clean separation of concerns**: `syncSelectionMode` isolates the tmux-specific behavior as a standalone function without touching `getOrCreateTerminalEntry`, making the default path unchanged and easy to reason about.
- **Correct capture-phase interception**: Using `addEventListener('mousedown', handler, true)` with `stopImmediatePropagation` is exactly the right pattern to win against xterm's own bubbling-phase listeners.
- **Leak-safe teardown**: Storing `selectionModeCleanup` on `TerminalEntry` and calling it both from `syncSelectionMode` (on re-attach) and from the dispose path ensures the listener is always removed before the element is destroyed.
- **Accurate documentation**: Both `behavior.md` and `ui-design.md` now capture the trade-off explicitly — this is exactly the kind of content that should live in `docs/` rather than in commit messages.
- **Test refactoring**: The `renderUseTerminal` helper removes significant repetition from the test file without obscuring intent.
- **Behavioral coverage**: The three new tests for tmux-active, local, and tmux-with-mouse-inactive cover the key branches in `installTmuxCopyFirstSelection`.

---

### Issues & Questions

**[Question] Private xterm.js API dependency**

`installTmuxCopyFirstSelection` accesses `term._core._selectionService.handleMouseDown` and `term._core.coreMouseService.areMouseEventsActive` — both underscore-prefixed internal properties not present in xterm.js's public TypeScript declarations:

```typescript
// frontend/src/hooks/useTerminal.ts
const core = (term as Terminal & {
  _core?: {
    _selectionService?: { handleMouseDown: (event: MouseEvent) => void }
    coreMouseService?: { areMouseEventsActive?: boolean }
  }
})._core
```

The optional chaining (`?.`) means a graceful no-op if these disappear, but it also means the feature silently stops working on an xterm upgrade without any compile-time or runtime error. Is there a plan to pin the xterm.js version until a public API alternative is available, or to add a runtime check (e.g., a `warn` in dev builds) so breakage is detectable?

**[Minor question] Silent no-op when `term.element` is null**

```typescript
function installTmuxCopyFirstSelection(term: Terminal): () => void {
  const element = term.element
  if (\!element) return () => {}
  ...
}
```

`syncSelectionMode` is called immediately after `attachTerminal`, and `attachTerminal` calls `term.open(container)` if the element isn't set yet — so `term.element` should always be non-null at this call site. Is the early return here an intentional defensive guard, or a signal that there's a case where `installTmuxCopyFirstSelection` might be called before `open`? If it's just defensive, a comment or assertion would help future readers understand why the null case is handled.

**[Minor question] No explicit test for `ssh_tmux` type**

The three new behavior tests cover `'tmux'`, `'local'`, and `'tmux'` with `areMouseEventsActive = false`. `'ssh_tmux'` is handled identically by `syncSelectionMode` (same `if` branch), but there's no test that explicitly documents and verifies that behavior. Given that the PR title and documentation both call out `ssh_tmux` by name, a fourth test case would make that coverage explicit and prevent a future regression if the two types are ever differentiated.

---

### Overall

The approach is sound and the implementation is careful. The two questions above are about the sustainability of the private-API dependency and minor coverage gaps — neither blocks this change from being useful and correct as written. Worth confirming the xterm version pinning question before the next xterm upgrade.

PR #126, opened 2026-06-15
fix: add cross-platform tmux selection modifier

tomo-chan:
## Summary

- enable xterm forced selection support for tmux-backed panes
- document the platform-specific modifier for browser-side selection
- keep tmux mouse mode unchanged while preserving a reliable copy path

## Why

tmux mouse mode intercepts plain drag selection in `tmux` and `ssh_tmux` panes. This keeps tmux mouse behavior intact while exposing the browser-side forced-selection path that xterm already supports.

## Impact

- plain drag in `tmux` and `ssh_tmux` panes still follows tmux mouse behavior
- macOS users can hold `Option` while dragging to force browser-side selection
- Linux and Windows users can hold `Shift` while dragging to force browser-side selection
- no backend, API, or config changes

## Validation

- [x] `make test-frontend`
- [x] `make check`

### DIFF
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -155,6 +155,11 @@ Remote Git inspection currently depends on `git rev-parse --path-format=absolute
 on the SSH target, which requires Git 2.31 or newer. Older remote Git versions degrade to "no git
 info" in the pane header for SSH-backed panes.
 
+For terminal text selection, panemux keeps tmux mouse mode unchanged for `tmux` and `ssh_tmux`
+panes. Plain drag therefore continues to follow tmux mouse behavior, while macOS users can hold
+`Option` during drag to force browser-side xterm selection and copy text without changing tmux
+session settings.
+
 ## REST API
 
 ### `GET /api/layout`
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -156,9 +156,9 @@ on the SSH target, which requires Git 2.31 or newer. Older remote Git versions d
 info" in the pane header for SSH-backed panes.
 
 For terminal text selection, panemux keeps tmux mouse mode unchanged for `tmux` and `ssh_tmux`
-panes. Plain drag therefore continues to follow tmux mouse behavior, while macOS users can hold
-`Option` during drag to force browser-side xterm selection and copy text without changing tmux
-session settings.
+panes. Plain drag therefore continues to follow tmux mouse behavior. To force browser-side xterm
+selection while tmux mouse mode is active, hold `Option` during drag on macOS or `Shift` during
+drag on Linux and Windows.
 
 ## REST API
 
--- a/frontend/src/hooks/useTerminal.test.ts
+++ b/frontend/src/hooks/useTerminal.test.ts
@@ -139,6 +139,16 @@ describe('useTerminal', () => {
     expect(TERMINAL_FONT_FAMILY).toContain('Hiragino Sans')
   })
 
+  it('enables Option-drag selection when terminal apps turn on mouse mode', () => {
+    const container = makeContainer()
+
+    renderHook(() => useTerminal({ sessionId: 's1', container }))
+
+    expect(mockTerminalCtor).toHaveBeenCalledWith(expect.objectContaining({
+      macOptionClickForcesSelection: true,
+    }))
+  })
+
   it('loads all addons on init', () => {
     const container = makeContainer()
     renderHook(() => useTerminal({ sessionId: 's1', container }))
--- a/frontend/src/hooks/useTerminal.test.ts
+++ b/frontend/src/hooks/useTerminal.test.ts
@@ -139,7 +139,7 @@ describe('useTerminal', () => {
     expect(TERMINAL_FONT_FAMILY).toContain('Hiragino Sans')
   })
 
-  it('enables Option-drag selection when terminal apps turn on mouse mode', () => {
+  it('enables forced selection for tmux mouse mode across platforms', () => {
     const container = makeContainer()
 
     renderHook(() => useTerminal({ sessionId: 's1', container }))
--- a/frontend/src/hooks/useTerminal.ts
+++ b/frontend/src/hooks/useTerminal.ts
@@ -279,6 +279,7 @@ function getOrCreateTerminalEntry(sessionId: string): TerminalEntry {
     customGlyphs: true,
     fontSize: 14,
     fontFamily: TERMINAL_FONT_FAMILY,
+    macOptionClickForcesSelection: true,
     // Hide xterm.js accessibility textarea (still exists for IME/a11y but invisible)
     screenReaderMode: false,
     theme: {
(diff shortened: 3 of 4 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- Minimal, well-targeted change: one option, one test, one docs paragraph — no extraneous scope.
- The test correctly locks in the xterm.js constructor option with `expect.objectContaining`, so a future accidental removal would be caught immediately.
- Documentation is updated in the same commit, consistent with the project's rule that behavior changes must be documented in the same change.
- `macOptionClickForcesSelection: true` is the correct xterm.js option for this use case — verified against current xtermjs.org docs.

---

### Issues & Questions

**[Minor question] Test name implies conditional behavior**

The test is named `'enables Option-drag selection when terminal apps turn on mouse mode'`. The phrase "when terminal apps turn on mouse mode" implies the option is conditionally applied, but `macOptionClickForcesSelection` is set at Terminal construction time and is always `true`. The clause describes the motivation, not the behavior, which may be confusing for a reader unfamiliar with the background.

Would `'passes macOptionClickForcesSelection: true to Terminal constructor'` or `'enables Option-drag selection for mouse-capturing terminal apps'` be clearer, or is the current phrasing intentional to document the use-case context?

**[Minor question] Non-macOS selection path undocumented**

The docs paragraph says "macOS users can hold `Option` during drag to force browser-side xterm selection" — correctly scoped to macOS (the option itself is macOS-specific). However, it leaves Windows/Linux users without any documented copy path for tmux-backed panes where mouse mode is active. Is documenting non-macOS workarounds (e.g., Shift+drag, which bypasses captured mouse events in most browsers) in scope, or intentionally deferred?

---

### Overall

Clean, focused PR. Both questions are minor and non-blocking. The implementation is correct and the test coverage is appropriate for a constructor-option change.

frontend/src/hooks/useTerminal.test.ts
```diff
@@ -139,6 +139,16 @@ describe('useTerminal', () => {
     expect(TERMINAL_FONT_FAMILY).toContain('Hiragino Sans')
   })
 
+  it('enables Option-drag selection when terminal apps turn on mouse mode', () => {
+    const container = makeContainer()
```
tomo-chan, line 143: The "when terminal apps turn on mouse mode" clause implies conditional behavior, but `macOptionClickForcesSelection` is always `true` at construction time — it describes the motivation, not a condition. Consider `'passes macOptionClickForcesSelection: true to Terminal constructor'` or `'enables Option-drag selection for mouse-capturing terminal apps'`.

docs/behavior.md
```diff
@@ -155,6 +155,11 @@ Remote Git inspection currently depends on `git rev-parse --path-format=absolute
 on the SSH target, which requires Git 2.31 or newer. Older remote Git versions degrade to "no git
 info" in the pane header for SSH-backed panes.
 
+For terminal text selection, panemux keeps tmux mouse mode unchanged for `tmux` and `ssh_tmux`
+panes. Plain drag therefore continues to follow tmux mouse behavior, while macOS users can hold
+`Option` during drag to force browser-side xterm selection and copy text without changing tmux
+session settings.
```
tomo-chan: This covers the macOS path but leaves Windows/Linux users without a documented copy path for tmux-backed panes with mouse mode active (e.g. Shift+drag bypasses captured mouse events in most browsers). Worth noting here if cross-platform copy UX is in scope, or marking explicitly out-of-scope if macOS is the only target for this feature.

frontend/src/hooks/useTerminal.ts
```diff
@@ -279,6 +279,9 @@ function getOrCreateTerminalEntry(sessionId: string): TerminalEntry {
     customGlyphs: true,
     fontSize: 14,
     fontFamily: TERMINAL_FONT_FAMILY,
+    // xterm uses Shift-drag for forced selection on non-macOS and can be
+    // configured to use Option-drag on macOS while terminal mouse mode is active.
```
tomo-chan: xterm.js's `shouldForceSelection` returns `event.shiftKey` unconditionally, so Shift+drag works on macOS too. "non-macOS" may mislead future maintainers. Consider: `"xterm supports Shift-drag for forced selection by default; this option additionally enables Option-drag on macOS while terminal mouse mode is active."`

tomo-chan (commented):
## Code Review (follow-up)

### Previous concerns — both resolved

- **Test name implies conditional behavior** — renamed to `'enables forced selection for tmux mouse mode across platforms'`. Clear improvement.
- **Non-macOS selection path undocumented** — docs now explicitly cover both `Option` (macOS) and `Shift` (Linux/Windows). Verified Shift+drag is default xterm.js behavior (`shouldForceSelection` returns `event.shiftKey` unconditionally) — no additional configuration needed, so the docs claim is accurate.

---

### Issues & Questions

**[Minor question] Code comment says "non-macOS" but Shift+drag works on all platforms**

`useTerminal.ts` lines 280-281:
```ts
// xterm uses Shift-drag for forced selection on non-macOS and can be
// configured to use Option-drag on macOS while terminal mouse mode is active.
```

The xterm.js `SelectionService` checks `event.shiftKey` in `shouldForceSelection` regardless of OS — Shift+drag bypasses mouse tracking mode on macOS too. The comment is accurate as a description of what we added (the macOS Option path), but "on non-macOS" may mislead a future maintainer. Something like:

```ts
// xterm supports Shift-drag for forced selection by default; this option
// additionally enables Option-drag on macOS while terminal mouse mode is active.
```

---

### Overall

Both prior concerns are addressed. Only a minor comment wording question remains — non-blocking.

frontend/src/hooks/useTerminal.ts
tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/59d1a2be0cf08e4d2ad1750274db59125d4db5d1 .

I updated the code comment to reflect xterm's actual default behavior: Shift-drag remains the built-in forced-selection path across platforms, and `macOptionClickForcesSelection` additionally enables Option-drag on macOS.

Re-verified locally with:
- `make test-frontend`

tomo-chan (commented):
## Code Review (follow-up #2)

### Previous concern — resolved

**[Minor question] Code comment says "non-macOS" but Shift+drag works on all platforms**

Addressed in commit `59d1a2b8` with the exact wording suggested:

```ts
// xterm supports Shift-drag for forced selection by default; this option
// additionally enables Option-drag on macOS while terminal mouse mode is active.
```

Thank you for the commit reply confirming the rationale.

---

### What's Good (new commits)

- `README.md` now surfaces the tmux selection key modifiers as a user-facing bullet in the Common Uses section — good placement alongside the other tmux-specific notes, and consistent with `docs/behavior.md`.

---

### Overall

All concerns from prior reviews are resolved. No new findings. Ready to merge.

PR #127, opened 2026-06-15
fix: add cross-platform tmux selection modifier

tomo-chan:
## Summary

- enable xterm forced selection support for tmux-backed panes
- document the platform-specific modifier for browser-side selection
- keep tmux mouse mode unchanged while preserving a reliable copy path

## Why

tmux mouse mode was still intercepting plain drag selection. The previous change only documented the macOS `Option` path; this update makes the behavior explicit across platforms by documenting the matching non-macOS modifier path as well.

## Impact

- plain drag in `tmux` and `ssh_tmux` panes still follows tmux mouse behavior
- macOS users can hold `Option` while dragging to force browser-side selection
- Linux and Windows users can hold `Shift` while dragging to force browser-side selection
- no backend, API, or config changes

## Validation

- [x] `make test-frontend`
- [x] `make check`

### DIFF
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -155,6 +155,11 @@ Remote Git inspection currently depends on `git rev-parse --path-format=absolute
 on the SSH target, which requires Git 2.31 or newer. Older remote Git versions degrade to "no git
 info" in the pane header for SSH-backed panes.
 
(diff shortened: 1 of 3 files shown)


tomo-chan:
Superseded by #126. The follow-up change was moved onto the original PR so the tmux selection work stays in a single review thread.

PR #130, opened 2026-06-16
fix: handle large interactive agent session logs

tomo-chan:
## Summary

- replace Scanner-based interactive agent session JSONL parsing with a size-safe line iterator
- apply the fix to both Codex session logs and Claude transcript parsing so large single-line records no longer fail workdir detection
- add regression tests covering records larger than 64 KiB for both parsers

## Root cause

Both parsers used bufio.Scanner, which aborts on tokens larger than its default limit and surfaced as token too long errors during interactive agent workdir lookup.

## Validation

- make build-frontend
- make test-go
- make lint-go

### DIFF
--- a/internal/session/local.go
+++ b/internal/session/local.go
@@ -1,7 +1,6 @@
 package session
 
 import (
-	"bufio"
 	"bytes"
 	"encoding/json"
 	"errors"
@@ -566,15 +565,13 @@ type codexSessionRecord struct {
 }
 
 func parseCodexSessionCWD(data []byte) (string, error) {
-	scanner := bufio.NewScanner(bytes.NewReader(data))
-
 	var latestTurnCWD string
 	var sessionMetaCWD string
 	var latestExecWorkdir string
-	for scanner.Scan() {
+	forEachJSONLRecord(data, func(line []byte) {
 		var rec codexSessionRecord
-		if err := json.Unmarshal(scanner.Bytes(), &rec); err != nil {
-			continue
+		if err := json.Unmarshal(line, &rec); err != nil {
+			return
 		}
 		execWorkdir, turnCWD, metaCWD := codexSessionRecordCWD(rec)
 		if execWorkdir != "" {
@@ -586,10 +583,7 @@ func parseCodexSessionCWD(data []byte) (string, error) {
 		if metaCWD != "" && sessionMetaCWD == "" {
 			sessionMetaCWD = metaCWD
 		}
-	}
-	if err := scanner.Err(); err != nil {
-		return "", fmt.Errorf("scan codex session log: %w", err)
-	}
+	})
 	// Codex may keep both process cwd and turn/session metadata pinned to the
 	// original pane directory while tool calls run inside a sibling worktree.
 	// The latest exec_command workdir is therefore the strongest signal.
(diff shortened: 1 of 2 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **`forEachJSONLRecord` abstraction** (`internal/session/local.go`): the helper is clean and reusable across both parsers. Handling CRLF (`TrimSuffix(\r)`) and skipping blank lines is defensive and correct.
- **No memory overhead**: since the callers already hold the full file as `[]byte`, the iterator works in-place with no additional allocation beyond what is already in memory.
- **Regression tests**: both `TestParseCodexSessionCWD_LargeExecCommandRecord` and `TestParseClaudeProjectCWD_LargeTopLevelCWDRecord` exercise records above the 64 KiB Scanner default, directly covering the failure mode described in the root-cause section.
- **Right-sized fix**: the change targets only the two affected parsers and does not widen scope unnecessarily.

---

### Issues & Questions

**[Question] Removal of `scanner.Err()` propagation — is the error set complete?**

Both `parseCodexSessionCWD` and `parseClaudeProjectCWD` previously returned an error when the Scanner hit `bufio.ErrTooLong`:

```go
if err := scanner.Err(); err \!= nil {
    return "", fmt.Errorf("scan codex session log: %w", err)
}
```

`forEachJSONLRecord` is intentionally infallible — it operates on an already-in-memory `[]byte` via `bytes.IndexByte`, which cannot fail. Since the underlying reader was always `bytes.NewReader(data)`, the only non-nil value `scanner.Err()` could ever return in practice was `bufio.ErrTooLong`, which this PR eliminates by design.

Is that assumption correct — i.e., `bufio.ErrTooLong` was the only reachable scanner error for these callers? If so, a short comment on `forEachJSONLRecord` (e.g., "operates on in-memory data and cannot error") would help future readers understand why there is no error return, and would make the intentional behavioral change explicit.

**[Minor question] `else { data = nil }` branch is not exercised by the new tests**

The `else` branch in `forEachJSONLRecord` correctly handles JSONL files that end without a trailing `\n`. Both new test fixtures include a trailing newline, so this branch is never reached. A one-line variant of either test (strip the final `\n`) would close that coverage gap and confirm the branch behaves as intended.

---

### Overall

Clean, well-scoped fix. The behavioral change — large-record error → successful CWD extraction — is clearly intentional and matches the PR description. The question above about error-removal rationale is the only thing I'd want clarified before merge; the minor question is non-blocking.

tomo-chan:
Addressed the review follow-up in commit https://github.com/tomo-chan/panemux/commit/7b4247fbb02fc6a2cf6219010d6a1ca24f5a120f.

- Added a comment on `forEachJSONLRecord` to make the intentional removal of Scanner error propagation explicit: it operates on already-buffered in-memory JSONL and avoids the Scanner token limit.
- Added a regression test for a large Claude transcript record without a trailing newline to cover the final-record branch.

Local verification rerun:
- `make lint-go`

tomo-chan (commented):
## Code Review — Follow-up

Both concerns from the previous review have been resolved in commit `7b4247f0`.

**[Question] resolved** — `forEachJSONLRecord` now carries a clear doc comment:

> iterates over already-buffered JSONL data without using bufio.Scanner, so it cannot surface Scanner's token-size limit. It also preserves a final unterminated record.

This makes the intentional removal of `scanner.Err()` propagation self-documenting for future readers.

**[Minor question] resolved** — `TestParseClaudeProjectCWD_LargeTopLevelCWDRecord_WithoutTrailingNewline` covers the `else { data = nil }` branch with a 70 KiB fixture, closing the coverage gap.

---

### Overall

No remaining concerns. The PR is clean and ready to merge.

PR #133, opened 2026-06-17
fix: reduce git-info polling and session log reads

tomo-chan:
## Summary

- reduce pane git-info polling to visible panes only and pause polling while the browser tab is hidden
- cache Codex/Claude session-log workdir resolution so unchanged local and remote jsonl files are not re-read
- document the new polling and session-log fingerprint behavior

## Test Plan

- [x] make build-frontend
- [x] make test-frontend
- [x] make test-go
- [x] make check

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -179,6 +179,11 @@ cross-repository reuse.
 
 For interactive Codex sessions, all four session implementations (`local`, `ssh`, `tmux`, and `ssh_tmux`) may inspect the open Codex session log under `~/.codex/sessions/...jsonl` when the Codex process has that file open. Panemux currently prefers the latest `exec_command.arguments.workdir` recorded in `response_item` / `function_call` log entries, then falls back to `turn_context.cwd`, then `session_meta.cwd`.
 
+To avoid repeatedly reparsing unchanged agent logs, panemux caches the last resolved workdir per
+session-log path fingerprint. Local sessions use file size plus modtime; SSH-backed sessions ask
+the remote host for lightweight file metadata first and only transfer the full `jsonl` contents
+again when that fingerprint changes.
+
 This ordering is intentional and reflects observed Codex behavior as of `codex-tui` `0.130.0`:
 
 - the interactive Codex process may keep its OS-level process cwd at the original pane directory
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -145,11 +145,23 @@ For pane-header Git status, local and local tmux panes inspect the local filesys
 and `ssh_tmux` panes run the equivalent Git inspection on the remote host. This allows headers to
 show branch/repository info even when the repository exists only on the SSH target.
 
+Pane-header Git and PR metadata is fetched immediately when a pane becomes visible, then refreshed
+every 10 seconds only while both the browser tab and that pane remain visible.
+
+When a pane is hidden behind maximize, its header Git/PR polling stops until it becomes visible
+again. Restoring the browser tab or making a pane visible again triggers an immediate refresh so
+newly created PR links become clickable without waiting for the next steady-state poll.
+
 When panemux detects that an interactive `codex` or `claude` session is operating in a sibling Git
 worktree for the same repository, pane-header Git/PR metadata and the VS Code open action prefer
 that worktree. After the agent exits, panemux keeps the last valid sibling worktree pinned for that
 pane session until a newer valid sibling worktree is detected or the pane itself changes to a
 different repository.
+
+To avoid repeatedly transferring and parsing unchanged interactive-agent logs, panemux reuses the
+last parsed Codex or Claude worktree result while the underlying session-log or transcript file
+fingerprint remains unchanged. For SSH-backed panes this check uses remote file metadata first and
+only reads the full `jsonl` contents again when the fingerprint changes.
 
 Remote Git inspection currently depends on `git rev-parse --path-format=absolute --git-common-dir`
 on the SSH target, which requires Git 2.31 or newer. Older remote Git versions degrade to "no git
--- a/frontend/src/components/TerminalPane.test.tsx
+++ b/frontend/src/components/TerminalPane.test.tsx
@@ -12,12 +12,15 @@ vi.stubGlobal('ResizeObserver', ResizeObserverMock)
 const { mockUseTerminal } = vi.hoisted(() => ({
   mockUseTerminal: vi.fn(),
 }))
+const { mockUseGitInfo } = vi.hoisted(() => ({
+  mockUseGitInfo: vi.fn(),
+}))
 
 vi.mock('../hooks/useTerminal', () => ({
   useTerminal: mockUseTerminal,
 }))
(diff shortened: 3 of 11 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Clean fingerprint abstraction** — `agentLogFingerprint` as a typed alias for `string` gives the cache key distinct types at a glance; `agentLogCacheEntry` and `cachedAgentLogCWD` in their own file keep the concern isolated.
- **Proper mutex discipline** — The cache always uses `Lock/Unlock` rather than `RLock/RUnlock`, which avoids the write-after-check race that commonly appears in two-phase cache patterns. The approach is intentionally conservative and correct.
- **`remoteFileFingerprintCmd` cross-platform coverage** — Cascading through GNU `stat -c`, BSD `stat -f`, and finally `wc -c` handles Linux, macOS, and minimal container environments in a single shell snippet.
- **`enabled` + `isVisible` composition** — Gating `fetchGitInfo` on both the maximize state and tab visibility means polling stops exactly when it is useless (hidden pane or background tab) and restarts immediately when either condition reverses. The separate React effects keep each concern testable in isolation.
- **`vi.useFakeTimers()` added to the interval test** — The prior test was asserting on a `setInterval` call site without advancing time, which could give a false green under some configurations; using fake timers makes the assertion deterministic.
- **`statFileFn` injectable override** — Matching the existing `readFileFn` pattern keeps fingerprint tests fully in-process without touching the real filesystem.

---

### Issues & Questions

**[Minor question] `remoteClaudeProjectReadCmd` may be dead code after this PR**

The PR renames the original `remoteClaudeProjectReadCmd` to `remoteClaudeProjectShellPath` and then re-adds `remoteClaudeProjectReadCmd` as a thin wrapper that returns `"cat " + remoteClaudeProjectShellPath(meta)`. However, the refactored `remoteClaudeSessionCWD` now calls `remoteCachedAgentLogCWD(... remoteClaudeProjectShellPath(meta) ...)` directly, so `remoteClaudeProjectReadCmd` no longer appears to be called anywhere. The new SSH cache tests also compute the expected `cat` command manually rather than calling this function.

The `unused` linter (enabled in `.golangci.yml`) should catch this if the function is truly unreachable. Did `make lint-go` pass? If the linter does not flag it, is the function being kept intentionally (e.g., for future callers)?

Relevant location: `internal/session/ssh.go`, `remoteClaudeProjectReadCmd`.

---

**[Minor question] Process-level cache has no eviction path**

`agentLogCache` is a package-level `map[string]agentLogCacheEntry` that grows as new `(cacheKey, fingerprint)` pairs are inserted. Entries are never removed. For a long-running panemux instance that opens and closes many panes pointing at distinct log paths, this accumulates indefinitely.

Is the assumption that the number of distinct cache keys stays small enough in practice that eviction is not needed? If so, a short comment in `agent_log_cache.go` would make the decision explicit for future maintainers.

---

**[Question] `wc -c` fingerprint branch omits modtime**

In `remoteFileFingerprintCmd`, the final fallback produces `"<bytes> -"` (a literal `-` in place of a modification time). If a session log is rewritten in-place with a new entry that leaves the total byte count unchanged — unlikely but theoretically possible — the cache would serve stale data until the byte count changes.

Is this edge case accepted as low-risk given that the `wc -c` branch only fires when both `stat -c` and `stat -f` are unavailable, and log entries are appended (growing the file)? Documenting the assumption (append-only log growth) as a brief inline comment would clarify the intent.

---

**[Minor question] `make lint` absent from test plan**

The project's quality gate (`DEVELOPMENT.md`: "make check must pass before make build") includes linting. The PR test plan lists `make build-frontend`, `make test-frontend`, and `make test-go` but not `make lint` or `make check`. Given the dead-code question above, it would be good to confirm the lint gate passes before merge.

---

### Overall

The caching strategy is well-structured and the frontend visibility changes are correct and well-tested. The main open question is whether `remoteClaudeProjectReadCmd` is intentionally kept or is dead code — confirming that `make lint-go` passes cleanly would resolve it.

tomo-chan:
Follow-up on the review summary:

- Removed the dead `remoteClaudeProjectReadCmd` helper and cleaned up the session-log cache implementation in commit https://github.com/tomo-chan/panemux/commit/3468cdc7a472e43ffab1259b1bba321a0827a395.
- Ran the full repository quality gate with `make check`; that covers frontend build, lint, tests, and coverage.
- The remaining cache behavior is intentional: the fallback fingerprint path uses size-only metadata only when neither GNU nor BSD `stat` is available, and the logs are treated as append-oriented in that degraded path.

tomo-chan (commented):
## Code Review — Follow-up (commit 3468cdc7)

All concerns from the first review have been addressed.

### Resolved

- **`remoteClaudeProjectReadCmd` dead code** — removed; `remoteClaudeProjectShellPath` is the only entry point now. ✓
- **`make lint` / `make check`** — author confirms the full quality gate passed. ✓
- **Cache eviction** — acknowledged as intentionally unbounded; the number of distinct keys is bounded by the number of panes × agent logs a user ever opens in a session. ✓
- **`wc -c` fingerprint with no modtime** — acknowledged; the fallback only fires when both `stat` variants are absent, and append-only log growth makes a same-size overwrite effectively impossible in practice. ✓

### New in 3468cdc7

**`agentLogCacheStore` named type** — extracting the anonymous struct into a named `agentLogCacheStore` is cleaner. One minor nit: the conventional Go field order places the `sync.Mutex` before the fields it guards (matching the pattern in the standard library and the `sync` package docs). The current order has `entries` first and `mu` second. This does not affect correctness.

**Variable naming in `remoteCachedAgentLogCWD`** — renaming `err` to `readErr`/`parseErr` inside both the closure and the fallback path eliminates any ambiguity with the outer `fingerprint, err :=` assignment. Good catch.

---

### Overall

The PR is in good shape. The single nit above (mutex field order) is purely cosmetic and does not block merge.

tomo-chan (commented):
## Code Review — Follow-up (commit 5b6b0c44)

### Addressed from the previous round

The mutex field-order nit from the last review prompted a broader convention pass in `agent_log_cache.go`. The result is better than just reordering fields:

- **Method encapsulation** (`load`, `store`, `reset`) — mutex access is now owned by the struct rather than scattered at call sites. This is the idiomatic Go pattern for struct-level locking and removes all the manual `Lock/Unlock` pairs from `cachedAgentLogCWD`.
- **`readAndParseRemoteAgentLog` helper** — the duplicated read+parse block that existed in both the fingerprint-ok and fallback paths of `remoteCachedAgentLogCWD` is now a single function. Both paths call the same code.

### Remaining cosmetic point

`agentLogCacheStore` still declares `entries` before `mu`. The Go convention (followed by `sync.Map`, `http.Server`, etc.) is to place the mutex first so readers see the lock before the data it protects. Since `mu` is now accessed only through the new methods, the practical impact is zero — this is purely a documentation nit. Leaving it or fixing it are both fine.

### Overall

No blocking issues remain. The PR is clean and ready to merge.

PR #135, opened 2026-06-20
chore: enforce make check before push

tomo-chan:
## Summary

- enforce `make check` as a required local completion gate
- install a tracked repo-local pre-push hook via `make install-deps`
- document the new hook and completion requirements for agents and contributors

## Why

Recent local changes have been reaching CI without a full `make check` pass, which causes avoidable lint and validation failures after push. This change moves that quality gate earlier by making it part of local completion rules and by blocking pushes when `make check` fails.

## Validation

- `make install-deps`
- `make check`

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -3,6 +3,12 @@
 set -eu
 
 repo_root=$(git rev-parse --show-toplevel)
+git_local_env_vars=$(git rev-parse --local-env-vars)
+
+# Git exports repository-scoped environment variables into hooks. Clear them so
+# nested git commands during make check can target their own test repositories.
+# shellcheck disable=SC2086
+unset $git_local_env_vars
 
 echo "pre-push: running make check"
 cd "$repo_root"
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+set -eu
+
+repo_root=$(git rev-parse --show-toplevel)
+
+echo "pre-push: running make check"
+cd "$repo_root"
+make check
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,12 +13,12 @@ Before building or running tests, always run:
 make install-deps
 ```
 
-This installs npm packages and downloads Go modules. Required once per environment.
+This installs npm packages, downloads Go modules, and configures the repo-local Git hooks. Required once per environment.
 
 ## Read This First
 
 1. Read [DEVELOPMENT.md](DEVELOPMENT.md) for build, test, branch, and PR workflow rules.
-   This includes the path-sanitization rule for tests, fixtures, PR bodies, PR comments, and review replies.
+   This includes the path-sanitization rule for tests, fixtures, PR bodies, PR comments, review replies, and the mandatory `make check` completion rule.
 2. Read [docs/overview.md](docs/overview.md) for the product purpose and current boundaries.
 3. Read [docs/architecture.md](docs/architecture.md) for implementation structure and security design.
 4. Read [docs/security.md](docs/security.md) before implementing changes that affect command execution, shell argument handling, SSH path handling, or `gosec` posture.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -114,6 +114,7 @@ Example: `Config.sshConfigPath` uses `sshconfig.DefaultPath()` only when the ove
 
 - `make install-deps` configures the repo-local Git hooks path to `.githooks`.
 - The tracked `pre-push` hook runs `make check` and blocks `git push` when it fails.
+- The hook clears Git's repository-scoped hook environment variables before running `make check` so nested test Git repositories behave the same way they do outside hook execution.
 - Do not bypass the hook for ordinary development. Fix the failing checks instead.
 
 ### Documentation updates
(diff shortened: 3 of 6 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Tracked `.githooks/` directory** — versioning the hook in `.githooks/` rather than `.git/hooks/` is the canonical approach for distributing hooks via a repo. Developers and agents get it automatically on `make install-deps`.
- **`install-deps-ci` deliberately excludes `install-hooks`** — CI runners shouldn't have their git config mutated during a reproducible install. Keeping the two targets separate is the right call.
- **Hook script is well-written** — `set -eu`, a double-quoted variable for `$repo_root`, and dynamic root detection via `git rev-parse --show-toplevel` are all correct defensive patterns.
- **Documentation is consistent** — AGENTS.md, DEVELOPMENT.md, and README.md are all updated in the same commit, so there's no ambiguity about which docs are authoritative.
- **Explicit no-exceptions wording in DEVELOPMENT.md** — `There are no exceptions for frontend-only, docs-adjacent, or small code changes.` directly addresses the documented failure mode (lint/validation failures reaching CI) and will be effective for agents.

---

### Issues & Questions

**[Minor question] `chmod +x` in `install-hooks` is redundant**

```makefile
install-hooks:
	chmod +x .githooks/pre-push
	git config core.hooksPath .githooks
```

The file is committed as mode `100755` (`new file mode 100755` in the diff), so Git preserves the execute bit on clone on POSIX systems. `chmod +x` is a no-op on a fresh clone. Is this intentional as a defensive measure for Windows contributors who may have `core.fileMode=false` configured, or can it be dropped?

---

### Overall

Clean, focused change. The design is correct: tracked hooks, separate CI target, safe hook script, and consistent documentation. One minor clarification on the redundant `chmod +x` would be helpful, but it doesn't block merge.

tomo-chan:
Thanks. `chmod +x` is intentional as a defensive step.

The hook is committed with mode `100755`, so a normal POSIX clone should already preserve executability. I kept `chmod +x .githooks/pre-push` in `install-hooks` to make the setup path resilient when that bit is lost locally or when a contributor has platform/file-mode behavior that leaves the tracked hook non-executable at install time.

There is no functional dependency on it in CI because `install-deps-ci` does not call `install-hooks`.

Current PR validation is green:
- `make install-deps`
- `make check`
- GitHub checks: `Analyze (actions)`, `Analyze (go)`, `Analyze (javascript-typescript)`, `CodeQL`, `ci`, `label`, `validate`

tomo-chan:
Follow-up for the hook flake:

- Fixed in https://github.com/tomo-chan/panemux/commit/af705c306985b6ec8fb3d61e48faf60ff777fc59
- Root cause was Git's repository-scoped hook environment leaking into `make check`, so nested Git commands in temp test repos could target the wrong repository during `pre-push`.
- The hook now clears `git rev-parse --local-env-vars` before running `make check`.

Verified locally:

- `make check`
- `git push` with the tracked `pre-push` hook enabled

GitHub checks are running again on the updated head.

tomo-chan (commented):
## Code Review (follow-up: af705c34)

- **`chmod +x` in `install-hooks`** — Author confirmed this is intentional: defensive against environments where tracked file mode is lost locally or where `core.fileMode=false` strips the execute bit. No action needed.

---

### What's Good

- **Canonical env-var list via `git rev-parse --local-env-vars`** — using Git's own API to enumerate which variables to clear is significantly better than a hardcoded list. If Git adds or removes hook-scope variables in a future version, the hook adapts automatically.
- **`$repo_root` captured before the unset** — the `cd "$repo_root"` line still works correctly because `repo_root` is a plain shell variable, not a Git env var.
- **Comment explains the why** — the block comment describing nested test-repository isolation is exactly the kind of non-obvious constraint that belongs in the script.
- **`# shellcheck disable=SC2086` correctly scoped** — placed on the specific `unset` line, not at the file level, with the reason implicit in the comment above it.
- **DEVELOPMENT.md updated** — the new Push protection bullet documents the env-var clearing so contributors understand the hook's behaviour without reading the script.

---

### Issues & Questions

**[Minor question] Minimum Git version for `--local-env-vars`**

```sh
git_local_env_vars=$(git rev-parse --local-env-vars)
```

The current Git documentation confirms this flag exists, but I could not determine the exact version it was introduced in. If the project has a documented minimum Git version requirement, does `--local-env-vars` fall within it? If not, is there a known floor below which the hook would silently fail (e.g. returning an error into `git_local_env_vars`)?

With `set -eu` active, an unrecognised flag would cause `git rev-parse --local-env-vars` to exit non-zero, which would abort the hook before `make check` runs — a safe failure mode, but it would break pushes for anyone on a very old Git version. Documenting the floor (or noting it's been supported since `git x.y`) would make the tradeoff explicit.

---

### Overall

The env-var isolation fix is the right approach and is implemented cleanly. The one open item is low-stakes: documenting the minimum Git version required for `--local-env-vars` so contributors know the floor. No blocking issues.

.githooks/pre-push
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+set -eu
+
+repo_root=$(git rev-parse --show-toplevel)
+git_local_env_vars=$(git rev-parse --local-env-vars)
+
```
tomo-chan, line 7: `git rev-parse --local-env-vars` is the canonical way to get this list, which is the right call. One minor open question: what is the minimum Git version that supports this flag? With `set -eu`, an unsupported flag would abort the hook before `make check` runs — safe, but would break pushes on very old Git installs. If there's a known minimum version for the project, is it worth documenting here or in DEVELOPMENT.md?

PR #137, opened 2026-06-20
feat: add workspace overview dashboard

tomo-chan:
## Summary
- add a persistent workspace overview dashboard rail that shows all workspaces and pane status at a glance
- aggregate live session state and git metadata across every pane for dashboard display
- cover the new dashboard, polling hooks, and schemas with frontend tests and document the behavior

## Test Plan
- [x] make test-frontend
- [x] make check


tomo-chan (commented):
## Code Review

### What's Good

- **Documentation updated in the same PR**: `docs/behavior.md` and `docs/ui-design.md` are updated alongside the implementation — the polling semantics, best-effort contract, and interaction model are all documented before the PR merges.
- **Schema-first**: `SessionInfoListSchema` is defined and exported from `schemas/index.ts`, and both hooks validate the response through Zod before consuming it, consistent with the project's runtime-validation rule.
- **Visibility-aware polling**: both `useSessionsOverview` and `useGitInfoMap` subscribe to `visibilitychange` and gate polling on tab visibility, matching the existing pane-header Git polling behaviour.
- **Attention routing on pane click**: `handleSelectOverviewPane` clears both pane and workspace attention before switching workspace — prevents stale attention badges after the user acts on a pane from the dashboard.
- **Hook test coverage**: `useSessionsOverview.test.ts` covers stop-while-hidden / resume-after-visible, error retention (previous state kept on non-ok response), and the disabled gate. `useGitInfoMap.test.ts` covers the stale-cleanup case when the ID list shrinks between renders.

---

### Issues & Questions

**[Question] Panes with no session data are counted as "connected"**

While the first `useSessionsOverview` fetch is in flight, `sessionsById` is `{}` and every pane's `state` is `undefined`. The counter's else-branch lands all of them in the `connected` bucket, so the workspace card briefly shows e.g. "3 up 0 down 0 exited" before real data arrives.

Is treating "no data yet" as optimistically connected intentional? If not, a fourth bucket or suppressing the count line while `sessionsById` is empty would avoid the misleading initial state.

**[Question] Dashboard is always visible with no collapse gesture**

The dashboard is rendered unconditionally when `workspaces` is truthy and occupies a fixed `min-width: 280px` column on the right of every workspace view. There is no toggle or resize handle.

Is this the intended final design (always-on rail), or is a collapse/minimize gesture planned? On narrower viewports (e.g. a laptop browser at ~1000 px wide) the terminal area gets noticeably squeezed.

**[Minor question] `collectChildPanes` silently drops a pane when `LayoutChild` carries both `pane` and `children`**

```ts
if (child.pane && (\!child.children || child.children.length === 0)) {
  panes.push(child.pane)
  return
}
// falls through to child.children loop — pane field is ignored
```

If `LayoutChildSchema` enforces that `pane` and `children` are mutually exclusive this is safe. Does it? If the schema permits both, a pane that also carries child splits would be silently omitted from the dashboard count and row list.

---

### Overall

The feature is well-structured: schema-first, visibility-aware, documented alongside the implementation, and both new hooks are thoroughly tested. The two questions above are about UX intent (loading-state display and always-visible layout) rather than correctness bugs and can be resolved post-merge if the current behaviour is intentional.

frontend/src/components/WorkspaceOverviewDashboard.tsx
```diff
@@ -0,0 +1,239 @@
...
+          exited: 0,
+        }
+
+        for (const pane of panes) {
+          const state = sessionsById[pane.id]?.state
+          if (state === 'disconnected') counts.disconnected += 1
+          else if (state === 'exited') counts.exited += 1
+          else counts.connected += 1
+        }
+
```
tomo-chan: While `sessionsById` is empty on the initial render, `state` is `undefined` and falls here, counting every pane as "connected". Is the intent to show optimistic "all up" counts while loading, or should panes without session data be excluded from the summary line?

frontend/src/components/WorkspaceOverviewDashboard.tsx
```diff
@@ -0,0 +1,239 @@
...
+  return panes
+}
+
+function collectChildPanes(
+  child: LayoutChild,
+  panes: Array<{ id: string; type: string; title?: string }>,
+) {
+  if (child.pane && (!child.children || child.children.length === 0)) {
+    panes.push(child.pane)
```
tomo-chan: If `LayoutChildSchema` enforces that `pane` and `children` are mutually exclusive, this condition is safe. If both can be present (e.g. a future nested split type), the pane is silently dropped — the function falls through to the `children` loop without pushing `child.pane`. Worth a comment or a schema assertion to make the invariant explicit.

tomo-chan: Addressed in https://github.com/tomo-chan/panemux/commit/4832dec7e1fa8041d809e796542884d224cd557c . I made the invariant explicit in `collectChildPanes`: the dashboard now treats any node carrying `pane` metadata as the terminal leaf for overview purposes instead of silently dropping it when malformed data also contains children.

tomo-chan:
Follow-up for the review summary:

- The always-visible right rail is intentional for this v1. The goal of this change is a persistent at-a-glance dashboard rather than a transient panel, so I kept it always on and documented that separation from the movable workspace bar.
- I also pushed a small follow-up fix in https://github.com/tomo-chan/panemux/commit/4832dec7e1fa8041d809e796542884d224cd557c to avoid reporting unfetched panes as `up`, and to make the pane-tree invariant explicit in the dashboard traversal.

Local verification re-run after the follow-up commit:
- `make test-frontend`
- `make check`

tomo-chan (commented):
## Code Review

### What's Good

- **Schema-first discipline**: `SessionInfoListSchema` lands in `schemas/index.ts` with dedicated acceptance/rejection tests before the hook consumes it — exactly matching the project's schema-first rule.
- **Hook test breadth**: Both `useSessionsOverview` and `useGitInfoMap` cover the disabled state, visibility toggling, polling cadence, and error recovery, using table-appropriate separate `it` blocks.
- **Visibility-aware polling**: Both hooks guard `useEffect` on `isVisible` so polling stops when the tab is hidden and fires an immediate refetch when visibility returns — consistent with how pane-header polling works.
- **Accessible markup**: `<aside aria-label="Workspace overview dashboard">` and descriptive `aria-label` on every button give screen readers full coverage without extra wrapper divs.
- **`collectWorkspacePanes` correctness**: Treating malformed nodes that carry both `pane` and `children` as leaf nodes (preferring `pane`, stopping recursion) is the right call, and the dedicated test exercises it.
- **`ReadonlySet<string>` for attention props**: Communicates immutability intent at the type boundary rather than relying on convention.

---

### Issues & Questions

**[Question] Dead branch: `state === 'connecting'` in pane counts**

In `WorkspaceOverviewDashboard.tsx`, the count loop reads:

```ts
else if (state === 'connected' || state === 'connecting') counts.connected += 1
```

`state` here is `sessionsById[pane.id]?.state`, which is `SessionInfo['state'] | undefined`. `SessionInfo['state']` is derived from `SessionStateSchema = z.enum(['connected', 'disconnected', 'exited'])`, so `state === 'connecting'` can never be true after Zod validation. The pane-row display text correctly falls back with `session?.state ?? 'connecting'`, but that UI string doesn't flow into the count loop.

Is `'connecting'` a planned future API state? If so, `SessionStateSchema` should be updated first so the Zod guard doesn't reject it. If not, removing the dead branch would keep the counts strictly aligned with the schema.

**[Minor question] `"keeps the previous state when response is not ok"` test may not exercise the failure path**

In `useSessionsOverview.test.ts`, the test sets up two mocked responses — the second returning `ok: false` — but does not call `vi.useFakeTimers()`. Without fake timers, the 10-second polling interval never fires, so `fetchMock` is only called once (the initial mount). The two `act` blocks re-await the already-settled first result. The final assertion passes, but it only confirms the initial connected state is retained — it never actually sends the `ok: false` response to the hook.

Consistent with the `"polls every 10 seconds while visible"` test in the same file, calling `vi.useFakeTimers()` and advancing by 10 000 ms would trigger the second poll and properly verify that a 500 response is silently dropped.

**[Minor question] `onSelectWorkspace` inline handler in `App.tsx`**

`onSelectPane` is a stable `useCallback` (`handleSelectOverviewPane`), but `onSelectWorkspace` is an inline arrow function passed directly to `WorkspaceOverviewDashboard`:

```tsx
onSelectWorkspace={(workspaceId) => {
  clearWorkspaceAttention(workspaceId)
  void setActiveWorkspace(workspaceId)
}}
```

If the component is ever wrapped in `React.memo`, the inline function would defeat memoization for that prop. Worth extracting to a `useCallback` alongside `handleSelectOverviewPane` for consistency.

---

### Overall

The dashboard is well-structured, the schema and test discipline is solid, and the visibility-aware polling matches the established pane-header pattern. The one logic question worth resolving before merge is the `'connecting'` branch in the count loop — either the schema needs to include the state, or the branch should be removed. The two minor questions are non-blocking.

tomo-chan:
Follow-up for the latest review summary:

- Removed the dead `state === "connecting"` branch from the workspace count loop so the summary strictly matches the validated `SessionInfo` schema. The row-level fallback still renders `connecting` when a pane has no fetched session state yet.
- Fixed the `useSessionsOverview` non-ok-response test so it actually advances the polling interval and exercises the second mocked `500` response instead of only re-awaiting the initial fetch.
- Also extracted the dashboard workspace-select callback in `App.tsx` to a stable `useCallback` for consistency with the pane-select path.

Addressed in https://github.com/tomo-chan/panemux/commit/0c38d4ce98a5eed0f1fddf0b41ab4232cfc7c08e .

Local verification re-run after this follow-up commit:
- `make test-frontend`
- `make check`

tomo-chan (commented):
## Code Review (round 2)

### Resolved since last review

- **Count-loop dead branch removed**: The `|| state === 'connecting'` check in the workspace count loop is gone — counts now align strictly with the validated `SessionInfo` schema. ✓
- **`useSessionsOverview` failure-path test fixed**: The test now calls `vi.useFakeTimers()` and advances 10 000 ms to actually deliver the `ok: false` response to the hook. ✓
- **`onSelectWorkspace` extracted to stable callback**: `handleSelectWorkspacePaneSummary` is now a `useCallback`, consistent with the rest of the pane-select path. ✓

### What's Good (new commits)

- **Integrated rail design**: Moving the summary surface into the workspace bar instead of a separate right-side panel is the right architectural call — it avoids permanently stealing horizontal terminal space and gives the summary a natural home alongside workspace navigation.
- **`workspaceDropHandlers` helper**: Extracting the six drag/drop event handlers into a factory function removes substantial duplication between the tab header div and the workspace details section.
- **`WorkspacePaneSummary.state` explicitly includes `'pending'`**: The new `'pending'` variant in the summary type correctly represents unseen-session state without touching the API schema.
- **Retry focus loop with cancellation**: The `attemptFocus` closure uses a `cancelled` flag and `window.clearTimeout` to safely abandon in-flight retries when the effect re-runs — a clean pattern for async focus that avoids stale state.
- **`flexWrap: 'wrap'` for workspace cards**: Wrapping workspace cards in horizontal mode means each card can show its full pane list without truncation, and the layout degrades gracefully when terminal width is narrow.

---

### Issues & Questions

**[Question] Residual dead branch in `collectChildPaneSummaries`**

`session?.state` is `SessionInfo['state'] | undefined` — i.e. `'connected' | 'disconnected' | 'exited' | undefined` after Zod validation. The check in `App.tsx`:

```ts
const state = session?.state === 'connecting' ? 'pending' : session?.state ?? 'pending'
```

The `=== 'connecting'` arm is never true for the same reason flagged in the previous review round. The expression is equivalent to `session?.state ?? 'pending'`. This survived the `0c38d4c6` fix because it's in `collectChildPaneSummaries` rather than the count loop. Worth a one-line cleanup.

**[Minor question] `CSS.escape()` instead of hand-rolled `escapeAttributeValue`**

In `App.tsx`, `focusPaneSurface` uses:

```ts
function escapeAttributeValue(value: string): string {
  return value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')
}
// ...
`[data-pane-id="${escapeAttributeValue(paneId)}"]`
```

This only escapes `\` and `"`. CSS attribute strings can contain other characters that are valid in a quoted string but may behave differently across parsers. The platform-standard API for this exact use case is `CSS.escape()`, available in all browsers panemux targets and in jsdom:

```ts
`[data-pane-id="${CSS.escape(paneId)}"]`
```

If pane IDs are guaranteed to be UUID-like strings with no special characters, the risk is low — but `CSS.escape()` is the safer and shorter form. Worth switching.

**[Question] `focusPaneSurface` querySelector may match header buttons before the terminal textarea**

The selector passed to `querySelector` is:

```ts
'.xterm-helper-textarea, textarea, [tabindex]:not([tabindex="-1"]), button, input'
```

`querySelector` returns the **first element in document order** that matches any alternative. In `TerminalPane`, the `PaneHeader` (which contains interactive buttons) appears before the xterm surface in the DOM. If any header button lacks `tabindex="-1"`, the `button` alternative matches it before `.xterm-helper-textarea` matches the terminal, and focus lands on the header rather than the terminal surface.

Do the pane-header buttons carry `tabindex="-1"`? If not, is the intent to focus the xterm textarea specifically, and should `.xterm-helper-textarea` be queried in isolation (falling back to the broader selector only when xterm isn't mounted yet)?

---

### Overall

All three findings from the previous round are cleanly resolved. The redesign from a separate right-side rail to an integrated workspace-bar panel is a strong improvement. The two questions above are the main things to resolve before merge: the residual dead branch is a one-liner, and the focus selector / `CSS.escape()` items are low-effort but worth confirming.

tomo-chan:
Follow-up for the latest review summary:

- Addressed the focus-targeting point in https://github.com/tomo-chan/panemux/commit/894a1dc9492ea733202df45d2e2da47d2c3c2288 by preferring `.xterm-helper-textarea` before broader interactive-element fallbacks when focusing a pane from the workspace overview, so summary clicks land on the terminal surface instead of header buttons when both are present.
- The same commit also switches the pane lookup selector to prefer `CSS.escape()` (with the existing minimal fallback kept only for environments where `CSS.escape` is unavailable), which covers the attribute-selector escaping concern.
- On the `collectChildPaneSummaries` question: I did not remove the `connecting` normalization because the current frontend schema still allows it in `SessionInfoSchema` (`state: z.enum(['connecting', 'connected', 'disconnected', 'exited'])`). The summary intentionally maps `connecting` to the local `pending` display state rather than surfacing it as a separate overview status.

Local verification re-run after the follow-up commit:
- `npm test -- --run src/App.test.tsx`
- `npm run build`
- `make check`

tomo-chan (commented):
## Code Review (round 3)

### Resolved since last review

- **`state === 'connecting'` clarified**: The author confirmed `SessionInfoSchema` includes `'connecting'` in its state enum, so the normalization to `'pending'` is correct. The updated form `session?.state === 'connecting' || \!session ? 'pending' : session.state` makes both conditions explicit. ✓
- **`CSS.escape()` adopted**: `escapeAttributeValue` now prefers `CSS.escape()` with a minimal fallback for environments where it is unavailable. ✓
- **`focusPaneSurface` selector tightened**: `.xterm-helper-textarea` is now queried first, with the broader interactive-element fallback used only when no xterm element is found. The dedicated test case (with a header button before the textarea) verifies this. ✓

### What's Good (new commits)

- **Hover/focus overlay for horizontal bars**: Keeping pane detail cards as an on-demand overlay for top/bottom bars prevents the workspace bar from consuming vertical terminal space; the overlay is only as tall as it needs to be and is dismissed on focus-out.
- **`onFocusCapture` / `onBlurCapture` for keyboard access**: Expanding the overlay on `focusCature` and collapsing on `blurCapture` (with the `currentTarget.contains(relatedTarget)` guard to keep the overlay open while focus moves within it) makes the summary panel keyboard-navigable, not just mouse-navigable. The relatedTarget null-check via `instanceof Node` is also type-correct.
- **`overlayPositionStyle` covers all four bar positions**: Each position variant gets a named anchor (below, above, right, left) so the overlay stays adjacent to its tab regardless of where the bar is docked.
- **Inline sections for vertical bars**: Keeping pane cards always expanded inline for left/right bars avoids the hover-to-see interaction when the bar is already taking horizontal space.

---

### Issues & Questions

**[Minor question] `aria-orientation` is now always `"horizontal"` for the tablist**

The diff changes:

```diff
-  aria-orientation={vertical ? 'vertical' : 'horizontal'}
+  aria-orientation="horizontal"
```

When the workspace bar is on the left or right, the workspace tabs visually appear in a top-to-bottom stack (one card per row, since `minWidth: 200` fills the 280–360 px bar). With `aria-orientation="horizontal"`, screen-reader users navigate with left/right arrow keys rather than up/down. For a visually vertical list this could be surprising.

The change is consistent with the underlying `flexDirection: 'row'` (wrapping rows), but the ARIA attribute describes the navigation experience, not the flex direction. Was this intentional? If the tabs are intended to feel like a horizontal list that happens to wrap, `horizontal` is correct; if they should feel like a vertical list when the bar is docked left/right, `vertical` would be more natural for keyboard navigation.

**[Minor question] `overflow: 'visible'` on the workspace bar container removes the previous scroll/clip boundary**

The outer bar wrapper changed:

```diff
-  overflowX: vertical ? 'hidden' : 'auto',
-  overflowY: vertical ? 'auto' : 'hidden',
+  overflow: 'visible',
```

`overflow: visible` is required so the absolutely-positioned overlay can extend beyond the bar bounds without being clipped. The trade-off is that the bar can no longer scroll its own content. With many workspaces in a horizontal bar, cards now wrap to a second (or third) row rather than scrolling, growing the bar's height and pushing the terminal area down.

For a small number of workspaces this is fine, but is there a maximum intended workspace count before the layout becomes unwieldy? If not, was a scroll fallback considered and ruled out, or is it worth adding a `max-height` + `overflow-y: auto` on the inner tablist container once the bar is in a constrained context?

**[Suggestion] Duplicate pane card JSX for overlay and inline sections**

The pane card markup (status dot, title, pill badges, detail row) appears identically in two places in `WorkspaceTabs.tsx`: inside `{overlaySummary && (…)}` and inside `{inlineSummary && (…)}`. Extracting it to a local `PaneCard` component would halve the code without changing behavior. Happy to file a follow-up issue if that would be helpful.

---

### Overall

All round-2 findings are fully addressed and the overlay/inline redesign is a clear improvement over the previous always-visible rail. The two minor questions above are both non-blocking; the `aria-orientation` one is the more important of the two to confirm before merge since it affects keyboard accessibility.

tomo-chan (commented):
## Code Review (round 4)

### What changed in the new commits

`eab932fb` adds `flexDirection: vertical ? 'column' : 'row'` to the workspace item wrapper, so for left/right bars the tab-header card and inline pane section stack in a column instead of sitting side by side. `d39777ff` covers this with a test that checks the computed `flexDirection: 'column'` on the wrapper element. Both are correct fixes.

---

### Carried forward from round 3

**[Minor question] `aria-orientation` is still always `"horizontal"` for left/right bars**

The round-3 question about this hasn't had a written reply yet, and neither new commit changes the attribute. Restating it briefly so it doesn't fall off the radar:

When the bar is docked left or right, workspace items each take `flex: '1 1 100%'` and `minWidth: '100%'`, so they still appear as a visual vertical stack — one workspace card per row. With `aria-orientation="horizontal"`, screen-reader users navigate with left/right arrow keys. For a visually vertical list this is non-obvious.

If the intent is a single consistent keyboard model regardless of bar position, that is a reasonable call — just worth documenting in the UI design notes so the decision is explicit rather than implicit.

**[Minor question] `overflow: 'visible'` and the many-workspace layout**

Still no response on this one. Restating: with `overflow: 'visible'` on the outer bar container and `flexWrap: 'wrap'` on the tablist, a horizontal bar with many workspaces wraps to multiple rows and grows vertically. There is no scroll fallback. Is there a practical maximum workspace count where this layout is still usable, or is this expected to grow unbounded?

**[Suggestion] Duplicate pane card JSX**

Pane card markup is still copied verbatim between the overlay section and the inline section. Happy to file a follow-up issue if that would be useful.

---

### Overall

The PR is in good shape. The two minor questions above are the only open items; neither blocks merge, but a brief acknowledgment (or doc update for `aria-orientation`) would make the intent clear to future contributors.

tomo-chan:
Follow-up on the round-4 notes:

- On `aria-orientation`: leaving it as `horizontal` is intentional. Even on left/right side bars, the workspace tabs still use the same wrapped horizontal tablist model rather than switching to a different keyboard interaction contract by dock position. I agree this should be explicit rather than implicit, so I'll treat the current behavior as an intentional consistency choice rather than an oversight.
- On `overflow: visible`: also intentional for this iteration. The top/bottom hover overlay needs to escape the bar bounds, and for now I preferred preserving that interaction over adding an additional scroll container. I agree that a many-workspace fallback is worth revisiting, but I see it as a follow-up UX constraint/tuning task rather than a blocker for this PR.
- On duplicated pane-card JSX: agreed. I did not refactor it in this PR because I wanted to keep the merge/conflict surface low while landing the interaction changes, but I agree that extracting a local card renderer would be the right cleanup pass.

Latest merge/conflict-resolution work is in https://github.com/tomo-chan/panemux/commit/e1fc5f24f694be37f41e359aaa61329e7a65b240 .

Local verification already re-run on that merge commit via `make check`. GitHub required checks are still pending as of this comment (`ci`, `Analyze (go)`).

tomo-chan (commented):
## Code Review (round 5)

### Round-4 questions resolved

All three open items from round 4 now have explicit author decisions:

- **`aria-orientation` "horizontal"**: intentional consistency choice — the wrapped-row tablist model uses a single horizontal keyboard navigation contract regardless of dock position. The author will document this as an intentional decision rather than an oversight. ✓
- **`overflow: 'visible'` / many-workspace fallback**: intentional for this iteration. The overlay needs to escape bar bounds without a clip container; a scroll fallback for large workspace counts is tracked as a follow-up UX task. ✓
- **Duplicate pane-card JSX**: acknowledged as a follow-up extraction. The duplication was left intentionally to keep the merge surface small during this PR. ✓

`e1fc5f24` is a merge from `main` — the diff footprint and changed-file list are unchanged from round 4, with no new logic introduced by the merge commit.

---

### Overall

All substantive findings across five review rounds have been addressed or consciously deferred with documented rationale. `make check` is confirmed passing on the merge commit. The PR is ready to merge once CI goes green.

tomo-chan (commented):
## Code Review (round 6)

### Resolved since last review

All items from round 5 were confirmed closed by the author. This round covers only the new commit \`eef3fca2\` ("Reuse pane git info in workspace summaries").

---

### What's Good

- **Pub/sub cache layer is well-scoped**: `useGitInfoSnapshotMap` subscribes to updates from the shared `gitInfoCache` without issuing any new network requests, so the overview doesn't add a second polling layer on top of what the visible pane headers already do. The listener is cleaned up on unmount and the set is pruned when empty — no leak risk.
- **Cross-consumer cache test is explicit**: The test "shares cached git info with snapshot consumers without triggering extra fetches" pins the no-extra-fetch contract in an observable way rather than relying on implementation inspection.
- **`__resetGitInfoCacheForTests` extended**: `gitInfoListeners.clear()` was added alongside `gitInfoCache.clear()`, so tests can't bleed listener state across cases.

---

### Issues & Questions

**[Question] `useGitInfoMap` appears to be unused in production code**

`frontend/src/hooks/useGitInfoMap.ts` is added as a new file with its own 82-line test suite, but no production file imports it. Searching the entire `frontend/src` tree finds no caller:

```
# grep -r "useGitInfoMap" frontend/src  →  no matches
```

`App.tsx` uses `useGitInfoSnapshotMap` from `useGitInfo.ts` instead. Should `useGitInfoMap` be deleted? If it's being kept for a planned follow-up, a brief inline comment noting the intent would help reviewers understand why it's there.

---

### Overall

One open question: the `useGitInfoMap` hook and its test file appear to be dead code in this commit. Everything else in the cache-listener implementation is clean and the design choice to reuse pane-triggered cache updates (rather than independent polling) is sound. Please confirm or remove `useGitInfoMap` before merging.

tomo-chan (commented):
## Code Review (round 7)

### Resolved since last review

No author reply yet to the round-6 `useGitInfoMap` question — carrying it forward below.

---

### What's Good

- **`data-testid` is safe despite appearing twice**: `workspace-pane-group-${workspace.id}` is added to both the overlay (`overlaySummary`) and inline (`inlineSummary`) pane card containers, but the two are gated on `\!vertical` and `vertical` respectively, so they are mutually exclusive in the DOM. `getByTestId` is safe to use and won't match multiple elements for the same workspace.
- **Visual hierarchy reinforced by testing**: The border-left and indent styling is asserted in tests (`toHaveStyle`), so this visual grouping cue is regression-protected, not just cosmetic.

---

### Issues & Questions

**[Question] `useGitInfoMap` still appears unused (carried from round 6)**

`frontend/src/hooks/useGitInfoMap.ts` and its test file `useGitInfoMap.test.ts` were added in `eef3fca2` but no production file imports `useGitInfoMap`. `App.tsx` uses `useGitInfoSnapshotMap` instead. The new commit `71a94802` does not address this.

Should `useGitInfoMap` be deleted before merging, or is there a planned use in a follow-up?

---

### Overall

`71a94802` is a clean visual refinement with test coverage. The only remaining open item is whether `useGitInfoMap` is intentionally kept or should be removed.

tomo-chan:
Addressed the remaining round-6/7 question by removing the unused hook and test:

- Removed the dead frontend/src/hooks/useGitInfoMap.ts implementation.
- Removed the dead frontend/src/hooks/useGitInfoMap.test.ts coverage that only exercised that unused hook.

Change is in https://github.com/tomo-chan/panemux/commit/1823b263927b72c354c1a7a1c05d69c04d2f8307 .

Local verification re-run after the cleanup:
- npm test -- --run src/hooks/useGitInfo.test.ts src/App.test.tsx src/components/WorkspaceTabs.test.tsx
- npm run build
- make check

GitHub checks had not been re-associated with the latest push yet at the moment of posting this reply.

tomo-chan (commented):
## Code Review (round 8)

### Round-6/7 question resolved

**`useGitInfoMap` removed**: Commit \`1823b263\` deletes both `frontend/src/hooks/useGitInfoMap.ts` (63 lines) and `frontend/src/hooks/useGitInfoMap.test.ts` (82 lines). The author confirmed the hook was superseded by `useGitInfoSnapshotMap` and removed both the implementation and the now-orphaned test coverage. No other files affected.

---

### Overall

All open items across rounds 1–7 are now resolved. The PR is clean.

PR #138, opened 2026-06-20
fix: preserve maximize state across workspace switches

tomo-chan:
## Summary

- preserve maximized pane state per workspace in the frontend
- restore the previous maximized pane when switching back to a workspace
- cover the behavior in `App` tests and document the workspace-scoped maximize rule

## Test Plan

- [x] `npm test -- --run src/App.test.tsx`
- [x] `make lint-frontend`
- [x] `make test-frontend`
- [x] `make check`

### DIFF
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -183,6 +183,7 @@ describe('App workspace deletion', () => {
     expect(document.querySelector('[data-pane-id="main"]')).toHaveAttribute('data-attention', 'true')
     expect(document.querySelector('[data-pane-id="side"]')).toHaveAttribute('data-attention', 'true')
 
+    // Both visible panes render the same mock controls, so target the first one.
     fireEvent.click(screen.getAllByRole('button', { name: 'Clear main' })[0])
 
     expect(document.querySelector('[data-pane-id="main"]')).not.toHaveAttribute('data-attention')
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -159,10 +159,10 @@ describe('App workspace deletion', () => {
 
   it('keeps workspace attention while another pane in that workspace still needs attention', () => {
     currentWorkspaces = { ...workspaces, active: 'ops' }
-    mockTerminalPane.mockImplementation(() => {
+    mockTerminalPane.mockImplementation(({ pane }: { pane: { id: string } }) => {
       const ctx = useContext(LayoutActionsContext)
       return (
-        <div>
+        <div data-pane-id={pane.id} data-attention={ctx?.hasPaneAttention(pane.id) ? 'true' : undefined}>
           <button onClick={() => ctx?.onPaneAttention('main')}>Notify main</button>
           <button onClick={() => ctx?.onPaneAttention('side')}>Notify side</button>
           <button onClick={() => ctx?.clearPaneAttention('main')}>Clear main</button>
@@ -171,19 +171,26 @@ describe('App workspace deletion', () => {
       )
     })
 
-    render(<App />)
+    const { rerender } = render(<App />)
     fireEvent.click(screen.getByRole('button', { name: 'Notify main' }))
     fireEvent.click(screen.getByRole('button', { name: 'Notify side' }))
 
     expect(screen.getByRole('tab', { name: 'Dev' })).toHaveAttribute('data-attention', 'true')
 
-    fireEvent.click(screen.getByRole('button', { name: 'Clear main' }))
+    currentWorkspaces = workspaces
+    rerender(<App />)
 
-    expect(screen.getByRole('tab', { name: 'Dev' })).toHaveAttribute('data-attention', 'true')
+    expect(document.querySelector('[data-pane-id="main"]')).toHaveAttribute('data-attention', 'true')
+    expect(document.querySelector('[data-pane-id="side"]')).toHaveAttribute('data-attention', 'true')
 
-    fireEvent.click(screen.getByRole('button', { name: 'Clear side' }))
+    fireEvent.click(screen.getAllByRole('button', { name: 'Clear main' })[0])
 
-    expect(screen.getByRole('tab', { name: 'Dev' })).not.toHaveAttribute('data-attention')
+    expect(document.querySelector('[data-pane-id="main"]')).not.toHaveAttribute('data-attention')
+    expect(document.querySelector('[data-pane-id="side"]')).toHaveAttribute('data-attention', 'true')
+
+    fireEvent.click(screen.getAllByRole('button', { name: 'Clear side' })[0])
+
+    expect(document.querySelector('[data-pane-id="side"]')).not.toHaveAttribute('data-attention')
   })
 
   it('confirms before deleting a workspace from edit-mode tabs', () => {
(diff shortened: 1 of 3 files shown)


frontend/src/App.test.tsx
```diff
@@ -174,14 +176,10 @@ describe('App workspace deletion', () => {
     fireEvent.click(screen.getByRole('button', { name: 'Notify side' }))
 
     expect(screen.getByRole('tab', { name: 'Dev' })).toHaveAttribute('data-attention', 'true')
-    expect(screen.getByText('Notify main').closest('[data-pane-id="main"]')).toHaveAttribute('data-attention', 'true')
```
tomo-chan, line 177: These four assertions (two here and two after the `Clear main` step) were the only coverage of `hasPaneAttention` flowing through `LayoutActionsContext` to individual pane components. The tab-level assertion above is preserved, but per-pane context propagation is no longer exercised. Was dropping these intentional?

frontend/src/App.tsx
```diff
@@ -17,7 +17,7 @@ const DEFAULT_DISPLAY: DisplayConfig = { show_header: true, show_status_bar: tru
 
 export const App: React.FC = () => {
   const { layout, workspaces, displayConfig, error, updateSizes, splitPane, closePane, swapPanes, createPane, movePane, setActiveWorkspace, addWorkspace, deleteWorkspace, renameWorkspace, setWorkspaceTabPosition } = useLayout()
-  const [maximizedPaneId, setMaximizedPaneId] = useState<string | null>(null)
+  const [maximizedPaneIdsByWorkspace, setMaximizedPaneIdsByWorkspace] = useState<Record<string, string | null>>({})
   const [dragSourcePaneId, setDragSourcePaneId] = useState<string | null>(null)
```
tomo-chan, line 21: When `deleteWorkspace` is called, the entry for that workspace ID will stay in this map for the lifetime of the session. Not a correctness bug (the deleted ID never matches `activeWorkspaceId` again), but the map grows without bound across many add/delete cycles. Worth clearing the key inside the `deleteWorkspace` handler, or intentional to keep it simple?

tomo-chan (commented):
## Code Review

### What's Good

- **Per-workspace state map with derived accessor** (`App.tsx`): Storing `maximizedPaneIdsByWorkspace` as `Record<string, string | null>` and deriving the current `maximizedPaneId` via `useMemo` is clean. The derived value mirrors the old API shape exactly, so no other component needed to change.
- **`layoutContainsPane` guard in the derived memo**: Checking that the recorded pane ID still exists in the current layout means a closed pane can never be stuck as maximized — the stale map entry is silently ignored on the next render without requiring coordinated cleanup in `closePane`.
- **Identity check in `setMaximizedPaneId`**: The `if ((current[activeWorkspaceId] ?? null) === paneId) return current` early return avoids triggering downstream re-renders when the value hasn't changed. Good optimization.
- **`devLayout` rename**: Naming the fixture after the workspace it belongs to makes the test data relationships immediately clear.
- **`useLayout` mock now tracks active workspace**: Returning the layout for the currently active workspace makes the mock behave like the real hook and correctly exercises the `layoutContainsPane` guard in the new maximize logic.
- **`docs/behavior.md` updated in the same PR**: The persistence rule for maximize state is documented alongside the implementation change.

---

### Issues & Questions

**[Question] Four per-pane attention assertions dropped**

The mock refactoring in `'keeps workspace attention while another pane in that workspace still needs attention'` removes `data-pane-id` and `data-attention` from the mock output, which eliminates four assertions that verified `hasPaneAttention(paneId)` propagates through `LayoutActionsContext` to individual pane components. The tab-level assertion (`data-attention` on the `Dev` tab) remains, but per-pane context propagation is no longer covered.

Was this an intentional coverage trade-off (e.g. this path is covered by a different test), or did it fall out of the refactor by accident?

**[Minor question] Deleted workspace entries linger in `maximizedPaneIdsByWorkspace`**

When `deleteWorkspace` is called, the entry for that workspace ID stays in the map indefinitely. There is no correctness issue — the deleted ID will never match `activeWorkspaceId` again — but the map grows unboundedly if users create and delete many workspaces over a session. Should `deleteWorkspace` clear the corresponding key, or is keeping the handler simple the intentional trade-off here?

---

### Overall

The per-workspace maximize state approach is well-designed and correctly handles workspace switching, pane closure, and layout validation. Ready to merge pending confirmation on the dropped attention assertions.

tomo-chan:
Follow-up for the review questions:

- Restored the per-pane attention assertions in https://github.com/tomo-chan/panemux/commit/263eca1d5cf3b3022abde9a16adbb4e09ccbdf8c
- The test now switches back into the workspace and re-checks pane-level `data-attention`, so `LayoutActionsContext.hasPaneAttention()` is exercised again instead of only the workspace-tab indicator.
- The same commit also prunes `maximizedPaneIdsByWorkspace` entries when a workspace disappears, so deleted workspace IDs do not accumulate in the frontend session map.

Verified locally:

- `npm test -- --run src/App.test.tsx`
- `make lint-frontend`
- `make test-frontend`
- `make check`

tomo-chan (commented):
## Code Review

### What's Good

- **Per-workspace map with derived accessor** (`App.tsx`): Storing `maximizedPaneIdsByWorkspace` as `Record<string, string | null>` and deriving the current `maximizedPaneId` via `useMemo` is a clean pattern — no new context layer or global state needed, and the call sites stay unchanged.
- **`layoutContainsPane` guard in the memo**: Checking that the stored pane ID is still present in the active layout correctly handles the «maximize a pane, then close it» case without any extra cleanup path.
- **Identity-equality optimization in `useEffect`**: Returning `current` when no workspace IDs were removed avoids a spurious re-render on every workspace event that doesn't involve deletion. Good React discipline.
- **`setMaximizedPaneId` wraps the workspace-keyed setter cleanly**: Consumers of `LayoutActionsContext` call `onMaximize(paneId)` exactly as before; the workspace scoping is fully internal.
- **Test coverage**: The two new tests (`preserves maximize state`, `drops maximize state`) each verify a distinct behavior, and the round-trip case (switch away → switch back → state intact) is explicitly covered.
- **Docs updated in the same change**: The `behavior.md` bullet is in the right place.

Both concerns raised in the previous review — per-pane attention assertion coverage and map growth on workspace deletion — have been fully addressed in the follow-up commit.

---

### Issues & Questions

**[Minor question] `getAllByRole(...)[0]` in the attention test**

In the refactored `mockTerminalPane` for the attention test, the button labels are now hardcoded to `'main'` and `'side'` rather than using the dynamic `pane.id`:

```tsx
<button onClick={() => ctx?.clearPaneAttention('main')}>Clear main</button>
<button onClick={() => ctx?.clearPaneAttention('side')}>Clear side</button>
```

Because every rendered pane (both `main` and `side`) produces this same fixed set of buttons, there are two `Clear main` buttons in the DOM when the dev workspace is active, which is why `getAllByRole('button', { name: 'Clear main' })[0]` is needed instead of `getByRole`.

Was the hardcoding intentional — to decouple the test from workspace-specific pane IDs that change across rerenders (e.g. `'ops-main'` vs `'main'`)? If so, a short comment near the `[0]` call explaining why multiple matches exist would help future readers understand why `getByRole` isn't used here.

---

### Overall

The implementation is clean, well-tested, and the documentation is updated. The one open item is a minor clarification question about the intentionality of the hardcoded button labels in the attention test. Ready to merge once that's confirmed or annotated.

tomo-chan:
Confirmed and annotated in https://github.com/tomo-chan/panemux/commit/416779be1c66b12c5a74a548c95f17f7e1194101

- The hardcoded mock labels are intentional so the attention test can drive hidden-workspace pane IDs (`main` / `side`) even when the currently rendered workspace has different pane IDs such as `ops-main`.
- Because both visible mock panes render the same control labels after the rerender back into `dev`, there are duplicate `Clear main` / `Clear side` buttons by design.
- Added a short comment next to `getAllByRole(...)[0]` to make that explicit for future readers.

Verified locally:

- `npm test -- --run src/App.test.tsx`
- `make lint-frontend`
- `make check`

tomo-chan (commented):
## Code Review

### Follow-up on Previous Findings

All three concerns from the prior two reviews are now resolved:

- **Per-pane attention assertions** — restored in commit `263eca18`. The attention test now re-checks per-pane `data-attention` after switching back into the workspace, so `LayoutActionsContext.hasPaneAttention()` is exercised end-to-end again.
- **Map memory growth on workspace deletion** — addressed via the `useEffect` pruning loop in `App.tsx`. The `drops maximize state for workspaces that no longer exist` test verifies the GC path explicitly.
- **`getAllByRole(...)[0]` clarity** — annotated in commit `416779b6` with `// Both visible panes render the same mock controls, so target the first one.` The author also confirmed in the PR comment that the hardcoded labels are intentional: they allow the test to address specific pane IDs (`main` / `side`) even when the active workspace renders different IDs such as `ops-main`.

No new issues in the latest commit.

---

### Overall

All concerns addressed and confirmed. The implementation, tests, and documentation are in good shape. Ready to merge.

PR #140, opened 2026-06-21
refactor: reduce pane header git info refresh cost

tomo-chan:
## Summary
- replace 10-second pane header git-info polling with stale-on-interaction refresh
- refresh git info when panes become visible, focused, clicked, or actively used
- link terminal-side PR references like `#123` to the repository pull request when repo metadata is known

## Test Plan
- [x] make test-frontend
- [x] make check

### DIFF
--- a/frontend/src/components/TerminalPane.test.tsx
+++ b/frontend/src/components/TerminalPane.test.tsx
@@ -35,7 +35,11 @@ describe('TerminalPane drop zones', () => {
       reconnectFailed: false,
       restartSession: vi.fn(),
     })
-    mockUseGitInfo.mockReturnValue({ is_git: false })
+    mockUseGitInfo.mockReturnValue({
+      gitInfo: { is_git: false },
+      refreshIfStale: vi.fn(),
+      refreshNow: vi.fn(),
+    })
   })
 
   it('uses a top or bottom half-pane preview for horizontal edge drop zones', () => {
--- a/frontend/src/components/TerminalPane.tsx
+++ b/frontend/src/components/TerminalPane.tsx
@@ -31,12 +31,14 @@ export const TerminalPane: React.FC<TerminalPaneProps> = ({ pane }) => {
   const dragActive = Boolean(ctx?.dragSourcePaneId)
   const gitInfoEnabled = !ctx?.maximizedPaneId || ctx.maximizedPaneId === pane.id
 
+  const { gitInfo, refreshIfStale, refreshNow } = useGitInfo(pane.id, gitInfoEnabled)
+
   const { handleResize, connected, dims, sessionState, reconnectFailed, restartSession } = useTerminal({
     sessionId: pane.id,
     container: containerEl,
+    repoURL: gitInfo.repo_url,
+    onInteraction: refreshIfStale,
   })
-
-  const gitInfo = useGitInfo(pane.id, gitInfoEnabled)
 
   // Observe resize events for this pane
   useEffect(() => {
@@ -49,9 +51,10 @@ export const TerminalPane: React.FC<TerminalPaneProps> = ({ pane }) => {
   }, [containerEl, handleResize])
 
   const handleOpenVSCode = useCallback(() => {
+    void refreshNow()
     fetch(`/api/sessions/${pane.id}/open-vscode`, { method: 'POST' })
       .catch((err) => console.error('open-vscode failed:', err))
-  }, [pane.id])
+  }, [pane.id, refreshNow])
 
   const handleDragStart = useCallback((e: React.DragEvent) => {
     e.dataTransfer.effectAllowed = 'move'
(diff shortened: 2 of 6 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Polling → stale-on-interaction**: eliminating the 10-second interval is a clean improvement. Most panes are idle most of the time, and the 30 s staleness window with 5 s throttle is a well-judged trade-off.
- **Module-level cache with in-flight deduplication**: `getOrCreateCacheEntry` + `cacheEntry.inFlight` prevents redundant concurrent fetches cleanly. The `__resetGitInfoCacheForTests` export is the right test-isolation idiom for module-level singletons.
- **`#123` → GitHub PR link**: contextually useful and scoped to when `repo_url` is available, so it degrades gracefully for non-GitHub repos.
- **Test coverage**: throttle, staleness, deduplication, missing buffer line, and in-flight reuse are all covered. Table-driven test naming is clear.
- **`onInteraction` abstraction**: exposing this from `useTerminal` keeps the refresh trigger in the presentation layer and the mechanism in the hook — good separation.

---

### Issues & Questions

**[Question] Non-ok HTTP response test removed without replacement**

`useGitInfo.test.ts` previously had `'silently ignores non-ok responses — state stays at default'`, which verified that `result.current.is_git` stays false after a 404/500 response. That test was removed. The new `'throttles repeated stale checks after a non-ok response'` only asserts `fetch` call count; it does not assert that `gitInfo.is_git` is still false. Is the positive assertion intentionally dropped, or was it accidentally lost in the refactor?

**[Minor question] PR link range `start.x` / `end.x` semantics**

In `frontend/src/hooks/useTerminal.ts`:

```ts
links.push({
  range: {
    start: { x: hashIndex + 1, y },
    end:   { x: hashIndex + 1 + number.length + 1, y },
  },
```

Per the xterm.js typings, `IViewportRangePosition.x` is **0-based** and refers to the space *between* columns (cursor-like), so `x: 0` is the left edge of column 0 and `x: N` is the right edge of column N-1. For `#123` at 0-based text position `hashIndex`:

- expected start (left edge of `#`): `hashIndex`
- expected end (right edge of `3`): `hashIndex + 4`

The code produces `start.x = hashIndex + 1` and `end.x = hashIndex + 5`, each shifted one position to the right. If xterm.js actually uses 1-based x in its link provider (which would be inconsistent with the typings but consistent with `bufferLineNumber` being 1-based in `provideLinks`), then `start.x = hashIndex + 1` would be correct and `end.x = hashIndex + number.length + 2` would be exclusive-end. Could you confirm which convention xterm.js expects here? The existing tests don't assert the `range` field values, so there's no test coverage for this.

**[Minor question] Non-passive `wheel` listener**

In `frontend/src/hooks/useTerminal.ts`:

```ts
container.addEventListener('wheel', handleInteraction, { capture: true })
```

The handler only calls `void onInteraction()` and never calls `e.preventDefault()`, so adding `passive: true` to the options object would let the browser skip waiting for the listener before scrolling. Worth changing to `{ capture: true, passive: true }` (and matching the `removeEventListener` call) for scroll performance, especially on high-DPI trackpads.

---

### Overall

The polling-to-stale-interaction migration is a meaningful improvement and the code is well-structured. The two minor questions above (range semantics and passive listener) are worth confirming before merge; the missing non-ok response assertion is the one I'd want addressed before landing.

tomo-chan:
Follow-up addressed in https://github.com/tomo-chan/panemux/commit/1f1eb44dfe9279361ae3ed33f39ca6db4ef7f4fc

- restored the explicit `non-ok` response state assertion in `useGitInfo` tests
- marked the terminal `wheel` listener as passive because it does not call `preventDefault()`
- corrected the xterm PR link viewport range to use 0-based between-column coordinates, and added an assertion for the exact range in tests

Local verification run after the follow-up:
- `make check`

tomo-chan (commented):
## Code Review — Follow-up (commit 1f1eb442)

All three findings from the previous round are resolved.

### Resolved

**[Question] Non-ok HTTP response assertion** — ✅ `expect(result.current.gitInfo.is_git).toBe(false)` restored to `'throttles repeated stale checks after a non-ok response'`. The test now asserts both the throttle behaviour and the resulting state.

**[Minor question] `wheel` listener passive** — ✅ `{ capture: true, passive: true }` applied. The `removeEventListener` call correctly continues to use `true` — `passive` is not considered when matching listeners to remove, so this is spec-compliant.

**[Minor question] PR link range coordinates** — ✅ Both axes corrected to 0-based: `start: { x: hashIndex, y: y - 1 }`, `end: { x: hashIndex + number.length + 1, y: y - 1 }`. The new range assertion in the test (`start: { x: 10, y: 0 }`, `end: { x: 14, y: 0 }`) pins the exact values for `"Reviewing #123 now"` at buffer row 1 and confirms the fix is correct.

---

### Overall

No outstanding issues. The PR is clean to merge.

PR #142, opened 2026-06-22
feat: resize vertical workspace bar

tomo-chan:
## Summary

- add persisted `vertical_bar_width` workspace config and API support for vertical workspace bar resizing
- make left/right workspace bars resizable with a fixed footer for the `+` action and tab-position controls
- keep workspace tabs and inline summaries in a dedicated vertical scroll region so controls remain visible

## Test plan

- [x] `make check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Then open [http://localhost:8080](http://localhost:8080) in your browser.
 - Click a tab to switch workspaces without restarting the underlying sessions.
 - If a hidden workspace needs attention, its tab flashes until you open it.
 - You can add a workspace, rename it inline, delete it, or move the tab bar to the top, bottom, left, or right directly from the workspace bar.
+- When the bar is on the left or right, you can drag its inner edge to resize it. That shared vertical width is persisted with the workspace settings.
 - The workspace bar also shows per-workspace pane status at a glance. Each tab includes a compact pane-name summary, and each workspace can expose pane detail cards with connection state, SSH host alias, repository, branch, and PR number.
 - When the bar is on the top or bottom, pane detail cards appear as a hover/focus overlay attached to the workspace tab so the terminal area keeps its height.
 - When the bar is on the left or right, pane detail cards stay expanded inline under each workspace tab so the workspace-to-pane grouping remains visible without extra pointer travel.
@@ -153,6 +154,7 @@ ssh_connections:
 workspaces:
   active: dev
   tab_position: top           # top | bottom | left | right
+  vertical_bar_width: 280     # shared width in px when tab_position is left/right
   items:
     - id: dev
       title: "Development"
@@ -186,7 +188,7 @@ workspaces:
 
 Older config files with a top-level `layout:` are still accepted. When the config is next saved, panemux migrates that layout into a `default` workspace and writes the `workspaces:` format.
 
-The workspace bar is always available for workspace actions. Newly added workspaces start with a single local terminal pane, become active immediately, and are saved to the config. The same bar also exposes inline workspace rename, workspace delete with confirmation, a `+` workspace action, and tab-position controls. It also acts as a cross-workspace operational overview: pane detail cards can be selected to focus that pane, and those same cards can be dragged onto another workspace to move the pane there. Switching the active workspace and changing `tab_position` are persisted so the same workspace and tab placement are restored after restart. Agent confirmation prompts, including Codex MCP allow menus, can mark inactive workspace tabs and trigger browser notifications after notification permission has been granted. Clicking the notification switches to the relevant workspace.
+The workspace bar is always available for workspace actions. Newly added workspaces start with a single local terminal pane, become active immediately, and are saved to the config. The same bar also exposes inline workspace rename, workspace delete with confirmation, a `+` workspace action, and tab-position controls. When the bar is vertical, its width is shared across all workspaces and is also persisted. It also acts as a cross-workspace operational overview: pane detail cards can be selected to focus that pane, and those same cards can be dragged onto another workspace to move the pane there. Switching the active workspace, changing `tab_position`, and resizing the vertical workspace bar are persisted so the same navigation chrome is restored after restart. Agent confirmation prompts, including Codex MCP allow menus, can mark inactive workspace tabs and trigger browser notifications after notification permission has been granted. Clicking the notification switches to the relevant workspace.
 
 ### Pane types
 
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -53,7 +53,8 @@ ssh_connections:
 # workspaces, or move workspace tabs. New workspaces start with one local
 # terminal pane. Deleting a workspace asks for confirmation, and the last
 # remaining workspace cannot be deleted. The active workspace selection and
-# tab_position are saved.
+# tab_position are saved. `vertical_bar_width` controls the shared width in
+# pixels when the workspace bar is on the left or right.
 #
 # Split node  — has `direction` and `children` (no `pane`)
 # Leaf node   — has `pane` (no `direction` / `children`)
@@ -72,6 +73,7 @@ ssh_connections:
 workspaces:
   active: dev
   tab_position: top
+  vertical_bar_width: 280
   items:
     - id: dev
       title: "Development"
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -103,7 +103,7 @@ Pane settings in the frontend expose a directory browser for `cwd`. Local and lo
 Persistence behavior:
 
 - layout and workspace changes are persisted immediately when a save path is available
-- pane resize, split, close, quick-add create, move, workspace add/delete/rename, active-workspace changes, and `tab_position` changes all follow the same immediate-save path
+- pane resize, split, close, quick-add create, move, workspace add/delete/rename, active-workspace changes, `tab_position` changes, and vertical workspace-bar width changes all follow the same immediate-save path
 - when no `--config` is given, the default save path is `~/.config/panemux/config.yaml`; the directory is created automatically on first save
 - pane maximize state is frontend-local, tracked per workspace, and restored when the user switches away from a workspace and then returns
 
@@ -159,7 +159,8 @@ panes across all workspaces without mounting hidden terminal instances. The inte
 marks the currently focused pane so the overview stays aligned with the live terminal focus state.
 For `top` and `bottom` workspace bars, pane cards are shown in a hover/focus overlay anchored to
 the workspace tab. For `left` and `right` workspace bars, pane cards stay expanded inline beneath
-their workspace tab.
+their workspace tab. In vertical mode, the workspace tabs and inline cards scroll inside the bar,
+while the `+` action and tab-position controls stay pinned at the bottom.
 
 When a pane is hidden behind maximize, its header Git/PR polling stops until it becomes visible
 again. Restoring the browser tab or making a pane visible again triggers an immediate refresh so
@@ -201,7 +202,7 @@ Accepts a layout JSON document, validates it, updates in-memory state, and persi
 
 ### `GET /api/workspaces`
 
-Returns the active workspace ID, `tab_position`, and all workspace layouts.
+Returns the active workspace ID, `tab_position`, `vertical_bar_width`, and all workspace layouts.
 
 ### `PUT /api/workspaces/tab-position`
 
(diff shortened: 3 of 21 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **`SummarySection` extraction** (`WorkspaceTabs.tsx`): Pulling the pane card rendering into its own component eliminates ~250 lines of near-identical duplication for the overlay and inline cases. Parameterising via `overlay` is clean and keeps the call sites readable.
- **`applyWorkspaceSettingUpdate` helper** (`handler.go`): The shared 422 path for tab-position and vertical-bar-width is now DRY without introducing indirection that hurts readability.
- **Validation bounds are consistent**: Go uses `180`–`520` (`minWorkspaceVerticalBarWidth` / `maxWorkspaceVerticalBarWidth`) and Zod uses `.min(180).max(520)`. No silent disagreement.
- **Debounced persistence with optimistic UI**: 150 ms debounce on the API call while the local state tracks every pixel is the right model for a drag-resize. The response merge (`{ ...parsed, items: current.items }`) prevents discarding any layout change that raced the async save — same pattern as `setWorkspaceTabPosition`.
- **Backward compatibility**: `normalizedWorkspaces` defaults `VerticalBarWidth` to 280 when the field is absent (Go zero value from an old YAML), so existing configs load cleanly without a migration step.
- **Test coverage is thorough**: config, validate, handler (200/400/422), hook (optimistic + failure), and component (resizer presence, scroll region, footer structure) all have their own cases.

---

### Issues & Questions

**[Question] Drag triggers a React state update on every pixel**

In `WorkspaceTabs.tsx`, the `mousemove` listener calls `onVerticalBarWidthChange` on every event. In `useLayout.ts`, each call does two synchronous `setState` invocations (one for `error`, one for `workspaces`) plus a Zod parse before the debounce fires. React 18 batches these, but for a large workspace list with many inline pane cards in `SummarySection`, this could cause noticeable jank on slower machines.

An alternative is to keep a local `dragWidth` ref inside `WorkspaceTabs` and only commit to `onVerticalBarWidthChange` on `mouseup`, driving the visual width from that local ref during the drag itself. Is the current every-pixel approach intentional, or was the jank risk considered acceptable?

**[Minor question] `VerticalBarWidth == 0` skips validation in `validateWorkspaces`**

```go
// internal/config/validate.go
if workspaces.VerticalBarWidth \!= 0 {
    if err := ValidateWorkspaceVerticalBarWidth(workspaces.VerticalBarWidth); err \!= nil {
        errs = append(errs, err.Error())
    }
}
```

The zero-skip exists so that old configs missing the field (Go zero value from `omitempty`) don't fail validation — the normalizer later replaces 0 with 280. But a user who manually writes `vertical_bar_width: 0` in their YAML would pass validation and then silently get 280 applied at runtime. Is this the intended behaviour, or would a deliberate-zero be better rejected and flagged?

**[Minor question] Resizer is mouse-only**

The drag handle (`workspace-bar-resizer`) listens only to `mousedown`/`mousemove`/`mouseup`. Touch-screen users and keyboard-only users cannot resize the bar. Is this a known limitation to file as a follow-up, or is pointer-event / keyboard support planned?

---

### Overall

The feature is well-structured: schema-first (Zod then Go validate), symmetric bounds, clean component extraction, and solid test coverage. The three items above are questions and minor notes rather than blockers. Ready to merge once the drag-per-pixel question is resolved to the author's satisfaction.

tomo-chan:
Follow-up on the review:

- I addressed the drag/update concern in <https://github.com/tomo-chan/panemux/commit/f794b0c602db42a82f4fcdf715084b983a4f90e2>. The resizer now keeps the in-progress width local to `WorkspaceTabs` during drag and only calls `setWorkspaceVerticalBarWidth` when the drag ends, so we no longer push every pixel movement through `useLayout` state and Zod parsing.
- I also addressed the mouse-only limitation in the same commit: the separator is now focusable and supports keyboard resizing (`ArrowLeft` / `ArrowRight`, `Home`, `End`), while still handling pointer/mouse drag.
- I left the `vertical_bar_width: 0` behavior unchanged in this PR. Right now `0` is treated as the backward-compatibility sentinel for older configs where the field is absent, and tightening that distinction cleanly would need a config-loading change rather than just a validator tweak. I did not want to mix that migration decision into this UI PR.

Local verification on this branch:

- `make check`

Current GitHub checks were restarted by the follow-up push and are pending as of this comment.

tomo-chan (commented):
## Code Review (follow-up — commit f794b0c6)

### Previous concerns

- ✅ **Per-pixel state updates** — addressed. The local `previewVerticalBarWidth` state now buffers visual updates during drag, and `onVerticalBarWidthChange` is only called on commit. The design intent is clear.
- ✅ **Mouse-only resizer** — addressed. Keyboard support (`ArrowLeft`/`ArrowRight`/`Home`/`End`) and pointer events are added. `aria-valuenow`/`aria-valuemin`/`aria-valuemax`/`tabIndex={0}` make the separator accessible.
- 🔄 **`VerticalBarWidth == 0` skips validation** — still unresolved (no change in `validate.go`). Carrying forward.

---

### Issues & Questions

**[Question] Effect cleanup calls `stopResizing` on every drag pixel, ending the drag after the first render**

`stopResizing` is defined in the component body, closing over `previewVerticalBarWidth`. It is passed to the `useEffect` via closure and called explicitly in the cleanup:

```tsx
// WorkspaceTabs.tsx — inside useEffect cleanup
return () => {
  window.removeEventListener(...)
  // ...
  stopResizing()        // ← explicit call
}
```

Because `previewVerticalBarWidth` is in the dependency array, the effect re-runs every time `setPreviewVerticalBarWidth` is called — i.e., on every pixel of the drag. Each re-run first executes the old cleanup, which calls the **old** `stopResizing` (closed over `previewVerticalBarWidth = null` from before the drag started):

```
stopResizing() (old closure: previewVerticalBarWidth = null)
  → resizeStateRef.current = null   ← drag terminated
  → commitPreviewWidth()            ← previewVerticalBarWidth is null → early return ✓
  → setPreviewVerticalBarWidth(null) ← bar reverts to original width
```

After this cleanup the drag reference is null, so every subsequent `mousemove` → `updatePreviewWidth` → `if (\!current) return` is a no-op. The drag effectively ends on the first render frame.

The `App.test.tsx` change in this commit is circumstantial evidence of the bug: the previous mouse-drag test (`mouseDown → mouseMove → mouseUp → expect(320)`) was replaced with a keyboard test (`keyDown('ArrowRight') → expect(292)`), and `WorkspaceTabs.test.tsx` was not updated to cover mouse dragging under the new implementation.

A fix would be to move the committed-width tracking out of state into a ref, so `stopResizing` can read the final preview width without needing to be in the deps array:

```tsx
const previewWidthRef = useRef<number | null>(null)
// update ref (not state) on every move, update state too for visual refresh
// cleanup only removes listeners — does NOT call stopResizing()
// actual mouseup handler (not cleanup) calls onVerticalBarWidthChange(previewWidthRef.current)
```

Is this breakage confirmed, or am I missing something about how the effect cleanup interacts with the ref in this environment?

**[Question] Duplicate event listeners on `window` and `document`**

The effect registers the same handlers on both targets:

```tsx
window.addEventListener('pointermove', handlePointerMove)
window.addEventListener('pointerup',    stopResizing)
window.addEventListener('mousemove',   handleMouseMove)
window.addEventListener('mouseup',     stopResizing)
document.addEventListener('pointermove', handlePointerMove)
document.addEventListener('pointerup',    stopResizing)
document.addEventListener('mousemove',   handleMouseMove)
document.addEventListener('mouseup',     stopResizing)
```

Pointer/mouse events bubble from `document` to `window`, so each event fires the handler twice. The `if (\!resizeStateRef.current) return` guard in `stopResizing` makes the second call a no-op, but `handlePointerMove` / `handleMouseMove` would call `updatePreviewWidth` (and thus `setPreviewVerticalBarWidth`) twice per move event. Was `document` added intentionally for a specific browser scenario, or is this an oversight?

---

### Overall

The keyboard-resize path looks solid and is well-tested. The mouse-drag path has a stale-closure issue in the effect that likely causes it to end on the first render frame, and the `App.test.tsx` change from a drag test to a keyboard test is consistent with that. Please verify mouse dragging in a real browser before merging.

tomo-chan:
Follow-up on the second review:

- I fixed the drag lifecycle issue in <https://github.com/tomo-chan/panemux/commit/de4dc1382f67d6814e83bfc8a15ac1eee84e88a9>. The in-progress width is now tracked in a ref alongside the visual state, and effect cleanup only removes listeners. The actual commit happens only from the real `pointerup` / `mouseup` path, so the drag no longer terminates on the first render.
- I also removed the duplicate `document` listeners in the same commit. The resize handlers now bind only on `window`, so move/up events are no longer processed twice.
- I restored explicit mouse-drag regression coverage in both `WorkspaceTabs.test.tsx` and `App.test.tsx`, in addition to the keyboard-resize coverage added earlier.

Local verification on this branch:

- `make check`

GitHub checks were restarted by the follow-up push. `gh pr checks` did not have the new branch run indexed yet at the time of this comment, so those checks may still appear as pending/starting up momentarily.

tomo-chan (commented):
## Code Review (follow-up — commit de4dc138)

### Previous concerns

- ✅ **Effect cleanup terminating drag** — fixed cleanly. `previewVerticalBarWidthRef` is kept in sync with `previewVerticalBarWidth` state via `setPreviewWidth`, and `stopResizing` reads from the ref rather than the stale closure. Effect cleanup now only removes listeners — no more `stopResizing()` call there. `previewVerticalBarWidth` is also removed from the effect deps, so the effect no longer re-runs on every pixel. The mouse-drag regression tests are restored in both `App.test.tsx` and `WorkspaceTabs.test.tsx`.
- ✅ **Duplicate window + document listeners** — removed. Only `window` listeners remain.

---

### Remaining open item

**[Minor question] `VerticalBarWidth == 0` still skips validation in `validate.go`**

This was raised in the first review and has not been addressed. A user who manually writes `vertical_bar_width: 0` in their config passes validation and silently gets 280 at runtime. The zero-skip exists for backward compatibility (old files without the field deserialize to Go's zero value), but it means a deliberate zero is treated the same as an absent field. No change needed before merge if this is intentional — just confirming the decision.

---

### Overall

The drag lifecycle fix is correct and well-reasoned: the ref carries the committed value across renders so the effect closure stays stable, and the effect teardown no longer double-commits or kills the drag mid-flight. Mouse drag and keyboard resize both have test coverage. The only open item is the minor validation question above. Everything else is clean.

tomo-chan:
On the remaining `vertical_bar_width: 0` question:

- Yes, the current behavior is intentional in this PR. `0` is still acting as the compatibility sentinel for “field absent in an older config”, and `normalizedWorkspaces()` translates that legacy zero value to the default width of `280`.
- I did not tighten `validate.go` here because rejecting an explicit `0` cleanly would need us to distinguish “missing field” from “present but zero” at config load time, rather than only at validation time. In the current shape, changing the validator alone would risk turning older configs into load failures.
- I agree that treating an explicit `vertical_bar_width: 0` the same as an absent field is not ideal, but I see that as a separate config-format/migration decision rather than a blocker for this UI change.

So for this PR the decision is: keep the compatibility behavior as-is, and defer any stricter absent-vs-zero handling to a dedicated follow-up if we want to change the config contract later.

Current status:

- local verification already ran via `make check`
- current GitHub checks are green

tomo-chan (commented):
## Code Review (final — all concerns resolved)

### On `vertical_bar_width: 0` skipping validation

The author's explanation is well-reasoned: `vertical_bar_width int` with `yaml:\"...,omitempty\"` means an absent field and an explicit `0` are indistinguishable after deserialization — both arrive as Go's zero value. Tightening the validator to reject `0` without first teaching the loader to distinguish the two cases (e.g. via `*int` or a presence flag) would turn unmodified older configs into load failures, which is a worse tradeoff than the current silent normalization to `280`. This is the right call for this PR's scope.

---

### Summary

All raised concerns have been addressed or explained:

| Concern | Status |
|---|---|
| Per-pixel state updates through `useLayout` on drag | ✅ Local `previewVerticalBarWidth` state + ref |
| Mouse-only resizer | ✅ Keyboard support + pointer events added |
| Effect cleanup terminating drag on first render | ✅ Ref approach; cleanup is listener-only |
| Duplicate `window` + `document` listeners | ✅ `document` listeners removed |
| `VerticalBarWidth == 0` validation skip | ✅ Intentional — backward-compat sentinel; clean fix needs loader-level change |

The implementation is correct, the drag lifecycle is sound, and both mouse and keyboard paths have test coverage. Ready to merge.

PR #143, opened 2026-06-24
ci: add renovate for scheduled dependency updates

tomo-chan:
## Summary
- Upgrade `golangci-lint` to `v2.12.2`.
- Add Renovate scheduling for dependency updates, including the `golangci-lint` pin in `Makefile`.
- Document the scheduled dependency-update policy in `docs/maintenance.md`.

## Test Plan
- `make install-deps`
- `make check`

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended"],
+  "timezone": "Asia/Tokyo",
+  "schedule": ["before 9am on Monday"],
+  "packageRules": [
+    {
+      "matchManagers": ["github-actions"],
+      "groupName": "GitHub Actions",
+      "pinDigests": true,
+      "automerge": false
+    }
+  ],
+  "customManagers": [
+    {
+      "customType": "regex",
+      "description": "Keep the pinned golangci-lint tool version in sync.",
+      "managerFilePatterns": ["/(^|/)Makefile$/"],
+      "matchStrings": [
+        "GOLANGCI_LINT_VERSION := (?<currentValue>v(?<version>\\d+\\.\\d+\\.\\d+))"
+      ],
+      "depNameTemplate": "github.com/golangci/golangci-lint/v2",
+      "datasourceTemplate": "go"
+    }
+  ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "schedule": ["before 9am on the first day of the month"]
+  }
+}
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -17,7 +17,7 @@
       "description": "Keep the pinned golangci-lint tool version in sync.",
       "managerFilePatterns": ["/(^|/)Makefile$/"],
       "matchStrings": [
-        "GOLANGCI_LINT_VERSION := (?<currentValue>v(?<version>\\d+\\.\\d+\\.\\d+))"
+        "GOLANGCI_LINT_VERSION := (?<currentValue>v\\d+\\.\\d+\\.\\d+)"
       ],
       "depNameTemplate": "github.com/golangci/golangci-lint/v2",
       "datasourceTemplate": "go"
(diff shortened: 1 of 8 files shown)


.github/renovate.json
```diff
@@ -0,0 +1,30 @@
...
+    }
+  ],
+  "customManagers": [
+    {
+      "customType": "regex",
+      "description": "Keep the pinned golangci-lint tool version in sync.",
+      "managerFilePatterns": ["/(^|/)Makefile$/"],
+      "matchStrings": [
+        "GOLANGCI_LINT_VERSION := (?<currentValue>v(?<version>\\d+\\.\\d+\\.\\d+))"
+      ],
```
tomo-chan, line 21: The inner `version` named group (`v(?<version>\d+\.\d+\.\d+)`) is not consumed by Renovate — only `currentValue` is read. If the intent is to document the structure of the semver portion, a note in the `description` field above would be clearer. If you intended to use it as a `versionTemplate` (to strip the `v` for display), add `"versioningTemplate": "semver"` and wire up `extractVersionTemplate`. As-is, the extra group is harmless but slightly misleading.

tomo-chan (commented):
## Code Review

### What's Good

- **Renovate custom manager is well-configured.** Using `datasourceTemplate: "go"` with `depNameTemplate: "github.com/golangci/golangci-lint/v2"` is the right call — the Go module proxy is the authoritative source since the tool is installed via `go install`, and `currentValue` captures the full `v`-prefixed semver that the proxy expects.
- **Version comparison in `lint-go-deps` is correct.** golangci-lint v2 emits versions _without_ the `v` prefix (e.g., `golangci-lint has version 2.12.2 built…`), so stripping `v` from `GOLANGCI_LINT_VERSION` with `sed` makes both sides of the comparison land on the same form. The `|| true` guard handles the "binary absent" case cleanly.
- **Constant extraction is clean and consistent.** Extracting `responseErrorKey`, the config defaults, `interactiveAgentCodex/Claude`, and `controlMessageTypeResize/Replay` all follow the same pattern — magic strings that form part of a protocol or config contract are now single-definition constants. Easy to spot if any ever need to change.
- **Documentation is complete.** Both `DEVELOPMENT.md` and `docs/maintenance.md` are updated in the same commit, covering the `lint-go-deps` behavior change and the full Renovate scheduling policy so nothing is left as implicit knowledge.

---

### Issues & Questions

**[Minor question] Unused `version` capture group in `matchStrings` regex**

```json
"GOLANGCI_LINT_VERSION := (?<currentValue>v(?<version>\\d+\\.\\d+\\.\\d+))"
```

Renovate only consumes `currentValue` (and a small set of other [named groups](https://docs.renovatebot.com/modules/manager/regex/)). The inner `version` group capturing `\d+\.\d+\.\d+` is never read. If it's there as a readability aid — to make the structure of `currentValue` visually explicit — a comment in `description` might be a cleaner home for that intent. If it was intended to be used as `versionTemplate` or similar, it may need a matching `versioningTemplate` entry.

**[Minor question] `lockFileMaintenance` scope for Go modules**

The monthly `lockFileMaintenance` run will refresh `go.sum` entries for transitive dependencies. In Go, this is most meaningful for projects with many indirect deps that aren't otherwise pinned — for panemux, which has a modest dependency tree, the monthly PRs may largely be no-ops or security-patch bumps. Is there a plan to automerge lock-file-only PRs once they've proven low-risk (e.g., after a quarter of clean Renovate runs), or keep them on the manual-review track permanently?

---

### Overall

Solid PR. The Renovate setup covers all four dependency surfaces (Go modules, npm, GitHub Actions, and the Makefile tool pin) with correct scheduling and sensible defaults. The constant extractions are a clean code-hygiene pass with no behavioral risk. Nothing blocks merge.

tomo-chan:
Updated in [a9aedea](https://github.com/tomo-chan/panemux/commit/a9aedea): the Renovate regex no longer captures an unused `version` group; it now only matches `currentValue`.

I also plan to keep `lockFileMaintenance` on manual review for now. The current policy in `docs/maintenance.md` is no automerge unless we change it later, and I’m fine leaving monthly lockfile PRs on that track until we have evidence they’re noise-free.

tomo-chan (commented):
## Code Review (follow-up)

Both concerns from the previous review are resolved.

### What's Good

- **Unused `version` capture group removed** (`a9aedea`). The regex is now the minimal form Renovate needs — only `currentValue` is captured, which is exactly what the `go` datasource consumes. Clean fix.
- **`lockFileMaintenance` policy is documented** and the author's reply confirms manual review will remain the policy for now, consistent with the `docs/maintenance.md` entry ("keep automerge disabled unless project policy changes").

---

### Overall

No open concerns. The PR is ready to merge.

PR #145, opened 2026-06-26
feat: preserve scroll and resize through mouse tracking mode

tomo-chan:
## Summary

- **Shift+wheel for local scroll**: When a remote app (vim, tmux, etc.) enables mouse tracking mode, xterm.js routes wheel events to SSH instead of scrolling the local viewport. With high SSH latency this makes scrolling appear completely frozen. A new `wheel` event handler (capture phase) in `TerminalPane` intercepts `Shift+wheel` and calls `term.scrollLines()` directly, bypassing SSH forwarding regardless of mouse tracking state.

- **Suppress mouse events during resize drag**: `mousemove` events during `SplitDivider` drags pass through the terminal area, where xterm.js (in mouse tracking mode) forwards them as escape sequences to SSH. This congests the SSH write path and delays the `WindowChange` PTY resize request. Setting `pointer-events: none` on terminal containers while a drag is active (`isResizing` state on `LayoutActionsContext`) stops this forwarding entirely for the duration of the drag.

- **Stabilize `handleResize`**: `handleResize` previously closed over the React `connected` state, causing `useCallback` to return a new reference on every WebSocket reconnect. Because `TerminalPane`'s `ResizeObserver` effect depends on `handleResize`, this tore down and rebuilt the observer on every reconnect — creating a window where drag-resize and maximize events could be missed. Moving `connected` to a ref makes `handleResize` stable for the component lifetime.

## Test plan

- [x] `make test-frontend` passes (405 tests; 2 new: `handleResize reference is stable when connected changes`, `scrollLines calls term.scrollLines`)
- [x] `make lint-frontend` passes (TypeScript strict check)
- [x] `make test-go` passes for all `./internal/...` packages
- [x] `make lint-go` passes (0 issues)
- [x] `make check` passes end-to-end via pre-push hook
- [ ] Manual: ssh_tmux session with vim/tmux mouse mode — normal wheel sends to remote; Shift+wheel scrolls local viewport
- [ ] Manual: pane resize drag while mouse tracking active — terminal receives no spurious mousemove events during drag; PTY resize arrives sooner
- [ ] Manual: maximize button — terminal immediately resizes to fill workspace even during active WebSocket reconnects

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "panemux-frontend",
-  "version": "0.18.8",
+  "version": "0.20.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "panemux-frontend",
-      "version": "0.18.8",
+      "version": "0.20.0",
       "dependencies": {
         "@xterm/addon-fit": "^0.10.0",
         "@xterm/addon-web-links": "^0.11.0",
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -39,6 +39,7 @@ export const App: React.FC = () => {
   } = useLayout()
   const [maximizedPaneIdsByWorkspace, setMaximizedPaneIdsByWorkspace] = useState<Record<string, string | null>>({})
   const [dragSourcePaneId, setDragSourcePaneId] = useState<string | null>(null)
+  const [isResizing, setIsResizing] = useState(false)
   const [activePaneId, setActivePaneId] = useState<string | null>(null)
   const [attentionPaneIds, setAttentionPaneIds] = useState<Set<string>>(() => new Set())
   const { isOpen, currentPane, sshConnectionNames, saveError, isSaving, openSettings, closeSettings, saveSettings, addSSHConfigHost, detectShell, browseDirectories } =
@@ -305,6 +306,8 @@ export const App: React.FC = () => {
       maximizedPaneId,
       dragSourcePaneId,
       setDragSourcePaneId,
+      isResizing,
+      setIsResizing,
       displayConfig: displayConfig ?? DEFAULT_DISPLAY,
       onPaneAttention: notifyAttention,
       clearPaneAttention,
--- a/frontend/src/components/SplitContainer.test.tsx
+++ b/frontend/src/components/SplitContainer.test.tsx
@@ -31,6 +31,8 @@ function makeCtx(maximizedPaneId: string | null): LayoutActionsContextValue {
     maximizedPaneId,
     dragSourcePaneId: null,
     setDragSourcePaneId: vi.fn(),
+    isResizing: false,
+    setIsResizing: vi.fn(),
     displayConfig: { show_header: false, show_status_bar: false },
     onPaneAttention: vi.fn(),
     clearPaneAttention: vi.fn(),
--- a/frontend/src/components/SplitContainer.tsx
+++ b/frontend/src/components/SplitContainer.tsx
@@ -20,6 +20,8 @@ export interface LayoutActionsContextValue {
   maximizedPaneId: string | null
   dragSourcePaneId: string | null
   setDragSourcePaneId: (id: string | null) => void
+  isResizing: boolean
+  setIsResizing: (v: boolean) => void
   displayConfig: DisplayConfig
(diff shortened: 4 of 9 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **`connectedRef` pattern is correctly applied.** Moving `connected` out of `handleResize`'s deps list into a ref (with a synchronising `useEffect`) is idiomatic React and cleanly eliminates the ResizeObserver rebuild on every WebSocket reconnect. The test that asserts `handleResize` ref identity across open→close lifecycle explicitly validates the fix.
- **`scrollLines` has correct empty deps.** It reads only from `entryRef.current` (a ref), so the callback is stable for the lifetime of the component with no closure over mutable state.
- **`pointer-events: none` is a clean CSS-level fix.** No extra JS event listeners to manage; the property is driven directly from `isResizing` state, making it easy to reason about.
- **Capture-phase + `passive: false` combo is correct.** `{ capture: true, passive: false }` is exactly what is needed to both intercept the event before xterm.js's target-phase listener and to allow `preventDefault()`. The cleanup correctly mirrors the registration with `removeEventListener(..., true)`.
- **`onResizeStart/End` ref pattern in `SplitDivider` is consistent.** Using refs for the callbacks keeps `handleMouseDown` stable (only `isHorizontal` in deps) and avoids re-registering document-level `mousemove`/`mouseup` listeners on every render.

---

### Issues & Questions

**[Question] `deltaY / 40` assumes `DOM_DELTA_PIXEL` — may silently break on Firefox**

```tsx
const lines = Math.round(e.deltaY / 40) || (e.deltaY > 0 ? 1 : -1)
```

`WheelEvent.deltaMode` controls the unit of `deltaY`:
- `DOM_DELTA_PIXEL` (0) — pixels; dividing by ~40 is a reasonable heuristic.
- `DOM_DELTA_LINE` (1) — already in lines; Firefox on Linux/Windows often delivers this mode with `deltaY` values like `3`. `Math.round(3 / 40) === 0`, so the fallback always returns exactly `1`, making every Shift+wheel tick scroll exactly one line regardless of speed.
- `DOM_DELTA_PAGE` (2) — full pages; dividing by 40 gives a tiny fraction that rounds to 0.

The safest fix branches on `e.deltaMode`:

```tsx
const rawLines =
  e.deltaMode === 1 /* DOM_DELTA_LINE */
    ? Math.round(e.deltaY)
    : e.deltaMode === 2 /* DOM_DELTA_PAGE */
    ? Math.round(e.deltaY * 20)
    : Math.round(e.deltaY / 40)
const lines = rawLines || (e.deltaY > 0 ? 1 : -1)
```

Is this intentionally limited to pixel mode for now, or should we cover the other modes?

---

### Overall

The three-part fix is well-structured: each concern (scroll, drag noise, observer stability) is addressed at the right layer. The one open item is the `deltaMode` assumption in the wheel handler — it may already work fine for the target use-cases (Chromium + macOS trackpad), but worth confirming before merge.

tomo-chan (commented):
Good catch. Fixed in 86f5e5b.

Branched on `e.deltaMode`:
- `DOM_DELTA_LINE` (1): use `Math.round(e.deltaY)` directly
- `DOM_DELTA_PAGE` (2): `Math.round(e.deltaY * 20)`
- `DOM_DELTA_PIXEL` (0): `Math.round(e.deltaY / 40)` (unchanged)

Added component-level tests covering all three modes and the no-Shift no-op.

tomo-chan (commented):
## Code Review (follow-up)

### Resolved

**[Question] `deltaY / 40` assumes `DOM_DELTA_PIXEL`** — addressed in `86f5e5b`. The branching on `e.deltaMode` is correct:
- `DOM_DELTA_LINE` (1): `Math.round(e.deltaY)` — `deltaY` is already in lines, no divisor needed.
- `DOM_DELTA_PAGE` (2): `Math.round(e.deltaY * 20)` — reasonable page-to-line conversion.
- `DOM_DELTA_PIXEL` (0): `Math.round(e.deltaY / 40)` — unchanged from before.

The six new component-level tests in `TerminalPane.test.tsx` cover every branch including the pixel-rounding fallback (`deltaY: 10` → `0` → `1`) and the no-Shift guard. Good test coverage.

---

### Overall

All concerns from the previous round are resolved. The PR is clean — no outstanding questions.

PR #146, opened 2026-07-02
fix: stop ssh/ssh+tmux panes from getting stuck reconnecting forever

tomo-chan:
## Summary

Unstable ssh/ssh+tmux connections could leave a pane showing "reconnecting..." permanently, with `/git-info 404`s and `/restart 500`s recurring in the logs, and SSH connections frequently emitting a transient "no address associated with hostname" DNS error. Three compounding causes were found and fixed:

- **`RestartSession` orphaned panes on failure.** It removed a pane's session from the manager *before* attempting to recreate it, so a transient SSH dial failure (e.g. a momentary DNS blip) left the pane permanently orphaned until some later restart happened to succeed — explaining the recurring `/git-info` 404s and the pane's WebSocket 404ing forever. Now the replacement session is created first; the old session is only swapped out after success.
- **`dialSSHClient` made only one dial attempt.** A transient DNS error failed the whole connection (and thus `/restart`) immediately. The transport-dial step (not the SSH handshake/auth step) now retries a bounded number of times with a short backoff.
- **`useWebSocket`'s reconnect loop gave up silently.** After exhausting its fixed-count, fixed-delay retry budget, it left `connected=false` forever with no signal, so a pane could show "reconnecting..." indefinitely with no way to reach the manual restart button. It now uses exponential backoff and exposes an `exhausted` state that `useTerminal` routes into the existing disconnected-session recovery path (the same one-shot auto-restart + manual "Reconnect Session" button used for backend-reported disconnects).

The previously-reported workaround (opening the WorkingDirectory Browse dialog twice, then Save) worked because Save triggers `/restart` as a side effect — the user was unknowingly retrying the SSH dial until a transient blip cleared. No SSH connection caching/pooling exists anywhere in the codebase; that was a red herring.

## Test plan

- [x] `make check` passes (lint, tests, coverage) on the branch
- [x] New Go tests: `internal/api/handler_test.go` (`RestartSession` no longer orphans the session on `createSession` failure) and `internal/session/ssh_test.go` (dial retry succeeds/exhausts/respects budget; handshake failures are never retried)
- [x] New frontend tests: `useWebSocket.test.ts` (exponential backoff, `exhausted` state) and `useTerminal.test.ts` (WS exhaustion drives the same recovery path as a backend-reported disconnect, including the regression case where no status frame ever arrives)
- [x] Manually verified end-to-end: built the binary, pointed a pane at an unreachable SSH host, confirmed `/restart` returns 500 without removing the pane's prior session, confirmed `/git-info` keeps returning 200 instead of 404, and confirmed in the browser that ending a local shell surfaces the "Restart Session" button which successfully reconnects
- [x] `docs/behavior.md` updated to document `/restart`'s recreate-in-place semantics and the WS-exhaustion recovery path

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -46,6 +46,8 @@ type Handler struct {
 	readDirFn               func(name string) ([]os.DirEntry, error)
 	preferredCWDMu          sync.Mutex
 	preferredCWDBySession   map[string]preferredCWDState
+	restartMu               sync.Mutex
+	restartInFlight         map[string]struct{}
 }
 
 type preferredCWDState struct {
@@ -126,6 +128,7 @@ func NewHandler(cfg *config.Config, manager *session.Manager) *Handler {
 		manager:               manager,
 		sshConfigPath:         sshconfig.DefaultPath(),
 		preferredCWDBySession: make(map[string]preferredCWDState),
+		restartInFlight:       make(map[string]struct{}),
 	}
 	h.createSession = session.CreateFromConfig
 	h.detectLocalShellFn = session.DetectLocalShell
@@ -410,6 +413,18 @@ func (h *Handler) DeleteSession(w http.ResponseWriter, r *http.Request) {
 // RestartSession recreates a session from its original config.
 func (h *Handler) RestartSession(w http.ResponseWriter, r *http.Request) {
 	id := chi.URLParam(r, "id")
+
+	// Creating the replacement session happens outside the manager's lock (it
+	// can block on a real SSH dial), so two concurrent restarts for the same
+	// pane would otherwise each build their own session independently; the
+	// one that finishes last would win the manager swap while the other's
+	// session is silently discarded unused. Serialize per-id instead of
+	// letting that race play out.
+	if !h.beginRestart(id) {
+		http.Error(w, "a restart is already in progress for this session", http.StatusConflict)
+		return
+	}
+	defer h.endRestart(id)
 
 	var found *config.PaneConfig
 	for _, p := range h.cfg.AllPanes() {
@@ -902,6 +917,27 @@ func (h *Handler) clearPreferredCWD(sessionID string) {
 	defer h.preferredCWDMu.Unlock()
 
 	delete(h.preferredCWDBySession, sessionID)
+}
+
+// beginRestart claims the per-id restart guard, returning false if a restart
+// for id is already in progress.
+func (h *Handler) beginRestart(id string) bool {
+	h.restartMu.Lock()
+	defer h.restartMu.Unlock()
+
+	if _, ok := h.restartInFlight[id]; ok {
+		return false
+	}
+	h.restartInFlight[id] = struct{}{}
+	return true
+}
+
+// endRestart releases the per-id restart guard claimed by beginRestart.
+func (h *Handler) endRestart(id string) {
+	h.restartMu.Lock()
+	defer h.restartMu.Unlock()
+
+	delete(h.restartInFlight, id)
 }
 
 func (h *Handler) gitInfoLogScope(sess session.Session) string {
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -423,14 +423,18 @@ func (h *Handler) RestartSession(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	h.manager.Remove(id) //nolint:errcheck // ok if already gone
-	h.clearPreferredCWD(id)
-
+	// Create the replacement session before touching the manager. If this fails
+	// (e.g. a transient SSH dial error), the existing session for id stays
+	// registered instead of being orphaned, so /ws and /git-info keep working
+	// against it and the frontend's disconnected-status recovery path can retry.
 	sess, err := h.createSession(found, h.cfg.SSHConnections)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	h.manager.Remove(id) //nolint:errcheck // ok if already gone
+	h.clearPreferredCWD(id)
 	h.manager.Add(sess)
 	w.WriteHeader(http.StatusOK)
 }
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -92,6 +92,14 @@ func resolveKnownHostsFile(knownHostsFile string) (string, error) {
 // and ProxyCommand. Returns (client, jumpClient, error). jumpClient is non-nil only when
 // a ProxyJump is used; the caller must close jumpClient after closing client.
 func dialSSHClient(cfg SSHConfig) (*ssh.Client, *ssh.Client, error) {
+	return dialSSHClientUntil(cfg, nowFn().Add(dialRetryBudget))
+}
+
+// dialSSHClientUntil is dialSSHClient with an explicit retry deadline. A
+// ProxyJump hop shares the outer call's deadline (see dialThroughJump) rather
+// than starting a fresh dialRetryBudget window, so a multi-hop chain's total
+// retry time stays bounded by a single budget instead of multiplying per hop.
+func dialSSHClientUntil(cfg SSHConfig, deadline time.Time) (*ssh.Client, *ssh.Client, error) {
 	authMethods, err := buildAuthMethods(cfg)
 	if err != nil {
 		return nil, nil, err
@@ -116,7 +124,7 @@ func dialSSHClient(cfg SSHConfig) (*ssh.Client, *ssh.Client, error) {
 		Timeout:           30 * time.Second,
 	}
 
-	conn, jumpClient, err := dialTransportWithRetry(cfg, addr, port)
+	conn, jumpClient, err := dialTransportWithRetry(cfg, addr, port, deadline)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -143,10 +151,12 @@ const (
 	// dialRetryInitialBackoff is the delay before the second attempt; it
 	// doubles on each subsequent attempt (300ms, 600ms).
 	dialRetryInitialBackoff = 300 * time.Millisecond
-	// dialRetryBudget caps the total wall-clock time spent retrying at the
-	// same ceiling as a single dial attempt's own timeout, so a caller
-	// synchronously waiting on dialSSHClient (e.g. the /restart HTTP handler)
-	// never waits dramatically longer than it already tolerates today.
+	// dialRetryBudget caps the total wall-clock time spent retrying, shared
+	// across an entire ProxyJump chain via a single deadline (see
+	// dialSSHClientUntil), so a caller synchronously waiting on dialSSHClient
+	// (e.g. the /restart HTTP handler) never waits dramatically longer than
+	// the ceiling a single dial attempt already tolerated before retries were
+	// introduced.
 	dialRetryBudget = 30 * time.Second
 )
 
@@ -160,10 +170,10 @@ var (
 // dialTransportFn is the transport-dial step, injectable for tests. Assigned
 // in init (rather than at the var declaration) to avoid a spurious Go
 // initialization-cycle error: dialTransport transitively calls back into
-// dialTransportFn through dialThroughJump -> dialSSHClient, and Go's
+// dialTransportFn through dialThroughJump -> dialSSHClientUntil, and Go's
 // init-order analysis does not distinguish "referenced in a function body"
 // from "referenced at initialization time".
-var dialTransportFn func(cfg SSHConfig, addr string, port int) (net.Conn, *ssh.Client, error)
+var dialTransportFn func(cfg SSHConfig, addr string, port int, deadline time.Time) (net.Conn, *ssh.Client, error)
 
 func init() {
 	dialTransportFn = dialTransport
@@ -171,26 +181,37 @@ func init() {
 
 // dialTransport establishes the raw transport (TCP, ProxyJump, or
 // ProxyCommand) for an SSH connection, without performing the SSH handshake.
-func dialTransport(cfg SSHConfig, addr string, port int) (net.Conn, *ssh.Client, error) {
+// The direct-dial timeout is clamped to whatever of deadline remains, so a
+// retried attempt can never itself run past the shared retry budget.
+func dialTransport(cfg SSHConfig, addr string, port int, deadline time.Time) (net.Conn, *ssh.Client, error) {
 	switch {
 	case cfg.JumpHost != nil:
-		return dialThroughJump(*cfg.JumpHost, addr)
+		return dialThroughJump(*cfg.JumpHost, addr, deadline)
 	case cfg.ProxyCommand != "":
 		conn, err := dialViaProxyCommand(cfg.ProxyCommand, cfg.Host, port)
 		return conn, nil, err
 	default:
-		conn, err := net.DialTimeout("tcp", addr, 30*time.Second)
+		timeout := deadline.Sub(nowFn())
+		if timeout <= 0 {
+			return nil, nil, fmt.Errorf("dial %s: retry budget exhausted", addr)
+		}
+		conn, err := net.DialTimeout("tcp", addr, timeout)
 		return conn, nil, err
 	}
 }
 
 // dialTransportWithRetry retries dialTransport with a short backoff on
-// failure, bounded by dialRetryMaxAttempts and dialRetryBudget. Only the
-// transport step is retried; a successful transport dial followed by an SSH
+// failure, bounded by dialRetryMaxAttempts and deadline. Only the transport
+// step is retried; a successful transport dial followed by an SSH
 // handshake/auth failure is never retried by this function since that
-// happens in a separate step in dialSSHClient.
-func dialTransportWithRetry(cfg SSHConfig, addr string, port int) (net.Conn, *ssh.Client, error) {
-	start := nowFn()
+// happens in a separate step in dialSSHClientUntil.
+//
+// Each attempt's own timeout is clamped to the time remaining until deadline
+// (see dialTransport), so a slow/hanging attempt that consumes the whole
+// budget leaves no room for further attempts — the retry loop's own
+// before-each-attempt deadline check alone cannot do this, since it can only
+// observe elapsed time between attempts, not interrupt one already in flight.
+func dialTransportWithRetry(cfg SSHConfig, addr string, port int, deadline time.Time) (net.Conn, *ssh.Client, error) {
 	backoff := dialRetryInitialBackoff
 
 	var conn net.Conn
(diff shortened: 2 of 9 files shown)


internal/session/ssh.go
```diff
@@ -143,6 +133,84 @@ func dialSSHClient(cfg SSHConfig) (*ssh.Client, *ssh.Client, error) {
...
+	var jumpClient *ssh.Client
+	var err error
+
+	for attempt := 1; attempt <= dialRetryMaxAttempts; attempt++ {
+		conn, jumpClient, err = dialTransportFn(cfg, addr, port)
+		if err == nil {
+			return conn, jumpClient, nil
+		}
+		if attempt == dialRetryMaxAttempts || nowFn().Sub(start) >= dialRetryBudget {
+			break
```
tomo-chan, line 242: **[Question] `dialRetryBudget` doesn't bound total wall-clock time for slow-failing (as opposed to fast-failing) attempts.**

This check only runs *between* attempts, so it can't interrupt an attempt already in flight. A single transport-dial attempt can itself take up to 30s (`net.DialTimeout`, or the jump chain's own 30s SSH timeout). For a target that fails by hanging (e.g. firewall silently dropping SYN packets) rather than failing fast (DNS NXDOMAIN, the case this PR targets), all 3 attempts could each take ~30s, for a worst case of ~90s before `/restart` returns 500 — 3x the previous ~30s ceiling. See the review body for details.

internal/session/ssh.go
```diff
@@ -143,6 +133,84 @@ func dialSSHClient(cfg SSHConfig) (*ssh.Client, *ssh.Client, error) {
...
+	start := nowFn()
+	backoff := dialRetryInitialBackoff
+
+	var conn net.Conn
+	var jumpClient *ssh.Client
+	var err error
+
+	for attempt := 1; attempt <= dialRetryMaxAttempts; attempt++ {
+		conn, jumpClient, err = dialTransportFn(cfg, addr, port)
+		if err == nil {
```
tomo-chan, line 238: **[Minor question] Retries on any transport error, not just DNS-shaped ones.**

This retries on any error from `dialTransportFn`, not only DNS-resolution failures. It will also retry (adding ~900ms of backoff) on errors unlikely to be transient, e.g. "connection refused" or a deterministically-failing `ProxyCommand`. Was narrowing to DNS-shaped errors (e.g. `*net.DNSError`) considered?

internal/api/handler.go
```diff
@@ -423,14 +423,18 @@ func (h *Handler) RestartSession(w http.ResponseWriter, r *http.Request) {
...
+	// registered instead of being orphaned, so /ws and /git-info keep working
+	// against it and the frontend's disconnected-status recovery path can retry.
 	sess, err := h.createSession(found, h.cfg.SSHConnections)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	h.manager.Remove(id) //nolint:errcheck // ok if already gone
+	h.clearPreferredCWD(id)
```
tomo-chan, line 452: **[Minor question] Concurrent `/restart` requests for the same pane.**

Not a regression from this PR, but worth a sanity check given the reordering: two concurrent restarts for the same id each call `createSession` independently; the "loser" wins nothing and its freshly-created session gets closed via the winner's `Remove(id)` before ever being used. Worth a per-id guard, or out of scope?

tomo-chan (commented):
## Code Review

### What's Good

- Clear root-cause diagnosis tying together three independent, compounding bugs (orphaning on `/restart` failure, single-shot SSH dial, silent WS reconnect exhaustion) rather than papering over the symptom.
- `RestartSession` now creates the replacement session before touching the manager, so a failed create no longer orphans the pane — and it's backed by four focused new tests (`handler_test.go`) covering the failure path, `preferredCWD` preservation, the 500 body, and the no-prior-session case.
- `useWebSocket`'s new `exhausted` flag is wired into `useTerminal` through the *existing* disconnected-session recovery path instead of adding a parallel one — good reuse.
- `TestDialSSHClient_HandshakeFailure_NotRetried` uses a real `net.Listener` instead of a mock to prove handshake failures fail fast and are never retried — a meaningfully stronger test than asserting call counts against a stub would have been.
- `docs/behavior.md` updated in the same PR to describe the new `/restart` semantics and WS-exhaustion recovery path, per the repo's documentation rule.

---

### Issues & Questions

**[Question] `dialRetryBudget` doesn't bound total wall-clock time for slow-failing (as opposed to fast-failing) attempts**

```go
for attempt := 1; attempt <= dialRetryMaxAttempts; attempt++ {
    conn, jumpClient, err = dialTransportFn(cfg, addr, port)
    if err == nil {
        return conn, jumpClient, nil
    }
    if attempt == dialRetryMaxAttempts || nowFn().Sub(start) >= dialRetryBudget {
        break
    }
    sleepFn(backoff)
    backoff *= 2
}
```

The comment on `dialRetryBudget` says it caps "the total wall-clock time spent retrying at the same ceiling as a single dial attempt's own timeout, so a caller synchronously waiting on `dialSSHClient` ... never waits dramatically longer than it already tolerates today." But the budget check only runs *between* attempts — it can't interrupt an attempt already in flight. A single transport-dial attempt can itself take up to 30s (`net.DialTimeout(..., 30*time.Second)`, or the jump chain's own 30s SSH timeout). For a target that fails by *hanging* rather than failing fast — e.g. a firewall silently dropping SYN packets, rather than the DNS NXDOMAIN case this PR targets — all `dialRetryMaxAttempts` (3) attempts could each take close to 30s, giving a worst case of ~90s before `/restart` returns 500, three times the previous ~30s ceiling.

The existing retry tests (`TestDialWithRetry_StopsRetryingPastElapsedBudget` etc.) only exercise instantly-returning mocked errors, so this slow-attempt scenario isn't covered either.

Is a ~90s worst case for hung TCP connections acceptable here, or should the loop also check whether there's enough remaining budget to justify *starting* another attempt (e.g. skip the retry if `elapsed >= dialRetryBudget - expectedAttemptTimeout`)?

**[Minor question] Retry loop doesn't distinguish transient from permanent transport errors**

`dialTransportWithRetry` retries on *any* error from `dialTransportFn`, not just DNS-resolution failures. The PR description frames this fix around a specific "no address associated with hostname" DNS blip, but as written it will also retry (adding ~900ms of backoff) on errors that are very unlikely to be transient — e.g. "connection refused" from a closed port, or a misconfigured `ProxyCommand` that fails deterministically every time. Was narrowing the retry to DNS-shaped errors (e.g. `*net.DNSError`) considered, or is blanket retry-on-any-transport-error intentional given the modest added latency?

**[Minor question] Concurrent `/restart` requests for the same pane**

Not a regression introduced by this PR, but the reordering makes it worth a sanity check: two concurrent `POST /restart` calls for the same `id` each call `createSession` independently; whichever finishes last wins the `Remove`+`Add` swap, and the "loser's" freshly-created session gets closed via the winner's `Remove(id)` before ever being used by anyone. Since this PR is specifically about not losing sessions on `/restart`, is a per-id guard (e.g. a mutex keyed on session id) worth adding here, or is double-restart considered out of scope?

---

### Overall

Solid, well-tested fix for all three root causes, and the reasoning connecting them is clearly laid out in the PR description. The main thing I'd like confirmed before merge is the `dialRetryBudget` behavior for slow/hanging transport failures (as opposed to the fast-failing DNS case it was designed for) — the comment's stated guarantee doesn't hold in that case. The other two questions are minor and don't need to block merge.

tomo-chan:
Thanks for the review — addressed in 348ed79:

**`dialRetryBudget` worst-case for slow/hanging attempts** — confirmed this was a real gap, not just theoretical: the elapsed-time check only ran *between* attempts using a fixed `start` reference, so (a) a ProxyJump hop called `dialSSHClient` directly, giving it its own fresh 30s budget independent of the outer call — worst case multiplied per hop — and (b) even without a jump host, attempts that each finish just under the budget (rather than hanging for the full 30s) could push cumulative time well past the intended ceiling before the between-attempts check caught up.

Fixed by threading a single absolute `deadline` through `dialSSHClient` → `dialTransportWithRetry` → `dialTransport` → `dialThroughJump` (which now reuses the *same* deadline for the jump hop instead of computing a fresh one), and clamping each attempt's own dial timeout to `deadline.Sub(nowFn())` rather than a fixed 30s. A hanging attempt that consumes the whole budget now leaves no room for further attempts, and a ProxyJump chain's total retry time is bounded by one shared budget instead of multiplying per hop. Added `TestDialWithRetry_HangingFirstAttemptConsumesWholeBudget`, `TestDialWithRetry_ShrinksTimeoutForSubsequentAttempts`, and `TestDialTransport_JumpHostSharesRetryDeadlineWithOuterCall` to cover this.

**Retry on any transport error vs. DNS-shaped only** — left as blanket retry-on-any-transport-error. Narrowing to `*net.DNSError` would miss other plausibly-transient cases this fix is also meant to help with (e.g. `connection refused` during a brief sshd restart, `network is unreachable` during a NIC flap), and the now-bounded worst-case latency (~30s ceiling, same as before this PR, rather than multiplying) makes retrying a genuinely-permanent error like a misconfigured `ProxyCommand` cheap enough not to worry about.

**Concurrent `/restart` for the same pane** — added a per-id in-flight guard; a second concurrent restart for the same pane now gets `409` instead of racing to build its own session and potentially having it silently discarded when the other one wins the manager swap. See `TestRestartSession_ConcurrentRequests_SecondReturns409` and `TestRestartSession_GuardReleasedAfterCompletion`, and the updated `/restart` docs in `docs/behavior.md`.

internal/session/ssh.go
```diff
@@ -143,11 +141,110 @@ func dialSSHClient(cfg SSHConfig) (*ssh.Client, *ssh.Client, error) {
...
+// ProxyCommand) for an SSH connection, without performing the SSH handshake.
+// The direct-dial timeout is clamped to whatever of deadline remains, so a
+// retried attempt can never itself run past the shared retry budget.
+func dialTransport(cfg SSHConfig, addr string, port int, deadline time.Time) (net.Conn, *ssh.Client, error) {
+	switch {
+	case cfg.JumpHost != nil:
+		return dialThroughJump(*cfg.JumpHost, addr, deadline)
+	case cfg.ProxyCommand != "":
+		conn, err := dialViaProxyCommand(cfg.ProxyCommand, cfg.Host, port)
+		return conn, nil, err
```
tomo-chan, line 201: **[Suggestion] `ProxyCommand` transport dial still has no timeout/deadline enforcement.**

Unlike the default `net.DialTimeout` branch (now clamped to `deadline.Sub(nowFn())`), this branch never consults `deadline`. Not a regression — the original code had no timeout here either — but since this PR specifically hardens this path against hanging/slow dials, a hanging `ProxyCommand` would still block indefinitely, unbounded by `dialRetryBudget`. Worth a follow-up issue if `ProxyCommand` connections are commonly used here?

tomo-chan (commented):
## Code Review

### What's Good

- All three concerns from the previous review round are resolved with real engineering, not just acknowledgment:
  - The `dialRetryBudget` gap is fixed by threading a single absolute `deadline` through `dialSSHClientUntil` → `dialTransportWithRetry` → `dialTransport` → `dialThroughJump`, with each attempt's own dial timeout now clamped to whatever of the deadline remains. `TestDialWithRetry_HangingFirstAttemptConsumesWholeBudget` and `TestDialWithRetry_ShrinksTimeoutForSubsequentAttempts` directly exercise the two failure modes (a single hanging attempt, and cumulative near-budget attempts) that the previous "between-attempts-only" check missed.
  - The ProxyJump case now reuses the *same* deadline for the jump hop instead of computing a fresh one (`dialThroughJump` takes `deadline` and calls `dialSSHClientUntil(jumpCfg, deadline)`), closing the "multiplies per hop" gap — verified by `TestDialTransport_JumpHostSharesRetryDeadlineWithOuterCall`.
  - The concurrent-`/restart` race now returns `409` via a per-id in-flight guard (`beginRestart`/`endRestart`), verified by a genuinely concurrent test (`TestRestartSession_ConcurrentRequests_SecondReturns409`, using channels to force real overlap rather than just asserting on call counts) plus a guard-release test (`TestRestartSession_GuardReleasedAfterCompletion`).
- The decision to keep blanket retry-on-any-transport-error (rather than narrowing to DNS-shaped errors) is now backed by a clear justification in the reply thread — bounded worst-case latency makes retrying a permanent error cheap, and other plausibly-transient cases (`connection refused` during an sshd restart, a brief `network unreachable`) benefit too. Agreed this is a reasonable tradeoff.
- `docs/behavior.md` updated again to describe the new `409` response and the shared-deadline retry semantics.

---

### Issues & Questions

**[Suggestion] `ProxyCommand` transport dial still has no timeout/deadline enforcement**

In `dialTransport` (`internal/session/ssh.go`), the `ProxyCommand` branch calls `dialViaProxyCommand(cfg.ProxyCommand, cfg.Host, port)` without ever consulting `deadline` — unlike the default `net.DialTimeout` branch, which now clamps to `deadline.Sub(nowFn())`. This isn't a regression (the original code had no timeout here either), but since this PR is specifically hardening this code path against hanging/slow dials, a `ProxyCommand` that itself hangs (e.g. an unreachable bastion referenced in the command) would still block indefinitely, unbounded by `dialRetryBudget`. Worth filing a follow-up issue if `ProxyCommand`-based connections are a supported/common configuration here?

---

### Overall

Nice follow-up — both review rounds' concerns are addressed with dedicated regression tests rather than just point tweaks, and the reply comment is a good record of why the design landed where it did. Nothing here blocks merge; the `ProxyCommand` note is optional and out of scope for this PR.

tomo-chan:
Good catch, and worth being precise about where the actual hang risk sits.

Looking at `dialViaProxyCommand`, `c.Start()` itself is effectively non-blocking (it just forks `/bin/sh -c <cmd>` and returns immediately) — it's not going to hang the way `net.DialTimeout` on an unreachable host does, so there isn't really a "dial" moment to attach `deadline` to here the way the direct-TCP branch has. The actual hang risk is downstream: if the `ProxyCommand` subprocess (e.g. `ssh -W %h:%p bastion`) is itself slow to establish its own tunnel, that shows up as slow reads/writes on `proxyCommandConn`'s pipes once `ssh.NewClientConn` starts the handshake — i.e. it's a handshake-phase hang, not a dial-phase one.

That puts it on the same side of the line as the SSH handshake/auth step itself, which this PR deliberately leaves unbounded/unretried for *every* transport (not just ProxyCommand) — the whole design here is "retry the dial, never retry or rebound the handshake," so a hanging TCP-established-but-slow-handshake host has the same unbounded-wait characteristic today as a hanging ProxyCommand subprocess would. Fixing this specifically for ProxyCommand would mean adding real `SetReadDeadline`/`SetWriteDeadline` support to `proxyCommandConn` (currently no-ops) and deciding how that interacts with the handshake's own timeout — which is a meaningfully different, larger change than this PR's scope (retrying a flaky *dial*), and would want to cover the direct-TCP and jump-host handshake paths too for consistency, not just ProxyCommand.

Given you already flagged this as optional/non-blocking, I'm leaving it out of this PR and would rather file it as a separate follow-up ("bound SSH handshake time across all transport types, not just the dial step") so it gets designed and tested on its own rather than bolted onto this one. Let me know if you'd rather I open that issue now.

internal/session/ssh.go
tomo-chan: Thanks for the detailed explanation — one correction to the "TCP/jump handshake is already bounded the same way" framing, though it doesn't change your conclusion.

I checked `golang.org/x/crypto/ssh@v0.45.0` (the version this repo vendors): `ClientConfig.Timeout` is documented as "the maximum amount of time for the TCP connection to establish," and grepping the package, it's referenced in exactly one place — inside `ssh.Dial()`, where it's passed to `net.DialTimeout`. `ssh.NewClientConn` (which is what `dialSSHClientUntil` actually calls here, since panemux always dials its own `net.Conn` first) never reads `config.Timeout` at all, and `clientHandshake` never calls `SetDeadline`/`SetReadDeadline`/`SetWriteDeadline` on the connection.

So the `Timeout: 30 * time.Second` set on `sshCfg` in `dialSSHClientUntil` is dead for this codebase's call pattern — it has no effect on the handshake for *any* transport, not just `ProxyCommand`. The direct-TCP and jump-host handshake paths aren't actually bounded today either; they just happen not to hang in practice because a real `net.Conn`'s OS-level TCP behavior (RSTs, etc.) tends to surface handshake-phase failures quickly, whereas `proxyCommandConn`'s `SetReadDeadline`/`SetWriteDeadline` are no-ops by construction, so a stuck subprocess pipe genuinely never times out.

This doesn't change your scoping call — it's still pre-existing, still out of scope for a dial-retry PR, and your proposed follow-up ("bound SSH handshake time across all transport types, not just the dial step") is exactly the right framing given this. Just flagging so the follow-up issue description doesn't understate the gap by assuming the TCP/jump paths already have a working 30s handshake bound — they don't; `sshCfg.Timeout` is currently inert.

tomo-chan: Good catch on the correction — verified against the vendored golang.org/x/crypto/ssh@v0.45.0: config.Timeout is read in exactly one place (client.go:178, inside ssh.Dial()), and clientHandshake never touches it or calls any SetDeadline variant. So yes, sshCfg.Timeout is dead for every transport this codebase uses, not just ProxyCommand — thanks for tracking that down precisely.

Filed the follow-up as #147, framed the way you suggested (bound handshake time across TCP/ProxyJump/ProxyCommand, not just ProxyCommand). Also pushed ff25365, which replaces the misleading bare Timeout: 30 * time.Second with a comment explaining exactly why it's currently inert here, so nobody reading this code in the future assumes the handshake has a working bound — didn't want to leave that lying around silently even though the real fix is out of scope for this PR.

tomo-chan (commented):
## Code Review

### What's Good

- `ff25365` closes out the last open thread cleanly: it's a docs/comment-only change (the `Timeout: 30 * time.Second` value is unchanged, so no functional behavior differs from `348ed79`), replacing a misleading bare timeout value with a comment explaining precisely why it's currently inert for every transport and pointing to the real fix (`#147`) instead of silently leaving a footgun for future readers.
- Follow-up issue [#147](https://github.com/tomo-chan/panemux/issues/147) is filed with the corrected framing ("bound SSH handshake time across all transport types, not just ProxyCommand"), so the gap doesn't get lost or under-scoped.

---

### Issues & Questions

None remaining. All findings from the previous two review rounds (the `dialRetryBudget` worst-case latency gap, the ProxyJump per-hop budget multiplication, the concurrent-`/restart` race, and the `ProxyCommand`/handshake-timeout clarification) are resolved with code changes, dedicated regression tests, or — for the one genuinely out-of-scope item — a correctly-scoped follow-up issue.

---

### Overall

Nothing left to flag. This PR is in good shape from a review standpoint.

PR #149, opened 2026-07-06
fix: detect worktrees from Claude subagent transcripts, show multiple PRs

tomo-chan:
## Summary
- panemux only read a Claude Code session's top-level `<sessionId>.jsonl` transcript to detect the active git worktree, so worktree/PR info in the pane header silently disappeared whenever the actual work happened in a delegated Task subagent instead (subagent activity is recorded in separate `subagents/*.jsonl` files that panemux never read).
- Extend `internal/session/local.go` and `ssh.go` to also read every sibling subagent transcript for a Claude session, returning all distinct workdirs signaled (no recency/time-window filtering — all worktrees found in the session are candidates).
- Extend `internal/api/handler.go` to resolve and show every distinct active sibling worktree instead of just one, with per-worktree GitHub PR lookups run concurrently.
- Extend the pane header (`PaneHeader.tsx`) to show a single worktree inline as before, or collapse into a `<N> worktrees` trigger + popover menu when 2 or more are active.
- Update `docs/behavior.md` and `docs/ui-design.md` to describe the new resolution and display behavior.

## Test plan
- [x] `make test-go` — all Go tests pass, including new coverage for subagent transcript discovery (local + SSH), multi-worktree resolution/dedup, and concurrent PR lookups (partial failure handled).
- [x] `make test-frontend` — all frontend tests pass, including new `GitInfoSchema` worktree-array cases and `PaneHeader` menu open/close/Escape/backdrop-click cases.
- [x] `make lint` — clean.
- [x] `make coverage` — Go 86.1%, frontend 95.25% (both above the 80% gate).
- [x] `make build` — succeeds.
- [x] Manual verification: built the binary, ran it against an isolated config pointed at this worktree, and confirmed in a real browser that the single-worktree case renders unchanged, and the multi-worktree case (via a mocked `git-info` response) shows the `2 worktrees` trigger, opens a menu listing both worktrees with their PR links, and closes on Escape.

### DIFF
--- a/docs/behavior.md
+++ b/docs/behavior.md
@@ -560,6 +560,8 @@ When a pane moves to a different parent node, the component may be remounted by 
 - Claude session metadata is used only to identify the matching transcript (`sessionId`) and project directory key.
 - Claude transcript resolution prefers the latest top-level `cwd` recorded on transcript entries such as `user`, `assistant`, `attachment`, and `system`.
 - When a Claude transcript entry does not expose top-level `cwd`, panemux falls back to the latest non-auxiliary tool file path, then the latest `Bash` tool `cd ... &&` target, then file-history snapshot paths.
+- Claude Code also records delegated Task subagent activity in separate transcript files under `<sessionId>/subagents/*.jsonl`, next to the parent transcript. A subagent that does worktree-relative work there never updates the parent transcript's own `cwd`, so panemux additionally reads every subagent transcript file and resolves each one with the same rule above, independently of the parent.
+- panemux does not apply any recency or time-window filter when considering subagent transcripts: every distinct worktree signaled by the parent transcript or any subagent transcript for the current session is a candidate, regardless of how long ago that transcript file was last written.
 - This resolver intentionally does not depend on Claude `hooks` or `statusLine` configuration, so no extra Claude-side setup is required for pane-header Git or PR detection.
 - If the agent exits, no longer has an eligible worktree, or the resolved worktree is not a sibling worktree of the pane's current repository, panemux falls back to the pane's own working directory immediately.
 - For SSH panes, panemux resolves interactive agent processes from the remote process list on the current SSH connection.
@@ -567,6 +569,8 @@ When a pane moves to a different parent node, the component may be remounted by 
 - If the resolved repository origin can be converted into a browser URL, the header shows the repository name as a link to that repository page.
 - For SCP-style SSH origins such as `git@alias:owner/repo.git`, panemux uses `~/.ssh/config` `Host` aliases to resolve both the browser link hostname and GitHub PR lookup repo host when a matching alias exists; otherwise it treats the SSH host token as the hostname directly.
 - If the resolved branch has a GitHub pull request, the header shows a PR link labeled `#<number>`.
+- When the active agent (including its subagents) has diverged into more than one distinct sibling worktree of the same repository, panemux shows all of them instead of just one, deduplicated by the worktree's repository root; the pane's own base directory is shown only when nothing has diverged from it. Each distinct worktree gets its own independent GitHub PR lookup, so more than one PR link can be shown at once. See [ui-design.md](ui-design.md) for how the header presents more than one worktree.
+- The "last known worktree" sticky behavior applies per distinct worktree: if the active-workdir lookup transiently fails or returns nothing, panemux keeps showing the previously resolved set of worktrees until a subsequent lookup confirms they are no longer valid (e.g. the branch changed or the worktree was removed).
 
 ## Operational Assumptions
 
--- a/docs/ui-design.md
+++ b/docs/ui-design.md
@@ -131,6 +131,12 @@ When panemux can derive a repository page URL from the pane's Git origin, the re
 
 The PR shortcut is rendered as an inline text link labeled `#<number>`. It does not use a pill, badge, or outlined capsule treatment. On hover, the link shifts to a slightly stronger blue and shows an underline so it reads like the rest of the header chrome instead of a separate button system.
 
+### Multiple worktrees
+
+When the pane's active agent has diverged into exactly one sibling worktree, the header shows that worktree's repo, branch, and PR shortcut inline, exactly as described above.
+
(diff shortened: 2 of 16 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- **Verified against real Claude Code on-disk layout.** I checked an actual `~/.claude/projects/<project>/<sessionId>/subagents/*.jsonl` directory tree on this machine, and both the file location (`<sessionId>/subagents/*.jsonl` next to the parent transcript) and the top-level `cwd` field on transcript entries match exactly what `claudeTranscriptPaths`/`claudeSessionCWDs` (local.go) and `remoteClaudeSubagentsDirShellPath`/`remoteClaudeSessionCWDs` (ssh.go) assume. This is the load-bearing assumption for the whole feature and it holds up.
- **Concurrency is correct.** `lookupPRInfoForContexts` writes to `results[i]` from per-goroutine index `i` with no shared mutable state, so there's no data race. I ran `go test -race ./internal/api/... ./internal/session/...` on the branch and everything passes, including the new `TestGetGitInfo_MultipleActiveWorktrees_*` tests that exercise the partial-failure and dedup paths.
- **Backward/graceful degradation is well covered**: `TestGetActiveWorkdirs_NoSubagentsDirectory_BackwardCompatible` and the SSH `ls ... 2>/dev/null || true` pattern both confirm older Claude Code sessions without a `subagents` directory don't regress.
- **Sticky-worktree generalization is clean.** Reworking `preferredCWDBySession` from a single `preferredCWDState` to `[]preferredCWDState` and keeping the recall/remember/clear logic symmetric per-root is a nice way to extend the existing single-worktree "sticky" behavior to N worktrees without duplicating the fallback logic.
- Docs (`behavior.md`, `ui-design.md`) were updated in the same change and accurately describe the implementation (dedup by root, per-worktree PR lookup, no recency filter, sticky-per-worktree).
- Frontend test coverage for `PaneHeader` (single vs. collapsed-menu rendering, open/close via Escape and backdrop click) is thorough; I ran `vitest run` on `PaneHeader.test.tsx` and `schemas/index.test.ts` on the branch and all 83 tests pass.

---

### Issues & Questions

**[Minor question] `WorktreeMenu`'s `role="menu"` has no `menuitem` children or arrow-key navigation**

`frontend/src/components/PaneHeader.tsx` — the popover renders `role="menu"` with plain `<span>`/`WorktreeEntryLine` children (no `role="menuitem"`, no roving tabindex/arrow-key handling). Screen readers announcing this as a "menu" without menu items is a minor mismatch with the ARIA authoring practices for the menu pattern. Given this is a small, infrequently-used popover, this may be an acceptable simplification — was that a conscious scope decision, or worth a quick follow-up to switch to a plain list (`role="list"`/no role) if full menu semantics aren't intended?

**[Minor question] Slightly redundant top-level-field OR condition in `PaneHeader.tsx`**

```tsx
{gitInfo?.is_git && (gitInfo.repo || gitInfo.branch || (gitInfo.worktrees?.length ?? 0) > 0) && (
```

Since the backend (`GetGitInfo` in `handler.go`) always mirrors `worktrees[0]` into the top-level `Branch`/`Repo` fields whenever `Worktrees` is non-empty, the `(gitInfo.worktrees?.length ?? 0) > 0` clause looks unreachable in practice — any real response with a non-empty `worktrees` array will also have `repo` or `branch` set. Is this purely defensive for hand-built/mocked `gitInfo` payloads in tests, or is there a real backend path where `worktrees` is populated but both `repo` and `branch` are empty?

---

### Overall

I checked out the branch, ran `go test -race` on `internal/api` and `internal/session` (pass), ran the frontend `vitest` suites for the touched files (83/83 pass), confirmed `gofmt`/`go vet` are clean, and independently verified the core assumption about Claude Code's `subagents/*.jsonl` transcript layout against real data on this machine — it's correct. CI is green. This looks ready to merge; the two notes above are minor and non-blocking.

tomo-chan:
Addressed both review notes in 760635a:

- **`role="menu"` mismatch**: not a conscious "menu" choice — switched the popover to `role="group"` with `aria-label="Active worktrees"` (and dropped `aria-haspopup="true"` from the trigger) since the entries are plain links, not `menuitem` actions with arrow-key navigation. A plain labeled group is the correct semantics here.
- **Redundant `worktrees`-array OR clause**: confirmed it was dead in practice — `GetGitInfo` always mirrors `worktrees[0]` into the top-level `branch`/`repo` fields whenever `Worktrees` is non-empty, so a real response can never have a populated `worktrees` array with empty `repo`/`branch`. Removed the clause and updated the four test fixtures that had omitted the top-level fields to match the real backend contract instead of working around it defensively.

`make check` is green on the updated branch.

tomo-chan (commented):
## Code Review

### What's Good

Both notes from the previous review round were addressed in `760635a`:

- **`role="menu"` mismatch**: switched to `role="group"` with `aria-label="Active worktrees"` and dropped `aria-haspopup="true"`, matching the actual content (plain links, not `menuitem` actions with arrow-key navigation). `aria-expanded` on the trigger is still appropriate here since it's a generic disclosure-widget state attribute, not menu-specific.
- **Redundant `worktrees`-array OR clause**: removed, and the four `PaneHeader.test.tsx` fixtures that previously omitted the top-level `repo`/`branch`/`pr_*` fields were updated to include them, matching the real backend contract (`GetGitInfo` always mirrors `worktrees[0]` into the top-level fields) instead of relying on a defensive branch that could never be exercised in production.

I checked out `760635a` in a separate worktree and ran `vitest run` on `PaneHeader.test.tsx` and `schemas/index.test.ts` — all 83 tests pass. CI (`Analyze (go)`, `Analyze (javascript-typescript)`, `CodeQL`, `label`, `validate`) is green; `ci` was still running at review time but the change here is frontend-only and narrowly scoped, consistent with the passing suites I ran locally.

---

### Issues & Questions

None outstanding — no new issues introduced by this fix-up commit.

---

### Overall

Both previous notes are resolved correctly and verified locally. No blocking concerns; good to merge once the pending `ci` check finishes.

PR #151, opened 2026-07-09
fix: keep pane header worktree/PR info fresh after claude --resume

tomo-chan:
## Summary
- Pane headers only refreshed on click/keydown/focus/visibility-regain, so resuming a Claude Code session with `claude --resume` and then watching the agent work autonomously (no interaction with that pane) left the worktree/PR info stuck on stale data indefinitely.
- Add the steady-state 10-second poll already documented in `docs/behavior.md` ("refreshed every 10 seconds only while both the browser tab and that pane remain visible") to `useGitInfo`, so the header eventually catches up with zero interaction.
- Add a 30-second server-side TTL cache for `GET /api/sessions/{id}/git-info` (`internal/api/handler.go`) so the new steady-state polling — and any other concurrent viewer of the same session — doesn't repeat process/transcript scanning, remote git inspection over SSH, or `gh pr view` lookups on every request. Applies uniformly to local/tmux/ssh/ssh_tmux sessions. Cache entries are cleared on session delete/recreate so a new session never inherits stale data.
- Documented the new caching behavior and its bounded-staleness trade-off in `docs/behavior.md`.

## Test plan
- [x] `make test-go`
- [x] `make test-frontend`
- [x] `make lint`
- [x] `make check` (including the pre-push hook's own `make check` run)
- [x] New automated tests directly simulate the reported scenario: `frontend/src/hooks/useGitInfo.test.ts` asserts the pane header keeps refetching every 10s with zero pane interaction (and stops when disabled/hidden); `internal/api/handler_test.go` asserts a second request within 30s is served from the server cache without re-invoking `GetCWD`, that the cache expires and recomputes after 30s, and that a deleted/recreated session under the same ID never serves the old session's cached response.
- [x] Manual end-to-end verification: ran `go run .` and the Vite dev server against a real local session in a scratch git repo, opened it in a browser, and confirmed via network logs that `/api/sessions/{id}/git-info` is now called automatically roughly every 10 seconds with zero clicks/keystrokes on the pane, where previously it would never refire on its own.

### DIFF
--- a/frontend/src/hooks/useGitInfo.ts
+++ b/frontend/src/hooks/useGitInfo.ts
@@ -3,6 +3,7 @@ import { GitInfo, GitInfoSchema } from '../schemas'
 
 const GIT_INFO_STALE_MS = 30000
 const GIT_INFO_RECHECK_THROTTLE_MS = 5000
+const GIT_INFO_POLL_INTERVAL_MS = 10000
 
 interface GitInfoCacheEntry {
   gitInfo: GitInfo
@@ -85,6 +86,20 @@ export function useGitInfo(sessionId: string, enabled = true): UseGitInfoResult 
     if (!enabled || !isVisible) return
     void refreshIfStale()
   }, [enabled, isVisible, refreshIfStale])
+
+  // Steady-state poll: keeps the pane header current even when nothing
+  // interacts with this specific pane (e.g. an agent working autonomously
+  // after `claude --resume`). The backend caches its own expensive work
+  // (remote git/PR lookups) on a longer TTL, so it's safe to poll it
+  // unconditionally here rather than relying on the interaction-driven
+  // staleness throttle in refreshIfStale.
+  useEffect(() => {
+    if (!enabled || !isVisible) return
+    const interval = setInterval(() => {
+      void refreshNow()
+    }, GIT_INFO_POLL_INTERVAL_MS)
+    return () => clearInterval(interval)
+  }, [enabled, isVisible, refreshNow])
 
   useEffect(() => {
     setGitInfo(getOrCreateCacheEntry(sessionId).gitInfo)
--- a/frontend/src/hooks/useGitInfo.ts
+++ b/frontend/src/hooks/useGitInfo.ts
@@ -1,7 +1,15 @@
 import { useState, useEffect, useCallback } from 'react'
 import { GitInfo, GitInfoSchema } from '../schemas'
 
-const GIT_INFO_STALE_MS = 30000
+// Matches GIT_INFO_POLL_INTERVAL_MS: since the steady-state poll already
+// keeps the cache within 10s of fresh while a pane is visible, an
+// interaction-triggered refresh (click/focus/keydown, visibility regain)
+// should not tolerate data any older than that before treating it as stale.
+// Keeping this aligned with the poll interval also bounds the worst-case
+// staleness of an explicit refresh to this value plus the server-side cache
+// TTL, instead of compounding a larger, independent client-side window on
+// top of it.
+const GIT_INFO_STALE_MS = 10000
 const GIT_INFO_RECHECK_THROTTLE_MS = 5000
 const GIT_INFO_POLL_INTERVAL_MS = 10000
 
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -48,6 +48,9 @@ type Handler struct {
 	preferredCWDBySession   map[string][]preferredCWDState
 	restartMu               sync.Mutex
 	restartInFlight         map[string]struct{}
+	nowFn                   func() time.Time
+	gitInfoCacheMu          sync.Mutex
+	gitInfoCacheBySession   map[string]gitInfoCacheEntry
 }
 
 type preferredCWDState struct {
@@ -55,6 +58,22 @@ type preferredCWDState struct {
 	CommonDir string
 	Root      string
 }
+
+// gitInfoCacheEntry holds a previously computed git-info response for a
+// session, valid until expiresAt. Caching the whole response — not just the
+// remote git/PR lookups — keeps this simple and uniform across local, tmux,
+// ssh, and ssh_tmux sessions alike: local sessions also pay for process and
+// transcript scanning on every request, so they benefit from the same cap.
+type gitInfoCacheEntry struct {
+	expiresAt time.Time
+	response  gitInfoResponse
+}
+
+// gitInfoCacheTTL bounds how long a pane header may show git/PR metadata
+// that is no longer perfectly current, in exchange for not re-running
+// process/transcript scans, remote git inspection, and `gh pr view` lookups
+// on every poll. See docs/behavior.md "Pane Git and PR metadata".
+const gitInfoCacheTTL = 30 * time.Second
 
 // activeGitContext pairs a resolved git context with the working directory
 // (pane base cwd, or an agent-signaled sibling worktree path) it was
@@ -137,6 +156,8 @@ func NewHandler(cfg *config.Config, manager *session.Manager) *Handler {
 		sshConfigPath:         sshconfig.DefaultPath(),
 		preferredCWDBySession: make(map[string][]preferredCWDState),
 		restartInFlight:       make(map[string]struct{}),
+		nowFn:                 time.Now,
+		gitInfoCacheBySession: make(map[string]gitInfoCacheEntry),
 	}
 	h.createSession = session.CreateFromConfig
 	h.detectLocalShellFn = session.DetectLocalShell
@@ -274,6 +295,7 @@ func (h *Handler) DeleteWorkspace(w http.ResponseWriter, r *http.Request) {
 	for _, pane := range panesInLayout(workspace.Layout) {
 		_ = h.manager.Remove(pane.ID)
 		h.clearPreferredCWDs(pane.ID)
+		h.clearGitInfoCache(pane.ID)
 	}
 	writeJSON(w, h.cfg.WorkspacesView())
 }
@@ -410,6 +432,7 @@ func (h *Handler) DeleteSession(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	h.clearPreferredCWDs(id)
+	h.clearGitInfoCache(id)
 	h.cfg.RemovePaneFromLayout(id)
 	if err := h.cfg.SaveLayout(h.cfg.Layout); err != nil {
 		http.Error(w, "failed to save layout", http.StatusInternalServerError)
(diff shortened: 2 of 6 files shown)


frontend/src/hooks/useGitInfo.ts
```diff
@@ -86,6 +87,20 @@ export function useGitInfo(sessionId: string, enabled = true): UseGitInfoResult
...
+  // interacts with this specific pane (e.g. an agent working autonomously
+  // after `claude --resume`). The backend caches its own expensive work
+  // (remote git/PR lookups) on a longer TTL, so it's safe to poll it
+  // unconditionally here rather than relying on the interaction-driven
+  // staleness throttle in refreshIfStale.
+  useEffect(() => {
+    if (!enabled || !isVisible) return
+    const interval = setInterval(() => {
+      void refreshNow()
+    }, GIT_INFO_POLL_INTERVAL_MS)
```
tomo-chan, line 108: [Question] This poll now bypasses `refreshIfStale`'s throttle unconditionally every 10s, and combined with the new 30s server-side cache, explicit-refresh call sites (`handleOpenVSCode`'s `refreshNow()`, click/focus-triggered `refreshIfStale()` in TerminalPane.tsx) can now return data up to ~60s stale (30s client staleness + up to 30s server cache) instead of a fresh recompute. Was this compounding staleness considered acceptable for explicit user-triggered refreshes, or should those bypass the server cache?

internal/api/handler.go
```diff
@@ -798,42 +822,89 @@ func (h *Handler) GetGitInfo(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if resp, ok := h.cachedGitInfo(id); ok {
+		writeJSON(w, resp)
```
tomo-chan, line 826: [Suggestion] No single-flight protection here: if multiple viewers (two tabs, or the new steady-state poll racing a click-triggered refresh) hit this session right as the TTL expires, they can all trigger a concurrent `computeGitInfo` (including `gh pr view` network calls). Not a regression vs. pre-PR behavior, but a `singleflight.Group` keyed by session ID would collapse concurrent cache-misses into one computation — worth a follow-up issue rather than blocking this PR.

internal/api/handler_test.go
```diff
@@ -2583,6 +2593,121 @@ func TestGetGitInfo_GitNotFound_IsGitFalse(t *testing.T) {
...
+	req := httptest.NewRequest(http.MethodGet, "/api/sessions/local-recreated/git-info", nil)
+	r.ServeHTTP(rec, req)
+	assert.Equal(t, http.StatusOK, rec.Code)
+	var oldResp gitInfoResponse
+	require.NoError(t, json.NewDecoder(rec.Body).Decode(&oldResp))
+	assert.Equal(t, "main", oldResp.Branch)
+
+	require.NoError(t, mgr.Remove("local-recreated"))
+	h.clearGitInfoCache("local-recreated")
+
```
tomo-chan, line 2758: [Minor question] This calls `h.clearGitInfoCache(...)` directly rather than going through the real `DeleteSession` endpoint. The analogous `preferredCWDBySession` cache has `TestDeleteSession_ClearsPreferredCWD`/`TestRestartSession_ClearsPreferredCWD`, which assert clearing through the actual HTTP handlers. Worth adding the same through-the-endpoint assertion for `gitInfoCacheBySession` so a future handler refactor can't silently drop the `clearGitInfoCache` call without a test catching it?

tomo-chan (commented):
## Code Review

### What's Good

- `computeGitInfo` is a clean, behavior-preserving extraction from `GetGitInfo` — every early-return case (`IsGit: false`) and the final success payload map 1:1 to the pre-PR logic, so the cache layer is a pure addition rather than a rewrite risk.
- The new `nowFn` field follows the repo's existing injectable-clock convention (matching `Config.sshConfigPath`'s `DefaultPath()` override pattern called out in `DEVELOPMENT.md`), which is why the TTL tests can drive expiry deterministically without real sleeps.
- Cache invalidation is wired into all three session lifecycle events that can free up a session ID (`DeleteSession`, `RestartSession`, `DeleteWorkspace`), and `TestGetGitInfo_AfterSessionRecreatedWithSameID_DoesNotServeOldSessionsCache` specifically covers the trickiest case — a new session reusing an old session's ID must not inherit its cached git/PR data.
- The frontend test additions (`polls automatically every 10 seconds...`, `stops polling once the pane becomes disabled`, `stops polling while the browser tab is hidden...`) directly exercise the reported bug (zero-interaction staleness) plus the enabled/hidden state variants called out in `DEVELOPMENT.md`'s test-granularity guidance, rather than stopping at one happy-path test.
- `docs/behavior.md` is updated in the same change to describe the new bounded-staleness trade-off, per the "Documentation updates" rule in `DEVELOPMENT.md`.

---

### Issues & Questions

**[Question] Does the new server-side cache weaken explicit "refresh now" actions?**

Before this PR, every `/git-info` request recomputed fresh data. Now a request can be served from a cache entry up to `gitInfoCacheTTL` (30s) old. Two call sites in `TerminalPane.tsx` call `refreshNow()`/`refreshIfStale()` in response to explicit user actions rather than the steady-state poll:
- `handleOpenVSCode` calls `refreshNow()` immediately before opening VS Code — presumably to show the freshest state around an action tied to the current working directory.
- Mouse-down/focus handlers call `refreshIfStale()`, which itself only refetches once the *client-side* cache is >30s old — so in the worst case a click-triggered refresh can now return data that is up to ~60s old (30s client staleness + up to 30s server cache) instead of the previously-fresh recompute.

Was this compounding staleness window considered acceptable for explicit-refresh triggers, or should those call sites bypass the server cache (e.g. a `?fresh=true` param or a cache-bypass argument to `computeGitInfo`)? `docs/behavior.md`'s updated wording documents the trade-off for the general display case, but doesn't distinguish "background poll" from "user explicitly asked to refresh."

**[Minor question] `clearGitInfoCache` invalidation isn't exercised through the real HTTP handlers**

`TestGetGitInfo_AfterSessionRecreatedWithSameID_DoesNotServeOldSessionsCache` calls `h.clearGitInfoCache("local-recreated")` directly rather than going through `DeleteSession`. The analogous `preferredCWDBySession` cache has `TestDeleteSession_ClearsPreferredCWD` and `TestRestartSession_ClearsPreferredCWD`, which assert the clear-on-delete/restart behavior by hitting the actual `DELETE`/restart endpoints. Would it be worth adding the same lightweight through-the-endpoint assertions (checking `h.gitInfoCacheBySession[id]` is gone after a real `DELETE /api/sessions/{id}` and after `RestartSession`) so a future refactor that reorders the handler body can't silently drop the `clearGitInfoCache` call without a test noticing?

**[Suggestion] No single-flight protection on cache-miss**

`GetGitInfo` checks `cachedGitInfo`, and on a miss calls `computeGitInfo` unconditionally before `storeGitInfoCache`. If multiple viewers (e.g. two browser tabs, or the new steady-state poll racing a click-triggered refresh) hit the same session right as the TTL expires, they can all trigger a concurrent `computeGitInfo` — including the `gh pr view` network calls this cache is meant to reduce. This isn't a regression (every request did this before the PR), just an opportunity: a `singleflight.Group` keyed by session ID around `computeGitInfo` would collapse concurrent cache-misses into one computation. Worth a follow-up issue rather than blocking this PR.

---

### Overall

Solid, well-tested fix for the reported staleness bug, with a reasonable and clearly documented caching trade-off. The main thing I'd want confirmed before merge is whether the compounding staleness on explicit refresh actions (VS Code open, click/focus) is intentional — if a user expects "click to get current state" to mean *current*, the up-to-60s worst case might be worth tightening. The other two points are minor/non-blocking.

tomo-chan:
Addressed the review feedback:

- **Compounding staleness on explicit refresh actions**: lowered `GIT_INFO_STALE_MS` from 30s to 10s (matching `GIT_INFO_POLL_INTERVAL_MS`), so click/focus/keydown/tab-restore/VS-Code-open triggers now only skip a refetch if the client already has a response no older than 10s. Worst case for an explicit refresh is now 10s (client) + 30s (server cache) instead of 30s + 30s. Documented in `docs/behavior.md`. Went with tightening the client threshold rather than adding a cache-bypass parameter, since it's a smaller, more consistent change and the steady-state poll already keeps the client within 10s in practice.
- **Cache invalidation only exercised directly**: added `TestDeleteSession_ClearsGitInfoCache`, `TestRestartSession_ClearsGitInfoCache`, and `TestDeleteWorkspace_ClearsGitInfoCacheForRemovedPanes`, mirroring the existing `preferredCWDBySession` coverage for the same three call sites — these hit the real `DELETE`/restart/workspace-delete endpoints and assert the cache entry is gone, so a future refactor can't silently drop `clearGitInfoCache` without a test noticing.
- **Single-flight on cache-miss**: left as-is per your note that it's a non-blocking follow-up rather than something to fix in this PR.

`make check` passes locally (including via the pre-push hook).

tomo-chan (commented):
## Code Review (follow-up)

All three points from the previous review round are resolved in `daf2e582`:

### Issues & Questions — resolution status

**[Question] Compounding staleness on explicit refresh actions — Resolved**

`GIT_INFO_STALE_MS` is lowered from 30s to 10s (matching `GIT_INFO_POLL_INTERVAL_MS`), so click/focus/keydown/tab-restore/VS-Code-open triggers now only skip a refetch if the client already has a response no older than 10s. Worst case for an explicit refresh is now bounded at 10s (client) + 30s (server cache) = 40s instead of the previous 30s + 30s = 60s. The tightened threshold is explained inline in `useGitInfo.ts` and documented in `docs/behavior.md`. Tightening the client threshold rather than adding a cache-bypass parameter is a reasonable call — it's the smaller, more uniform change, and the steady-state poll already keeps the client within 10s in practice.

**[Minor question] `clearGitInfoCache` invalidation only exercised directly — Resolved**

`TestDeleteSession_ClearsGitInfoCache`, `TestRestartSession_ClearsGitInfoCache`, and `TestDeleteWorkspace_ClearsGitInfoCacheForRemovedPanes` now hit the real `DELETE`/restart/workspace-delete endpoints and assert the cache entry is gone, mirroring the existing `preferredCWDBySession` coverage for the same three call sites. This closes the gap where a future handler refactor could silently drop the `clearGitInfoCache` call without a test catching it.

**[Suggestion] No single-flight protection on cache-miss — Deferred, as expected**

Left as a non-blocking follow-up, consistent with my original note that this wasn't a regression and was worth tracking separately rather than fixing in this PR.

No new issues found in the incremental diff.

---

### Overall

All prior feedback has been addressed with targeted, well-tested changes. Nothing further needed from my side before merge.

PR #155, opened 2026-07-10
fix: detect Claude worktree switches via Bash cd, not only top-level cwd

tomo-chan:
## Summary
- `parseClaudeProjectCWD` preferred a Claude transcript record's top-level `cwd` field over a `Bash` tool call's `cd X && ...` target.
- In real Claude Code transcripts, that top-level `cwd` field always reflects the interactive process's own OS-level launch directory and never changes for its lifetime, so it is present on nearly every record. This permanently masked any sibling-worktree divergence reached via a plain Bash `cd` — the Bash-cd detection code already existed but was effectively unreachable in practice. Reproduced directly against a live transcript from this very session.
- Swap the priority so the Bash `cd` target wins over the top-level `cwd`, matching the precedence already used for Codex's `workdir` field (see `docs/architecture.md`). A generic file-touch path (`Read`/`Edit`/`Write`) remains a weaker signal than the top-level `cwd`, unchanged from before.
- This is not specific to `/resume` — it affects any Claude session (resumed or not) that uses a plain Bash `cd` (rather than a delegated Task subagent) to work in a sibling Git worktree. `ssh.go` needed no production changes since local and remote resolution share the same `parseClaudeProjectCWD` function; only its test fixture needed correcting since it encoded the same wrong priority.

## Test plan
- [x] `make test-go`
- [x] `make lint`
- [x] `make check` (including the pre-push hook's own `make check` run)
- [x] New test `TestParseClaudeProjectCWD_PrefersBashCDWorktreeOverTopLevelCWD` directly encodes the fix: a transcript with both a Bash `cd` to a sibling worktree and a top-level `cwd` pinned to the base directory now resolves to the worktree.
- [x] `TestActiveRemoteWorkdir_PrefersRemoteClaudeBashCDWorktreeOverTopLevelCWD` adds the same coverage for the SSH-backed resolver.
- [x] `TestActiveRemoteWorkdir_PrefersRemoteClaudeTopLevelCWDOverToolPaths` fixture corrected to use a `Read` file-touch instead of a `Bash cd` (it was previously asserting the buggy priority under a misleading name).
- [x] All existing `parseClaudeProjectCWD`/Claude transcript resolution tests still pass unchanged, confirming file-touch-vs-top-level-cwd precedence is untouched.


internal/session/local.go
```diff
@@ -750,28 +750,53 @@ func readClaudeProjectCWD(path string) (string, error) {
...
+			return
+		}
+		if candidate.IsDir {
+			latestBashCDPath = candidate.Path
+		} else {
+			latestFileTouchPath = candidate
 		}
 	})
+	if latestBashCDPath != "" {
+		return latestBashCDPath, nil
```
tomo-chan, line 787: [Question] latestBashCDPath is only ever overwritten by another Bash `cd`, never by a later top-level `cwd` or file-touch record. So once any Bash `cd` appears anywhere in the transcript, it wins indefinitely — even if the rest of the session's activity (other Bash calls without `cd`, or file touches) happens back in the original directory. Is that durable-signal behavior intentional (mirroring Codex's exec_command precedence), or should later cwd/file-touch signals be able to invalidate a stale cd after enough subsequent activity?

internal/session/local.go
```diff
@@ -750,28 +750,53 @@ func readClaudeProjectCWD(path string) (string, error) {
 	})
 }
 
+// parseClaudeProjectCWD resolves the effective working directory for a
+// Claude transcript, in priority order: the latest Bash `cd X && ...`
```
tomo-chan, line 754: [Minor question] This ~19-line comment restates most of the reasoning already written in docs/architecture.md and docs/behavior.md in this same PR. Intentional duplication (so the rationale is visible while reading the code), or would a shorter comment pointing to docs/architecture.md be preferable to avoid the two prose blocks drifting apart later?

tomo-chan (commented):
## Code Review

### What's Good

- Clear, well-substantiated root-cause narrative: the PR explains *why* the top-level `cwd` field was always winning (it's pinned at process launch and never updates), and states the bug was reproduced against a live transcript rather than just theorized.
- The fix is minimal and structural — splitting the single `latestPath` variable into `latestBashCDPath` and `latestFileTouchPath` cleanly separates the two signal strengths without adding new control flow complexity.
- Because `ssh.go`'s `activeRemoteWorkdirs` reuses the same shared `parseClaudeProjectCWD` (confirmed at `internal/session/ssh.go:1127`), no duplicate SSH-side logic needed touching — good use of the existing shared resolver instead of forking behavior.
- Test coverage is thorough and well-targeted: `TestParseClaudeProjectCWD_PrefersBashCDWorktreeOverTopLevelCWD` and its SSH counterpart directly encode the regression (both a Bash `cd` and a top-level `cwd` present, asserting the `cd` wins), and the renamed `TestActiveRemoteWorkdir_PrefersRemoteClaudeTopLevelCWDOverToolPaths` correctly keeps proving `cwd` still beats a mere file-touch. I ran these locally (`go test ./internal/session/... -run 'TestParseClaudeProjectCWD|TestActiveRemoteWorkdir'`) and all pass; `gofmt` and `go vet` are clean on the changed files.
- `docs/architecture.md` and `docs/behavior.md` were updated in the same change to reflect the new priority order, per this repo's documentation-updates rule.

---

### Issues & Questions

**[Question] Bash-`cd` signal persists for the rest of the transcript once set, regardless of later signals**

`internal/session/local.go` (new `parseClaudeProjectCWD`): `latestBashCDPath` is only ever overwritten by *another* Bash `cd`, never cleared or superseded by a later top-level `cwd` record or a later file-touch. Concretely: if a session does `cd ../worktree-a && ...` once early on, and the rest of the transcript is many turns of plain `Bash` calls (no `cd`) or `Read`/`Edit` calls back in the main tree, `parseClaudeProjectCWD` will keep resolving to `worktree-a` indefinitely, since neither `latestCWD` nor `latestFileTouchPath` can ever outrank a non-empty `latestBashCDPath`.

Is that the intended behavior (i.e., a Bash `cd` is treated as a durable "the agent has moved its base of operations here" signal, mirroring Codex's `exec_command workdir` precedence design mentioned in `docs/architecture.md`), or should a later plain file-touch/cwd record be able to invalidate a stale `cd` if enough time/records have passed? If it's intentional, it might be worth a one-line note in the doc comment or `docs/architecture.md` making that persistence explicit, since it's a slightly different property than "latest signal wins" — it's closer to "latest signal wins in its own tier, and top tier never decays."

**[Minor question] Function-doc comment on `parseClaudeProjectCWD` duplicates `docs/architecture.md`/`docs/behavior.md` prose almost verbatim**

The new ~19-line comment on `parseClaudeProjectCWD` (`internal/session/local.go`) restates most of the same reasoning already written in `docs/architecture.md` and `docs/behavior.md` in this same PR. Was this duplication intentional (e.g. so the rationale is visible without opening docs while reading the code), or would a shorter comment that just points to `docs/architecture.md` for the full rationale be preferable to avoid the two prose blocks drifting out of sync on a future edit?

---

### Overall

Solid, well-tested fix with a genuinely reproduced root cause and matching doc updates. Both open items are questions rather than blockers — the persistence behavior of the Bash-`cd` signal is the one worth a quick confirmation before merge, since it's a real (if likely intentional) behavioral property that isn't explicitly called out anywhere.

tomo-chan:
Addressed both review points in bbb3964:

- **Bash-cd persistence**: confirmed intentional — a Bash `cd` target is a durable "the agent moved its base of operations here" signal, mirroring Codex's `workdir` precedence, so it stays authoritative until a *later* `cd` replaces it. Made this explicit in `docs/architecture.md` and added `TestParseClaudeProjectCWD_BashCDPersistsOverLaterWeakerSignals` covering it directly.
- **Comment duplication**: shortened `parseClaudeProjectCWD`'s doc comment to point at `docs/architecture.md` for the full rationale instead of restating it inline, so the two don't drift independently.

`make check` passes locally with both commits applied.

tomo-chan (commented):
## Code Review

### What's Good

- Both open questions from the previous review round were addressed directly and substantively in `bbb3964`, not just brushed off:
  - **Bash-cd persistence**: confirmed intentional, with the reasoning now spelled out in `docs/architecture.md`'s "Pane Git/PR resolution" section — "an explicit `cd` is treated as a durable... signal, and the two weaker signals below it are not reliable enough evidence that the agent moved back to justify overriding it." That's a clear, reviewable design rationale rather than an implicit assumption.
  - **Comment duplication**: `parseClaudeProjectCWD`'s doc comment in `internal/session/local.go` now points to `docs/architecture.md`'s "Pane Git/PR resolution" section for the full rationale instead of restating it inline, and only documents the function-local priority order plus the persistence behavior. I verified that section heading exists and the referenced content matches.
- New test `TestParseClaudeProjectCWD_BashCDPersistsOverLaterWeakerSignals` directly encodes the persistence behavior: a Bash `cd` followed by a later record with both a pinned top-level `cwd` and an unrelated file-touch still resolves to the earlier `cd` target. This is exactly the right regression test for the behavior being clarified.
- I ran the full relevant test suite locally (`go test ./internal/session/... -run 'TestParseClaudeProjectCWD|TestActiveRemoteWorkdir' -v`) against `bbb3964` — all 27 tests pass, including the new one. `gofmt` and `go vet` are clean on the changed Go files.

---

### Issues & Questions

None remaining. Both prior findings are resolved by this commit.

---

### Overall

Both previous review points are resolved with matching doc updates, code comment fixes, and a new targeted test — nothing further from me. I re-ran the test suite against the latest commit (`bbb3964`) and confirmed it passes cleanly.

PR #157, opened 2026-07-15
fix: bound tmux/SSH/git exec calls with timeouts and add SSH keepalive

tomo-chan:
## Summary
- Pane metadata lookups (`GetCWD`, `GetActiveWorkdirs`, git-context inspection) ran tmux/ps/lsof/git commands and SSH exec-channel commands with no timeout, so a wedged remote tmux server or a hung `git` process left `GET /api/sessions/{id}/git-info` blocked indefinitely — observed in production as a 19-minute hang after opening the Pane Settings directory browser and cancelling it.
- `sshd`'s `ClientAliveInterval 60`/`ClientAliveCountMax 3` (max ~180s) ruled out a simple fully-dead-transport explanation for a 19-minute hang, pointing instead at an individual stuck remote command — the exec-timeout fix (Phase 1) is the direct fix for the reported symptom.
- SSH connections also had no liveness monitoring at all (`golang.org/x/crypto/ssh` has no `ServerAliveInterval` equivalent), so a half-dead transport left pane Read/Write/NewSession calls blocked forever with no way to detect the failure — this is the most likely cause of a pane's terminal appearing entirely unresponsive. Added as a defense-in-depth layer (Phase 2).

## Changes
- Added a package-level timeout to every local (`tmux`, `ps`, `lsof`, `git`) and SSH exec-channel command used for pane metadata inspection, following the existing `prLookupTimeout` pattern (5s for local/background-polling calls, 10s for user-triggered SSH calls like directory browsing).
- New `internal/session/ssh_timeout.go`: `runOutputWithTimeout` + `timeoutSessionRunner`, since `ssh.Session.Output` has no built-in timeout support.
- New `internal/session/ssh_keepalive.go`: background probe that sends an unrecognized SSH global request every 15s; two consecutive failures (~40s worst case) close the transport, which the existing `monitorSSHSession`/`classifySSHWaitError` machinery already reacts to by transitioning session state to `disconnected`.
- Bug found via an in-process fake-SSH-server integration test: forcing the transport closed made the pending `Session.Wait()` return `*ssh.ExitMissingError`, which was misclassified as `StateExited` (normal exit) instead of `StateDisconnected`. Fixed by marking the session disconnected before closing the transport and guarding the `Wait()`-driven state update so it can't downgrade that classification back to exited.
- Updated `docs/architecture.md` and `docs/behavior.md` to document the new timeout/keepalive behavior.

## Test plan
- [x] `make check` passes (lint, Go tests, frontend tests, coverage ≥ 80%)
- [x] `go test -race` passes for `internal/session` and `internal/api`
- [x] New table-driven/unit tests cover: timeout firing vs. completing just under timeout vs. erroring before timeout, for each local exec call site
- [x] New keepalive unit tests cover: consecutive successes never trigger `onDead`, consecutive failures trigger it exactly once, a single failure followed by success resets the counter, closing `stop` halts the loop before the next tick
- [x] New in-process fake-SSH-server integration tests verify end-to-end: a transport that stops answering keepalive probes is detected and the session transitions to `disconnected` within the shrunk keepalive window; a transport that keeps answering never gets disconnected by the keepalive loop

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -93,6 +93,14 @@ var gitExistsFn = func() error {
 
 var prLookupTimeout = 5 * time.Second
 
+// localGitCommandTimeout bounds how long a single local `git` invocation used
+// for pane git-context inspection (rev-parse, branch, config) may block
+// before being killed. Without this, a hung local git process (e.g. a stuck
+// network filesystem mount under the pane's cwd) left GetGitInfo blocked
+// indefinitely. Kept separate from prLookupTimeout so shrinking one in tests
+// does not affect the other.
+var localGitCommandTimeout = 5 * time.Second
+
 const responseErrorKey = "error"
 
 type sshConnectionsResponse struct {
@@ -1130,7 +1138,9 @@ func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error)
 }
 
 func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
-	branchCmd := exec.Command("git", "branch", "--show-current")
+	branchCtx, branchCancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
+	defer branchCancel()
+	branchCmd := exec.CommandContext(branchCtx, "git", "branch", "--show-current")
 	branchCmd.Dir = safeCWD
 	branchOut, err := branchCmd.Output()
 	if err != nil {
@@ -1140,7 +1150,9 @@ func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
 	}
 	branchOut = bytes.TrimSpace(branchOut)
 
-	originCmd := exec.Command("git", "config", "--get", "remote.origin.url")
+	originCtx, originCancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
+	defer originCancel()
+	originCmd := exec.CommandContext(originCtx, "git", "config", "--get", "remote.origin.url")
 	originCmd.Dir = safeCWD
 	originOut, err := originCmd.Output()
 	if err != nil {
@@ -1151,7 +1163,9 @@ func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
 }
 
 func runLocalGitContextCommand(safeCWD, originalCWD, operation string, args ...string) ([]byte, error) {
-	cmd := exec.Command("git", args...)
+	ctx, cancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, "git", args...)
 	cmd.Dir = safeCWD
 	out, err := cmd.Output()
 	if err == nil {
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -1088,6 +1088,14 @@ func (h *Handler) inspectGitContextForSession(sess session.Session, cwd string) 
 	return h.inspectLocalGitContext(cwd)
 }
 
+// inspectLocalGitContext runs up to four sequential git subprocesses
+// (rev-parse --show-toplevel, rev-parse --git-common-dir, then branch and
+// origin lookups inside localGitOptionalMetadata) against a single shared
+// deadline, rather than giving each an independent localGitCommandTimeout.
+// A stuck working directory (e.g. a hung network filesystem mount) makes
+// every git invocation against it hang the same way, so without a shared
+// deadline the calls could accumulate up to 4x localGitCommandTimeout
+// before the whole lookup fails through.
 func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error) {
 	safeCWD, err := sanitizeGitExecDir(cwd)
 	if err != nil {
@@ -1101,7 +1109,11 @@ func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error)
 		)
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
+	defer cancel()
+
 	toplevelOut, err := runLocalGitContextCommand(
+		ctx,
 		safeCWD,
 		cwd,
 		"git rev-parse --show-toplevel",
@@ -1114,6 +1126,7 @@ func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error)
 	toplevelOut = bytes.TrimSpace(toplevelOut)
 
 	commonDirOut, err := runLocalGitContextCommand(
+		ctx,
 		safeCWD,
 		cwd,
 		"git rev-parse --path-format=absolute --git-common-dir",
@@ -1126,7 +1139,7 @@ func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error)
 	}
 	commonDirOut = bytes.TrimSpace(commonDirOut)
 
-	branchOut, originOut := localGitOptionalMetadata(safeCWD)
+	branchOut, originOut := localGitOptionalMetadata(ctx, safeCWD)
 	root := string(toplevelOut)
 	return session.GitContext{
 		Branch:    string(branchOut),
@@ -1137,10 +1150,8 @@ func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error)
 	}, nil
 }
 
-func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
-	branchCtx, branchCancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
-	defer branchCancel()
-	branchCmd := exec.CommandContext(branchCtx, "git", "branch", "--show-current")
+func localGitOptionalMetadata(ctx context.Context, safeCWD string) ([]byte, []byte) {
+	branchCmd := exec.CommandContext(ctx, "git", "branch", "--show-current")
 	branchCmd.Dir = safeCWD
 	branchOut, err := branchCmd.Output()
 	if err != nil {
@@ -1150,9 +1161,7 @@ func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
 	}
 	branchOut = bytes.TrimSpace(branchOut)
 
-	originCtx, originCancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
-	defer originCancel()
-	originCmd := exec.CommandContext(originCtx, "git", "config", "--get", "remote.origin.url")
+	originCmd := exec.CommandContext(ctx, "git", "config", "--get", "remote.origin.url")
 	originCmd.Dir = safeCWD
 	originOut, err := originCmd.Output()
 	if err != nil {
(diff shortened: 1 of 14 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- Excellent test coverage across the full behavioral matrix: timeout-fires vs. completes-just-under-timeout vs. errors-before-timeout for every local exec call site, keepalive-loop unit tests with fake senders (consecutive successes/failures, single-failure-then-recovery, stop-before-next-tick), and genuine in-process fake-SSH-server integration tests that exercise the real `NewSSH` path end-to-end.
- The `*ssh.ExitMissingError` misclassification bug (write-up in `sshKeepaliveOnDeadFn`'s doc comment) is a sharp catch — a keepalive-forced transport close would otherwise look identical to a clean remote exit to `classifySSHWaitError`, and the fix (mark disconnected before closing, guard against downgrade in the `monitorSSHSession` callback) is well reasoned and clearly documented in code.
- `runOutputWithTimeout` cleanly decouples the timeout race from `*ssh.Session` (accepting `outputFn`/`closeFn` as plain funcs), which is what makes the unit tests in `ssh_timeout_test.go` possible without a real SSH connection.
- `docs/architecture.md` and `docs/behavior.md` updates accurately reflect the implementation (timeout values, keepalive interval/failure count, and the resulting state transition), consistent with the "document behavior in the same change" rule in `DEVELOPMENT.md`.
- Nice incidental cleanup: `NewTmuxSSH` now reuses the existing `validateTmuxSessionName` helper instead of duplicating the inline validation that `TmuxLocalSession` already had.

---

### Issues & Questions

**[Question] `sshKeepaliveOnDeadFn`'s `markDisconnected` has no guard against overwriting an explicit `Close()`, unlike the symmetric guard added for the reverse direction**

`internal/session/ssh_keepalive.go`'s `wireSSHSessionLifecycle` guards `monitorSSHSession`'s callback so a keepalive-detected `StateDisconnected` can never be downgraded back to `StateExited`:

```go
monitorSSHSession(sess, pw, func(newState State) {
    mu.Lock()
    defer mu.Unlock()
    if *state == StateDisconnected {
        return
    }
    *state = newState
})
```

But the callback built by `sshKeepaliveOnDeadFn` (`markDisconnected`) has no equivalent guard — it always sets `*state = StateDisconnected` unconditionally. `SSHSession.Close()` (`internal/session/ssh.go`) sets `s.state = StateExited` *before* it closes `s.client`:

```go
func (s *SSHSession) Close() error {
    s.keepaliveStopOnce.Do(func() { close(s.keepaliveStop) })
    s.mu.Lock()
    s.state = StateExited
    s.mu.Unlock()
    s.stdin.Close()
    s.session.Close()
    err := s.client.Close()   // <- this is what finally unblocks a probe already in flight
    ...
```

If the keepalive loop has already committed to a tick (i.e. it's inside `sshKeepaliveProbe`, blocked on `SendRequest`) at the moment `Close()` runs, `close(s.keepaliveStop)` doesn't reach it in time — it only unblocks on the *next* iteration's `select`. `client.Close()` a few lines later is what actually causes the in-flight `SendRequest` to error out, and if that was the failure that pushes `failures >= sshKeepaliveMaxFailures`, `onDead()` fires and unconditionally sets `*state = StateDisconnected`, strictly after `Close()` already set `StateExited`. Net effect: an intentional, clean session close can end up reporting as `disconnected` instead of `exited`.

Worth noting this window isn't purely theoretical/rare: it's largest exactly when the transport is already degraded (i.e. `SendRequest` is stuck for close to the full `sshKeepaliveProbeTimeout`) — which is plausibly correlated with a user noticing an unresponsive pane and closing it.

Was this considered acceptable, or should `markDisconnected` also check `*state != StateExited` before overwriting, mirroring the guard on the other side?

**[Minor question] `inspectLocalGitContext` chains four independently-timed-out git subprocesses — worst case ~20s instead of ~5s**

`internal/api/handler.go`'s `inspectLocalGitContext` calls `runLocalGitContextCommand` twice (toplevel, common-dir) and then `localGitOptionalMetadata`, which itself runs two more sequential git commands (branch, origin) — each with its own independent `context.WithTimeout(context.Background(), localGitCommandTimeout)`:

```go
func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
    branchCtx, branchCancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
    ...
    originCtx, originCancel := context.WithTimeout(context.Background(), localGitCommandTimeout)
    ...
```

In the motivating scenario described in the PR body (a stuck network filesystem mount under the pane's cwd), all four git invocations against the same `cwd` would likely hang simultaneously, so a single `GET /api/sessions/{id}/git-info` request could take up to ~20s (4 × 5s) to fail through, rather than being bounded by a single ~5s deadline. Was a shared deadline for the whole `inspectLocalGitContext` call (e.g. one `context.WithTimeout` passed down to all four commands) considered, or is the current per-command bound intentional/acceptable given it's still a dramatic improvement over the previous indefinite hang?

---

### Overall

Solid, well-tested fix for a real production hang, with unusually thorough attention to the subtle state-machine interaction between the new keepalive probe and the existing SSH session lifecycle. The one question worth resolving before merge is whether the `markDisconnected` callback needs the same "don't downgrade a more specific state" guard that `monitorSSHSession`'s callback already has, since as written it can overwrite an intentional `Close()`'s `StateExited` with `StateDisconnected` under a plausible race. The git-context timeout chaining is a minor, non-blocking follow-up question.

internal/session/ssh_keepalive.go
```diff
@@ -0,0 +1,145 @@
...
+// severed out from under an in-flight session." classifySSHWaitError
+// therefore classifies it as StateExited, not StateDisconnected, which
+// would misreport a keepalive-detected network failure as a normal session
+// exit and could suppress the frontend's reconnect behavior. Setting
+// StateDisconnected first, combined with the guard in the markExited
+// closure passed to monitorSSHSession, ensures that later, less specific
+// classification never overwrites this one.
+func sshKeepaliveOnDeadFn(markDisconnected func(), client, jumpClient *ssh.Client) func() {
+	return func() {
+		markDisconnected()
```
tomo-chan, line 57: This unconditionally sets state to StateDisconnected. If Close() has already run (setting StateExited) and this callback fires afterward — e.g. because a keepalive probe was already blocked in SendRequest when Close() ran, and client.Close() is what finally unblocked it — this overwrites the clean-exit state with disconnected. See the review body for the full race analysis. Should this guard on *state != StateExited, mirroring the downgrade guard added to monitorSSHSession's callback just above?

internal/api/handler.go
```diff
@@ -1130,7 +1138,9 @@ func (h *Handler) inspectLocalGitContext(cwd string) (session.GitContext, error)
 }
 
 func localGitOptionalMetadata(safeCWD string) ([]byte, []byte) {
```
tomo-chan: This plus runLocalGitContextCommand's two callers means a single GetGitInfo request can chain up to 4 sequential git subprocesses, each independently bounded by localGitCommandTimeout. In the stuck-mount scenario from the PR description, that's a worst case of ~20s rather than ~5s. Intentional, or worth a single shared deadline for the whole inspectLocalGitContext call?

tomo-chan:
Addressed both points in 83e6730:

**1. `markDisconnected` guard against overwriting an explicit `Close()`**

Confirmed the race is real. Added a symmetric guard: `markDisconnected` now refuses to overwrite `StateExited`, mirroring the existing guard against downgrading `StateDisconnected` back to `StateExited`. Verified with a new in-process fake-SSH-server test (`TestSSHSession_CloseDuringInFlightKeepaliveProbe_StaysExited`) that blocks a keepalive probe in flight, calls `Close()` while it's still blocked, and confirms the session stays `exited` — confirmed this fails without the guard (session ends up `disconnected`) and passes with it.

**2. Shared deadline across the four git subprocess calls**

Threaded a single `context.Context` from `inspectLocalGitContext` through `runLocalGitContextCommand` (both calls) and `localGitOptionalMetadata` (both calls), replacing the four independent `context.WithTimeout(localGitCommandTimeout)` calls. Added `TestInspectLocalGitContext_SharesDeadlineAcrossAllGitCommands`, which hangs the `branch`/`origin` lookups and confirms the whole call now returns within roughly one `localGitCommandTimeout` window instead of two.

While building that test, hit a real test-fixture gotcha worth flagging: a fake `git` script doing a plain `sleep 5` forks a child process that inherits the shell's stdout pipe. Killing the parent shell via `context.WithTimeout` doesn't close that pipe — `cmd.Output()` then blocks until the orphaned `sleep` child itself exits 5s later, silently defeating the timeout under test. Switched the fixtures to `exec sleep 5` (process replacement, no fork) to fix it. Worth keeping in mind for any future fake-subprocess test fixtures in this repo.

All existing + new tests pass (including `-race`), `make check` is green.

tomo-chan (commented):
## Code Review (follow-up)

Re-reviewed after commit `83e6730`, which addresses both points from the previous review round.

### Previously raised concerns — now resolved

**1. `markDisconnected` overwriting an explicit `Close()`'s `StateExited`** — Fixed with a symmetric guard in `wireSSHSessionLifecycle` (`internal/session/ssh_keepalive.go`): the `onDead` callback now checks `if *state == StateExited { return }` before setting `StateDisconnected`, mirroring the existing downgrade guard on the `monitorSSHSession` side. The new regression test, `TestSSHSession_CloseDuringInFlightKeepaliveProbe_StaysExited`, is a good design — it uses a real in-process SSH server that never replies to the keepalive probe, confirms (via a `probeStarted` channel) that a probe is genuinely blocked in `SendRequest` before calling `Close()`, then asserts the state stays `StateExited` after giving the keepalive loop ample time to process the now-unblocked failure. I ran it directly (`go test ./internal/session/... -run TestSSHSession_CloseDuringInFlightKeepaliveProbe_StaysExited -race -v`) along with the other keepalive integration tests — all pass.

**2. Four independently-timed-out git subprocesses in `inspectLocalGitContext`** — Fixed by threading a single shared `context.Context` from `inspectLocalGitContext` through both `runLocalGitContextCommand` calls and both calls inside `localGitOptionalMetadata`, replacing the four independent `context.WithTimeout` calls. Verified `TestInspectLocalGitContext_SharesDeadlineAcrossAllGitCommands` passes and correctly asserts the whole call stays under `2*localGitCommandTimeout` even with both `branch`/`origin` lookups hanging.

Nice catch during that second fix: the commit message notes a real test-fixture pitfall (a fake `git` script doing plain `sleep 5` forks a child that inherits the killed parent shell's stdout pipe, so `cmd.Output()` keeps blocking on the orphaned child and silently defeats the timeout under test) and fixes it by switching fixtures to `exec sleep 5` — applied consistently across `local_test.go` and `handler_test.go` as well, not just the new test. That's a subtle, easy-to-miss issue and good that it was caught and swept through the whole test suite rather than just patched locally.

I re-read the rest of the diff (no other files changed in this commit beyond `ssh_keepalive.go`, `handler.go`, `handler_test.go`, `local_test.go`, and `ssh_keepalive_integration_test.go`) and didn't find any new issues introduced by this round.

---

### Overall

Both review points are resolved with correct, well-tested fixes — verified by reading the diff and by running the new/related tests with `-race`. No remaining concerns from me. Good to merge from a correctness standpoint.

PR #159, opened 2026-07-21
fix: pass configured cwd to local tmux sessions

tomo-chan:
## Summary
- `internal/session/factory.go` never passed `pane.Cwd` to `NewTmuxLocal`, and `NewTmuxLocal` itself had no `cwd` parameter, so every local `type: tmux` pane ran `tmux new-session` without `-c` and inherited the panemux server process's own working directory instead of the `cwd` configured in `config.yaml`.
- Thread `cwd` through `factory.go` → `NewTmuxLocal` → `tmux new-session ... -c <cwd>`, matching how `ssh_tmux` already handles this (`internal/session/tmux_ssh.go`).
- `-c` only takes effect when tmux creates a brand-new session; attaching to an already-running session of the same name keeps that session's existing directory (same tmux-level caveat that already applies to `ssh_tmux`). Documented in `docs/behavior.md`.
- Verified `type: local`, `type: ssh`, and `type: ssh_tmux` already handle `cwd` correctly — no changes needed there.

## Test plan
- [x] `make check` passes (lint, Go tests + coverage, frontend tests + coverage, tsc)
- [x] New unit tests added first and confirmed failing before the fix (`TestTmuxLocalArgs_EmptyCwd_NoDashC`, `TestTmuxLocalArgs_WithCwd_AppendsDashC`), then passing after
- [x] Manual verification: built the binary, launched it with a `config.yaml` pane of `type: tmux` + a fresh `tmux_session` name + `cwd: /workspace/user/project`-style directory, and confirmed via `tmux display-message -p '#{pane_current_path}'` that the new session's cwd matched the configured value (previously it always fell back to the server's own launch directory)

### DIFF
--- a/internal/session/local_test.go
+++ b/internal/session/local_test.go
@@ -136,9 +136,19 @@ func TestLocalSessionGetCWD(t *testing.T) {
 }
 
 func TestNewTmuxLocal_InvalidSessionName_Error(t *testing.T) {
-	_, err := NewTmuxLocal("tmux-id", "title", "bad;session")
+	_, err := NewTmuxLocal("tmux-id", "title", "bad;session", "")
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "invalid tmux session name")
+}
+
+func TestTmuxLocalArgs_EmptyCwd_NoDashC(t *testing.T) {
+	args := tmuxLocalArgs("mysession", "")
+	assert.Equal(t, []string{tmuxNewSessionSubcommand, "-A", "-s", "mysession"}, args)
+}
+
+func TestTmuxLocalArgs_WithCwd_AppendsDashC(t *testing.T) {
+	args := tmuxLocalArgs("mysession", "/workspace/user/project")
+	assert.Equal(t, []string{tmuxNewSessionSubcommand, "-A", "-s", "mysession", "-c", "/workspace/user/project"}, args)
 }
 
 func TestProcessIDArg_RejectsNonPositivePID(t *testing.T) {
--- a/internal/session/tmux_local.go
+++ b/internal/session/tmux_local.go
@@ -32,13 +32,17 @@ type TmuxLocalSession struct {
 }
 
 // NewTmuxLocal creates a new session that attaches to a local tmux session.
-func NewTmuxLocal(id, title, tmuxSession string) (*TmuxLocalSession, error) {
+// cwd, when set, is only honored when tmux creates a brand-new session: if a
+// session named tmuxSession is already running, "-c" is ignored by tmux and
+// the pane keeps that session's existing working directory.
+func NewTmuxLocal(id, title, tmuxSession, cwd string) (*TmuxLocalSession, error) {
 	validatedSession, err := validateTmuxSessionName(tmuxSession)
 	if err != nil {
 		return nil, err
 	}
 
-	cmd := exec.Command("tmux", "new-session", "-A", "-s", validatedSession)
+	cmd := exec.Command("tmux")
+	cmd.Args = append([]string{"tmux"}, tmuxLocalArgs(validatedSession, cwd)...)
 	cmd.Env = append(os.Environ(), "TERM=xterm-256color")
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
 
@@ -191,6 +195,19 @@ func (s *TmuxLocalSession) Close() error {
 		s.cmd.Process.Kill()
 	}
 	return nil
+}
+
+const tmuxNewSessionSubcommand = "new-session"
+
+// tmuxLocalArgs builds the "tmux new-session" argument list. cwd, when
+// non-empty, is passed via "-c" as a discrete exec.Command argument (no
+// shell involved), matching the ssh_tmux "-c" handling in tmux_ssh.go.
+func tmuxLocalArgs(tmuxSession, cwd string) []string {
+	args := []string{tmuxNewSessionSubcommand, "-A", "-s", tmuxSession}
+	if cwd != "" {
+		args = append(args, "-c", cwd)
+	}
+	return args
 }
 
 func validateTmuxSessionName(tmuxSession string) (string, error) {
--- a/internal/session/tmux_local.go
+++ b/internal/session/tmux_local.go
@@ -14,8 +14,10 @@ import (
 	"github.com/creack/pty"
 )
 
+const tmuxBinary = "tmux"
+
 var tmuxLocalOutputFn = func(args ...string) ([]byte, error) {
-	return exec.Command("tmux", args...).Output()
+	return exec.Command(tmuxBinary, args...).Output()
 }
 
 // TmuxLocalSession attaches to an existing local tmux session via PTY.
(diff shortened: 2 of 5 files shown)


tomo-chan (commented):
## Code Review

### What's Good

- The fix threads `pane.Cwd` through `factory.go` → `NewTmuxLocal` → `tmux new-session ... -c <cwd>` in exactly the way `ssh_tmux` already does, keeping the two tmux-backed session types consistent.
- `cwd` is appended as a discrete `exec.Command` argument (`"-c", cwd`), not interpolated into a shell string, so this doesn't introduce a command-injection surface — consistent with `docs/security.md`'s requirement that arguments after the command may be user-supplied only when the target binary can't reinterpret them as commands.
- `tmuxLocalArgs` is extracted as a small pure function and tested directly (`TestTmuxLocalArgs_EmptyCwd_NoDashC`, `TestTmuxLocalArgs_WithCwd_AppendsDashC`) rather than only through the PTY-spawning constructor — good adherence to the "test at the smallest unit" rule, and it mirrors the existing pattern for `tmuxSSHCommand` in `tmux_ssh.go`.
- The known tmux caveat (`-c` only applies when tmux creates a brand-new session, not when attaching to an existing one) is documented in both the doc comment on `NewTmuxLocal` and `docs/behavior.md`, matching the `ssh_tmux` precedent.
- PR description clearly states the TDD sequence (failing tests first) and includes a manual verification step via `tmux display-message -p '#{pane_current_path}'`.

---

### Issues & Questions

**[Suggestion] Simplify the `exec.Command` construction**

```go
cmd := exec.Command("tmux")
cmd.Args = append([]string{"tmux"}, tmuxLocalArgs(validatedSession, cwd)...)
```

Since `tmuxLocalArgs` already returns the full argument list, this could be written more directly as:

```go
cmd := exec.Command("tmux", tmuxLocalArgs(validatedSession, cwd)...)
```

This avoids constructing an argument-less `Cmd` and then overwriting `cmd.Args` wholesale (which works today since `cmd.Path` is resolved independently via `exec.Command("tmux")`'s internal `LookPath`, but is a bit indirect to read). Not blocking, just a small clarity win.

**[Minor question] Factory-level wiring isn't covered by a test**

The bug this PR fixes was specifically that `factory.go` never forwarded `pane.Cwd` to `NewTmuxLocal` — i.e. the break was in the wiring, not in the tmux-argument logic itself. The new tests cover `tmuxLocalArgs` and the `NewTmuxLocal` signature directly, but there's no test asserting that `createSession`/`CreateFromConfig` actually passes `pane.Cwd` through for `TypeTmux` (something like the existing `TestCreateFromConfig_Local` pattern, but for tmux). I see `session/tmux*` integration tests are intentionally excluded from the unit suite per the `Makefile` coverage comment ("require live SSH/tmux"), so I understand a full `CreateFromConfig` round-trip may not be practical in CI. Was a lighter-weight wiring check (e.g. asserting the call site directly, or refactoring `createSession`'s `TypeTmux` branch to be independently testable) considered, so a future refactor of `factory.go` can't silently drop the `cwd` argument again without a test catching it?

---

### Overall

Small, focused fix that mirrors an existing, already-reviewed pattern (`ssh_tmux`'s `-c` handling), is well-tested at the unit level, and updates docs in the same change. No blockers — the two notes above are a minor readability suggestion and a test-coverage question, not correctness concerns.

internal/session/tmux_local.go
```diff
@@ -32,13 +32,17 @@ type TmuxLocalSession struct {
...
+// the pane keeps that session's existing working directory.
+func NewTmuxLocal(id, title, tmuxSession, cwd string) (*TmuxLocalSession, error) {
 	validatedSession, err := validateTmuxSessionName(tmuxSession)
 	if err != nil {
 		return nil, err
 	}
 
-	cmd := exec.Command("tmux", "new-session", "-A", "-s", validatedSession)
+	cmd := exec.Command("tmux")
+	cmd.Args = append([]string{"tmux"}, tmuxLocalArgs(validatedSession, cwd)...)
```
tomo-chan: [Suggestion] This could be simplified to `cmd := exec.Command("tmux", tmuxLocalArgs(validatedSession, cwd)...)` instead of constructing an argument-less `Cmd` and then overwriting `cmd.Args`. See the review body for details.

internal/session/local_test.go
```diff
@@ -136,11 +136,21 @@ func TestLocalSessionGetCWD(t *testing.T) {
...
 
 func TestNewTmuxLocal_InvalidSessionName_Error(t *testing.T) {
-	_, err := NewTmuxLocal("tmux-id", "title", "bad;session")
+	_, err := NewTmuxLocal("tmux-id", "title", "bad;session", "")
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "invalid tmux session name")
 }
 
+func TestTmuxLocalArgs_EmptyCwd_NoDashC(t *testing.T) {
+	args := tmuxLocalArgs("mysession", "")
```
tomo-chan, line 145: [Minor question] These cover `tmuxLocalArgs` directly, which is good — but is there a test that confirms `createSession`/`CreateFromConfig` for `TypeTmux` actually forwards `pane.Cwd` through? That factory-level wiring is exactly what was broken before this fix. See the review body for more context on why a full live-tmux integration test may not be practical here.

tomo-chan:
Addressed both points in 7e64420:

**Simplify `exec.Command`**: I tried the direct `exec.Command("tmux", tmuxLocalArgs(...)...)` form first, but it re-triggers gosec's G204 check (confirmed locally — `golangci-lint run ./...` flags it), which `make lint-go`/`make check` would then fail on. Per `docs/security.md`'s preference for structural fixes over suppressions in shipped code, I kept the `cmd.Args` assignment but added a comment explaining why it's written this way instead of the more direct form, so the next reader isn't left guessing.

**Factory-level wiring test**: Added an injectable `newTmuxLocalFn` seam in `factory.go` (same pattern as the existing `tmuxLocalOutputFn`/`listProcessesFn`/etc. seams in this package) plus two new tests — `TestCreateSession_Tmux_PassesCwdThrough` and `TestCreateSession_Tmux_EmptyCwd_PassesEmptyString` — that verify `createSession`'s `TypeTmux` branch forwards `pane.Cwd` through, without spawning a real tmux process. I confirmed these tests actually catch the original regression by temporarily reintroducing the bug (dropping `pane.Cwd` in the `TypeTmux` case) and watching `TestCreateSession_Tmux_PassesCwdThrough` fail before reverting.

`make check` passes with both changes.

tomo-chan (commented):
## Code Review

Thanks for the quick turnaround in 7e64420 — both prior points are resolved, and I independently verified both claims rather than taking them at face value:

- **`exec.Command` construction**: I checked out this branch and swapped in the direct form (`exec.Command(tmuxBinary, tmuxLocalArgs(...)...)`) locally, then ran `golangci-lint run ./internal/session/...`. It does reproduce `G204: Subprocess launched with a potential tainted input or cmd arguments` on that line, while the current `cmd.Args` assignment form reports 0 issues. So the earlier suggestion was actually wrong in practice — good catch, and the comment you added explaining why makes this non-obvious tradeoff clear for the next reader.
- **Factory-level wiring test**: I reproduced your regression check — reverting `factory.go`'s `TypeTmux` branch back to a hardcoded `""` for cwd causes `TestCreateSession_Tmux_PassesCwdThrough` to fail with the expected diff (`/workspace/user/project` vs `""`), confirming the new test actually catches the original bug class. The `newTmuxLocalFn` seam matches the existing `tmuxLocalOutputFn`/`listProcessesFn` pattern in this package, so it's consistent with prior art rather than a one-off.

All new/changed tests pass locally (`TestCreateSession_Tmux_PassesCwdThrough`, `TestCreateSession_Tmux_EmptyCwd_PassesEmptyString`, `TestNewTmuxLocal_InvalidSessionName_Error`, `TestTmuxLocalArgs_*`).

No further findings from me on this round.

---

### Overall

Both review points from the previous round are addressed and verified independently (one turned out to be a real lint constraint, not just a style preference). No blockers — good to merge from my side.

PR #182, opened 2026-08-23
test: drive every /api route through the real router (#180 item 1)

tomo-chan:
Closes the two remaining boxes of #180's roadmap item 1 — gate **G3(a)** in [docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md).

- [x] `/api/*` の正常系結合テストを実ルータ経由に移す
- [x] `openurl_routes_test.go` / `server_test.go` に残る個別ルート担保の整理（chi の優先順位テスト 2 本は意図的に残す）

## What this adds

`internal/server/api_integration_test.go` walks the router `server.New()` actually builds and drives **every** `/api` route through it — middleware stack, `/api` vs `/api/board` mount precedence, and the SPA catch-all included. `route_table_test.go` already pinned *which* routes exist and which are authenticated; this is the other half of the same question: that each of them reaches its real handler and answers the way the frontend expects.

The table is **exhaustive by construction**, the same shape as the route table's own check (design decision D3):

- a registered `/api` route with no case fails the test;
- a case naming a route the router does not register fails it too;
- both runner states are walked, so a route registered only when the command center is enabled cannot escape either check.

So adding a route now forces an integration case for it. No new `*_routes_test.go` file is needed for the next feature either.

## Coverage of the table

22 of the 24 routes drive their genuine success path, including a **real local pane** created and deleted through `POST`/`DELETE /api/sessions` and reused by `git-info`, `open-url` and `restart`.

The two that cannot carry a written reason and are listed in `documentedIntegrationGaps` — a gap can be accepted, but not accumulated quietly, because adding one means changing a literal a reviewer sees in the diff:

| Route | Why not the success path |
|---|---|
| `POST /api/sessions/{id}/open-vscode` | a 2xx needs a `code` binary on PATH and a real editor launch |
| `POST /api/board/broadcast` | delivery needs a reachable remote agmsg host over SSH |

Both still assert the response came from the route's *own* handler rather than chi's 404, so the dispatch itself is proven.

That literal is backed by a derived check: `apiEnv.do` records every status it observes, and a case with no `notHappyPathReason` must have driven at least one 2xx. Neither half works alone — the literal cannot see a case that quietly stops asserting a success path, and the derived check puts nothing in the diff for a reviewer to notice.

### Hermeticity (principle 5)

- `HOME` points at a temp directory, set before `New()` because `api.NewHandler` resolves `sshconfig.DefaultPath()` once at construction. `~/.ssh/config` is therefore read **and written** inside the test, never on the developer's machine.
- `XDG_CACHE_HOME` points inside it too: `os.UserCacheDir` prefers that variable over `$HOME/.cache`, and creating a local pane installs the browser shim there, so `HOME` alone left three real files outside the temp home on any machine that exports it.
- The config carries no file path, so `config.write()` is a no-op and the layout/workspace writes these routes perform stay in memory.
- Pane `cwd` is a temp directory, so `git-info` answers "not a repository" without walking this checkout.
- Every server built here is `Shutdown` on cleanup, releasing the port-forward sweeper goroutine `New()` starts.

## Consolidated away

Now covered by the table:

- `TestServer_APIRoutesWired`
- `TestServer_WorkspaceRenameRouteWired`
- `TestServer_OpenURLRouteIsWired`
- `TestServer_OpenURLRouteIsNotBehindBoardAuth` (also covered by `TestServer_NonBoardAPIRoutesStayUnauthenticated`)

`openurl_routes_test.go` is deleted. Its `TestServer_ShutdownClosesPortForwards` moves to `server_test.go` — it is a lifecycle test about state `New()` owns, not a route test.

**Kept deliberately**: the two chi precedence tests. `/api/workspaces/tab-position` and `/api/workspaces/vertical-bar-width` are both matched by the `/api/workspaces/{id}` pattern registered after them, and no exhaustiveness check over the route *set* can see registration order — both routes exist either way. Their doc comment now says so.

## Verified by perturbation, not assumed

| Perturbation | Result |
|---|---|
| Delete one case from `apiCases` | red — "every /api route must be driven through the real router" |
| Add a case for `GET /api/ghost` | red — "these cases name routes the router does not register" |
| `New()`: `SetCommandCenterAvailable(commandRunner != nil)` → `(false)` | red — **only this test catches it**; the api package's handler tests set that flag themselves |

One claim was written and then removed because the perturbation contradicted it: an earlier comment said the `open-url` case would catch `New()` dropping `SetPortForwards`. It does not — `PostOpenURL` returns for a local pane before consulting the registry. The comment now says what the assertion actually covers, and points at `TestServer_ShutdownClosesPortForwards` for the rest.

## Review rounds (729d1a1)

Eight findings across two rounds, every one reproduced by the reviewer and each re-verified here after the fix.

**Three cases passed against a broken handler** — worth stating plainly, since this file is the G3(a) gate:

| Case | Perturbation that used to stay green | Fix |
|---|---|---|
| `PUT /api/workspaces/active` | handler ignores the requested id entirely | it asserted the fixture back to itself; now creates a second workspace and switches away from whichever is active |
| `GET /api/board/status` | handler returns an empty body | seeds a real status, asserts the envelope `useBoardStatus` parses |
| `GET /api/display` | handler returns `{}` | seeds a `DisplayConfig`, asserts both fields |

**Two were about the guarantees the file advertises.** `registeredAPIRoutes` walked only the command-center-disabled router, so an `/api` route inside `registerRoutes`' `if commandRunner != nil` block escaped both checks — the file header and the rollout row were stronger than the code. And `documentedIntegrationGaps` took `notHappyPathReason` as its only input, so a happy-path case rewritten into an error-path one left the ledger untouched; that is the derived check described above.

**Three smaller ones**: the `XDG_CACHE_HOME` hermeticity hole (measured: three files written outside the temp home, now none); the missing `Shutdown` (measured with a throwaway probe: five envs leaked five goroutines, now zero); and a comment claiming httptest's default `RemoteAddr`/`Host` are loopback — they are `192.0.2.1:1234` and `example.com`, and acting on the comment turns the case into a 403 from the DNS-rebinding guard.

## Docs

`docs/quality-gateway.md` G3's status row, rollout row 1, and the maintainability pillar row are updated. G3(a) is marked present **for HTTP only** — the `/ws` half of that gate is still absent, and the row says so rather than rounding up.

## Test plan

- [x] `make test-go` — all packages pass
- [x] `make coverage-go` — 88.0% → **88.1%**
- [x] `make check` — passes (also enforced by the `pre-push` hook on this branch)
- [x] The three original perturbations above each confirmed red, then reverted
- [x] All eight review findings re-verified: each reviewer perturbation re-run after the fix and confirmed to fail the case it should
- [x] Goroutine and `XDG_CACHE_HOME` claims measured rather than reasoned about

---


internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+	"POST /api/workspaces": {run: func(t *testing.T, e *apiEnv) {
+		rr := e.do(t, http.MethodPost, "/api/workspaces", "")
+		assert.Equal(t, http.StatusCreated, rr.Code)
+		assert.Len(t, e.cfg.WorkspacesView().Items, 2)
+	}},
+
+	"PUT /api/workspaces/active": {run: func(t *testing.T, e *apiEnv) {
+		rr := e.do(t, http.MethodPut, "/api/workspaces/active", `{"id":"default"}`)
+		assert.Equal(t, http.StatusOK, rr.Code)
+		assert.Equal(t, "default", e.cfg.Workspaces.Active)
```
tomo-chan: **This assertion is vacuous — the case passes against a handler that ignores the requested id entirely.**

`testConfig()` already has `Workspaces.Active == "default"` and exactly one workspace, and the case PUTs `{"id":"default"}`. So asserting `Active == "default"` afterwards restates the starting state; nothing here separates "the switch was applied" from "the handler returned 200 and did nothing".

Verified by perturbation against this branch:

```go
// internal/api/handler.go PutActiveWorkspace
if !h.cfg.SetActiveWorkspace(h.cfg.Workspaces.Active) { // requested id ignored
```

Workspace switching is now completely broken, and `TestServer_APIIntegration/PUT_/api/workspaces/active` still passes (`ok panemux/internal/server`).

Every sibling workspace case deliberately moves a value away from its default — `tab-position` top→right, `vertical-bar-width` 280→320, `{id}` Default→Renamed — which is exactly why they *would* catch their handler regressing. This one is the odd one out. The `DELETE /api/workspaces/{id}` case above already shows the fix pattern:

```go
require.Equal(t, http.StatusCreated, e.do(t, http.MethodPost, "/api/workspaces", "").Code)
added := e.cfg.WorkspacesView().Items[1].ID

rr := e.do(t, http.MethodPut, "/api/workspaces/active", fmt.Sprintf(`{"id":%q}`, added))
assert.Equal(t, http.StatusOK, rr.Code)
assert.Equal(t, added, e.cfg.Workspaces.Active)
```


---

internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+		)}
+		rr = enabled.do(t, http.MethodGet, "/api/session-token", "")
+		assert.Equal(t, http.StatusOK, rr.Code)
+		assert.Contains(t, rr.Body.String(), `"command_center_enabled":true`)
+	}},
+
+	"GET /api/board/status": {run: func(t *testing.T, e *apiEnv) {
+		rr := e.do(t, http.MethodGet, "/api/board/status", "")
+		assert.Equal(t, http.StatusOK, rr.Code)
+	}},
```
tomo-chan, line 444: **This case asserts only the status code, so it passes with an entirely empty response body — and it is not listed in `documentedIntegrationGaps`.**

Verified by perturbation:

```go
// internal/api/board.go GetBoardStatus
w.WriteHeader(http.StatusOK) // no body at all
return
```

`TestServer_APIIntegration/GET_/api/board/status` still passes.

This matters more here than for most routes: `/api/board/status` is the endpoint `useBoardStatus` polls and Zod-parses, so an empty or wrongly-shaped body takes the dashboard down. The file's own header says a happy-path case asserts "the response a working frontend request produces" — a bare 200 is the one thing that does not distinguish the handler from chi's dispatch alone.

Its sibling one case down already does the right thing (`assert.Contains(rr.Body.String(), `"messages":[]`)`) and would catch the equivalent regression. Suggest matching it:

```suggestion
	"GET /api/board/status": {run: func(t *testing.T, e *apiEnv) {
		rr := e.do(t, http.MethodGet, "/api/board/status", "")
		assert.Equal(t, http.StatusOK, rr.Code)
		// The cache is empty, so the envelope the dashboard's useBoardStatus
		// hook parses must still be present and well-formed.
		assert.Contains(t, rr.Body.String(), `"statuses":{}`)
	}},
```

(If the envelope key differs, adjust — the point is that *something* about the body is asserted. The alternative is to give the case a `notHappyPathReason` and add it to `documentedIntegrationGaps`, which is what that mechanism is for.)

---

internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+		assert.Equal(t, http.StatusOK, rr.Code)
+		// The pane's cwd is a temp directory, so the answer is a valid
+		// "not a repository" one rather than anything about this checkout.
+		assert.Contains(t, rr.Body.String(), `"is_git":false`)
+	}},
+
+	"GET /api/display": {run: func(t *testing.T, e *apiEnv) {
+		rr := e.do(t, http.MethodGet, "/api/display", "")
+		assert.Equal(t, http.StatusOK, rr.Code)
+		assert.Contains(t, rr.Body.String(), "{")
```
tomo-chan: **`Contains(body, "{")` is true of any JSON object, so this asserts nothing beyond the 200 above it.**

Verified by perturbation:

```go
// internal/api/handler.go GetDisplay
writeJSON(w, struct{}{}) // the whole display configuration dropped
```

`TestServer_APIIntegration/GET_/api/display` still passes.

I suspect the reason is that `testConfig()` sets no `Display` fields, so there was nothing distinctive to assert. Seeding one in the case (the same way the `GET /api/ssh-connections` case seeds `e.cfg.SSHConnections` after `New()` — the handler holds the `*config.Config` pointer, so a post-construction write is visible) makes it real:

```go
"GET /api/display": {run: func(t *testing.T, e *apiEnv) {
	e.cfg.Display.<Field> = <value>

	rr := e.do(t, http.MethodGet, "/api/display", "")
	assert.Equal(t, http.StatusOK, rr.Code)
	assert.Contains(t, rr.Body.String(), `"<json_tag>":<value>`)
}},
```

---

internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+		"a route whose success path is not driven here must be listed, with a reason, in documentedIntegrationGaps")
+}
+
+// registeredAPIRoutes is every /api route on the real router, as
+// "METHOD /pattern". It deliberately reuses walkRoutes so this test and the
+// route table cannot disagree about what "a route" is.
+func registeredAPIRoutes(t *testing.T) []string {
+	t.Helper()
+
+	srv := New(testConfigWithToken(integrationToken), session.NewManager(), board.NewBoardCache(), nil, nil, emptyFS)
```
tomo-chan: **The exhaustiveness guarantee — this file's whole value proposition — has a hole: it only walks the command-center-*disabled* router.**

`commandRunner` is `nil` here, so `registerRoutes`' `if commandRunner != nil` block is never entered. Any `/api` route registered inside it is invisible to both the `missing` and the `stale` check.

Verified by perturbation:

```go
// internal/server/server.go registerRoutes
if commandRunner != nil {
	boardCommandHandler := ws.NewBoardCommandHandler(commandRunner, authToken)
	r.Get("/ws/board-command", boardCommandHandler.ServeHTTP)
	r.Get("/api/command-center/ping", func(w http.ResponseWriter, _ *http.Request) { w.WriteHeader(200) })
}
```

`TestServer_APIIntegration` → **green**, with no case for the new route. (`TestServer_RouteTable_CommandCenterEnabled_AddsOnlyBoardCommandWS` does go red, so it is not a total blind spot — but that test only pins that the route *exists*, which is precisely the half this file exists to complement.)

Note that `route_table_test.go` deliberately walks *both* runner states for exactly this reason. This table walks one, which makes the file header's "the test walks the real router and fails when a registered /api route has no case here" and `docs/quality-gateway.md`'s new rollout row ("itself exhaustive, so a new route has no test until someone writes one") stronger than what the code delivers. Union both routers:

```suggestion
func registeredAPIRoutes(t *testing.T) []string {
	t.Helper()

	// Both runner states: a route registered only when the command center is
	// enabled is still a route the frontend can reach, and walking only the
	// disabled router would let it escape this table's exhaustiveness check.
	// route_table_test.go pins both states for the same reason.
	seen := map[string]bool{}
	for _, runner := range []*commandcenter.Runner{nil, commandcenter.NewRunner(commandcenter.RunnerConfig{})} {
		srv := New(testConfigWithToken(integrationToken), session.NewManager(), board.NewBoardCache(), nil, runner, emptyFS)
		for _, route := range walkRoutes(t, srv) {
			_, pattern, _ := strings.Cut(route, " ")
			if strings.HasPrefix(pattern, "/api/") {
				seen[route] = true
			}
		}
	}

	out := slices.Collect(maps.Keys(seen))
	sort.Strings(out)
	require.Positive(t, len(out), "no /api routes were found — the router changed")
	return out
}
```

(needs `maps` in the import block.)

---

internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+
+		rr := e.do(t, http.MethodGet, "/api/directories?path="+e.home, "")
+		assert.Equal(t, http.StatusOK, rr.Code)
+		assert.Contains(t, rr.Body.String(), `"name":"child"`)
+	}},
+
+	"GET /api/session-token": {run: func(t *testing.T, e *apiEnv) {
+		// httptest.NewRequest's default RemoteAddr and Host are both
+		// loopback, which is exactly what this route requires.
+		rr := e.do(t, http.MethodGet, "/api/session-token", "")
```
tomo-chan: **This comment is factually wrong, and it directly contradicts `do`'s own (correct) comment 30 lines above.**

`httptest.NewRequest` sets `RemoteAddr` to `192.0.2.1:1234` (TEST-NET-1) and, for a path-only target, `Host` to `example.com` — neither is loopback. `do` says exactly that, and overrides both *because* of it. This route works here only thanks to those two lines.

That matters because the comment actively invites removing them. Verified by perturbation — dropping `req.Host` / `req.RemoteAddr` from `do`:

```
Error: Not equal: expected: 200, actual: 403
Error: "forbidden\n" does not contain "\"command_center_enabled\":true"
```

i.e. `GetBoardSessionToken`'s DNS-rebinding guard, which is the one thing standing between this token and any LAN client (see docs/security.md). A future cleanup that trusts this comment breaks the case and would reasonably conclude the *guard* is at fault.

```suggestion
		// do() gives every request a loopback RemoteAddr and Host — not
		// httptest's own defaults, which are 192.0.2.1:1234 and example.com.
		// This route is the reason do() overrides them: see
		// api.GetBoardSessionToken's DNS-rebinding guard.
```

---

internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+		mgr:   mgr,
+		cache: cache,
+		home:  home,
+	}
+	t.Cleanup(func() {
+		for _, s := range mgr.List() {
+			_ = mgr.Remove(s.ID())
+		}
+	})
+	return e
```
tomo-chan, line 132: Minor, but worth closing while the cleanup hook is right here: this releases the sessions and not the server.

`New()` unconditionally calls `portforward.New(portforward.Options{})`, and with `SweepInterval == 0` that takes the `opts.SweepInterval >= 0` branch and starts a `go r.sweep(time.Minute)` that only exits on `Close()`. So every `New()` in this file — one per subtest, plus `registeredAPIRoutes` and the second server in the `GET /api/session-token` case — leaves a goroutine behind, ~30 in all.

Nothing fails today (the package finishes in ~25 ms, long before a tick), but it is a bit pointed given this same diff moves `TestServer_ShutdownClosesPortForwards` into `server_test.go` with the comment "process-wide state `New()` owns and `Shutdown` must release". It also makes the package hostile to a `goleak`-style check later, which is plausibly where G4 goes.

```suggestion
	t.Cleanup(func() {
		for _, s := range mgr.List() {
			_ = mgr.Remove(s.ID())
		}
		// New() starts a port-forward sweeper goroutine that only exits on
		// Close(); Shutdown is what releases it. See
		// TestServer_ShutdownClosesPortForwards.
		_ = e.srv.Shutdown(context.Background())
	})
```

(needs `context` in the import block.)

---

tomo-chan (commented):
Code review of the diff (`913b8a8...c7f878c`). The direction is right and the exhaustiveness-by-construction shape is the good idea here — a route added to `Mount` genuinely does force a case. Six findings, all verified by perturbation against this branch rather than reasoned about.

Two are about the guarantee itself:

- `registeredAPIRoutes` only walks the command-center-**disabled** router, so an `/api` route registered inside `registerRoutes`' `if commandRunner != nil` block escapes both the `missing` and `stale` checks. `route_table_test.go` deliberately walks both states; this one walks one, which makes the file header's claim and `docs/quality-gateway.md`'s new rollout row stronger than the code.
- The `GET /api/session-token` comment asserts that httptest's default `RemoteAddr`/`Host` are loopback. They are not (`192.0.2.1:1234` / `example.com`), and `do`'s own comment says so correctly. Acting on this comment turns the case into a 403 from the DNS-rebinding guard.

Three are cases that pass against a broken handler — relevant because this file is the G3(a) gate and G4(b) red-check is the next rollout item:

| Case | Perturbation that stays green |
|---|---|
| `PUT /api/workspaces/active` | handler ignores the requested id entirely |
| `GET /api/board/status` | handler returns an empty body |
| `GET /api/display` | handler returns `{}` |

`PUT /api/workspaces/active` is the notable one: it is the only workspace case that asserts a value equal to its own starting state, while `tab-position`, `vertical-bar-width` and `{id}` all deliberately move one.

The sixth is minor — `newAPIEnv` never calls `Shutdown`, leaking the sweeper goroutine `New()` starts, ~30 across the file.

Nothing here is a production-behavior bug; the diff is test + docs only, and the consolidation itself checks out (`TestServer_APIRoutesWired`, `TestServer_WorkspaceRenameRouteWired` and both deleted `openurl_routes_test.go` route tests are genuinely subsumed, and `TestServer_ShutdownClosesPortForwards` moved intact).

---

internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+//     for POST /api/ssh-config/hosts, writing) the developer's own files.
+//     It has to be set before New(), because api.NewHandler resolves
+//     sshconfig.DefaultPath() once at construction.
+//   - the config carries no file path, which makes config.write() a no-op, so
+//     the layout and workspace writes these routes perform stay in memory.
+func newAPIEnv(t *testing.T) *apiEnv {
+	t.Helper()
+
+	home := t.TempDir()
+	t.Setenv("HOME", home)
```
tomo-chan, line 106: **`HOME` alone does not make this hermetic: `createPane` writes outside the temp home whenever `XDG_CACHE_HOME` is set.**

Creating a local pane installs the browser shim (`internal/session/local.go:76` → `browserShimEnvForLocalSession` → `installLocalBrowserShim`), which resolves its directory with `os.UserCacheDir()` — and on Linux that prefers `XDG_CACHE_HOME` over `$HOME/.cache`. The doc comment above says the point of the temp `HOME` is that these routes read and write "inside the test rather than … the developer's own files", but the shim escapes it.

Verified by perturbation on this branch:

```sh
XDG_CACHE_HOME=/tmp/xdgcache-probe go test ./internal/server/ -run 'TestServer_APIIntegration$' -count=1
# ok  panemux/internal/server  0.032s
find /tmp/xdgcache-probe -type f
# /tmp/xdgcache-probe/panemux/bin/panemux-open
# /tmp/xdgcache-probe/panemux/bin/open
# /tmp/xdgcache-probe/panemux/bin/xdg-open
```

Three real files, mode 0700, written by the suite into a directory the test never chose. With `XDG_CACHE_HOME` unset the fallback is `$HOME/.cache`, so the temp `HOME` does cover it — which is exactly why this is easy to miss locally and easy to hit in a CI image or a desktop session that exports the variable.

One line in the same place closes it:

```suggestion
	home := t.TempDir()
	t.Setenv("HOME", home)
	// os.UserCacheDir prefers XDG_CACHE_HOME over $HOME/.cache, and creating
	// a local pane installs the browser shim there (see
	// session.installLocalBrowserShim), so HOME alone does not contain it.
	t.Setenv("XDG_CACHE_HOME", filepath.Join(home, ".cache"))
```

(`filepath` is already imported.)

---

internal/server/api_integration_test.go
```diff
@@ -0,0 +1,491 @@
...
+// the diff a reviewer reads.
+var documentedIntegrationGaps = []string{
+	"POST /api/board/broadcast",
+	"POST /api/sessions/{id}/open-vscode",
+}
+
+func TestServer_APIIntegration_DocumentedGaps(t *testing.T) {
+	var gaps []string
+	for route, c := range apiCases {
+		if c.notHappyPathReason == "" {
```
tomo-chan, line 543: **The gaps ledger is opt-in, so it catches only the gaps someone chose to declare — the case it is meant to catch is the one that quietly stops asserting a success path.**

`notHappyPathReason` is the sole input to this loop, and nothing anywhere requires a case that asserts a non-2xx to set it. So the ledger enforces "a declared gap must be listed", but not the property the surrounding comment claims: that a gap "can be accepted, but not accumulated quietly."

Verified by perturbation — rewriting an existing happy-path case as an error-path one, with no `notHappyPathReason` added:

```go
"PUT /api/layout": {run: func(t *testing.T, e *apiEnv) {
	rr := e.do(t, http.MethodPut, "/api/layout", `not json`)
	assert.Equal(t, http.StatusBadRequest, rr.Code)
}},
```

`go test ./internal/server/ -count=1` → **ok**. The route now has no success-path coverage at all, `documentedIntegrationGaps` is unchanged, and no reviewer sees a literal move in the diff — which is precisely the signal that list exists to produce.

This is the same shape as the exhaustiveness check two functions up, and it deserves the same treatment: derive the fact from the run rather than from a field the author remembers to set. Have `apiEnv.do` record the statuses it observed and have the harness require at least one 2xx unless the case declared a reason — e.g. a `statuses []int` field on `apiEnv` appended to in `do`, checked after `c.run` in `TestServer_APIIntegration`:

```go
if c.notHappyPathReason == "" {
	assert.True(t, slices.ContainsFunc(e.statuses, func(s int) bool { return s >= 200 && s < 300 }),
		"a case with no notHappyPathReason must drive the route's success path")
}
```

Then the perturbation above fails where it should, and `documentedIntegrationGaps` becomes self-enforcing rather than a convention.

---

tomo-chan (commented):
2件の追加指摘です。前回のレビューで挙がっていたものの、そのセッションから GitHub へ書き込めず未投稿になっていた分で、いずれも今回このブランチ（`c7f878c`）に対して perturbation で再検証しています。既存の 6 スレッドはすべて未対応のままです。

- `newAPIEnv` の hermeticity — `HOME` だけでは足りず、`XDG_CACHE_HOME` が設定された環境ではローカルペイン作成時に browser shim が temp home の外へ書き込みます（実際に 3 ファイル生成を確認）。
- `documentedIntegrationGaps` — opt-in なので、成功系だったケースが黙ってエラー系に落ちても何も落ちません（`PUT /api/layout` を 400 アサートに書き換えてもスイートは緑）。

どちらも本番コードの不具合ではなく、テストが守っていると主張している性質と実際の担保のズレです。

---

internal/server/api_integration_test.go
tomo-chan: Fixed in 729d1a1. You're right that this was the odd one out — it asserted the fixture back to itself.

One wrinkle your suggested patch runs into: `POST /api/workspaces` already switches to the new workspace, so `PUT {"id": added}` would also have been a no-op move. The case now takes whichever of the two is *not* currently active as its target, so the assertion is that the switch moved regardless of which way round that is.

Your exact perturbation (`SetActiveWorkspace(h.cfg.Workspaces.Active)`) now fails the case.

---

internal/server/api_integration_test.go
tomo-chan: Fixed in 729d1a1, going one step past the suggestion: the case now seeds a real status (`e.cache.RecordStatus("pane-1", …)`) before the request, so it asserts `"statuses":{"pane-1":` and `"state":"working"` rather than the empty envelope. That makes it fail both against a handler returning no body and against one returning a correctly-shaped but empty map.

Your perturbation (`w.WriteHeader(http.StatusOK); return`) now fails the case.

---

internal/server/api_integration_test.go
tomo-chan: Fixed in 729d1a1, and your diagnosis of the cause was right — `testConfig()` leaves both `DisplayConfig` booleans false, so there was nothing to assert that a zero value wouldn't also satisfy.

The case now seeds `config.DisplayConfig{ShowHeader: true, ShowStatusBar: false}` after `New()` (the post-construction write the `GET /api/ssh-connections` case already relies on) and asserts both fields. Seeding one `true` and one `false` is deliberate: `true` catches the whole configuration being dropped, `false` catches a handler that emits only the set fields.

Your perturbation (`writeJSON(w, struct{}{})`) now fails the case.

---

internal/server/api_integration_test.go
tomo-chan: Applied in 729d1a1 — `registeredAPIRoutes` now unions both runner states, with a comment recording that this is load-bearing rather than thorough, since a future cleanup would otherwise read the second `New()` as redundant.

Your perturbation (`/api/command-center/ping` inside the `if commandRunner != nil` block) now fails with `every /api route must be driven through the real router; add a case to apiCases for each`.

Also updated the rollout row in `docs/quality-gateway.md`, which — as you say — was claiming more than the code delivered.

Each server built here now gets a `t.Cleanup` Shutdown too, for the reason in the goroutine thread below.

---

internal/server/api_integration_test.go
tomo-chan: Applied in 729d1a1, close to your wording. You're right that the danger is the comment inviting the override's removal rather than the comment merely being wrong, so the replacement says which defaults httptest really uses and names the guard.

Re-confirmed your perturbation after the fix: dropping `req.Host`/`req.RemoteAddr` from `do` still turns this case into `expected: 200, actual: 403`.

---

internal/server/api_integration_test.go
tomo-chan: Applied in 729d1a1, in `newAPIEnv`'s cleanup and in `registeredAPIRoutes` for the two servers it builds.

Measured rather than assumed, with a throwaway probe that creates five envs and counts goroutines either side: **delta 5 without the Shutdown call, delta 0 with it**. So the leak was real at the rate you describe, and it is closed.

Taking the point about `goleak` as the reason this is worth doing now rather than the ~30 goroutines themselves — a package that leaks by construction is one nobody can add that check to later.

---

internal/server/api_integration_test.go
tomo-chan: Applied verbatim in 729d1a1, and reproduced both directions on this branch:

```
XDG_CACHE_HOME=/tmp/probe go test ./internal/server/ -run 'TestServer_APIIntegration$' -count=1
# without the fix:
/tmp/probe/panemux/bin/panemux-open
/tmp/probe/panemux/bin/open
/tmp/probe/panemux/bin/xdg-open
# with it: nothing under /tmp/probe/panemux at all
```

The reason this is easy to miss is exactly the one you name — `$HOME/.cache` is the fallback, so the temp `HOME` covers it locally and only stops covering it somewhere that exports the variable. Left the reason in the doc comment rather than just the line, since `t.Setenv("XDG_CACHE_HOME", …)` beside `t.Setenv("HOME", …)` otherwise reads as belt-and-braces.

---

internal/server/api_integration_test.go
tomo-chan: Applied in 729d1a1, essentially as suggested: `apiEnv.statuses` is appended to in `do`, and `TestServer_APIIntegration` requires at least one 2xx for any case with no `notHappyPathReason`.

Your perturbation now fails where it should:

```
--- FAIL: TestServer_APIIntegration/PUT_/api/layout
    a case with no notHappyPathReason must drive the route's success path; observed statuses: [400]
```

Kept `documentedIntegrationGaps` as well rather than replacing it, and said why in a comment: the derived check cannot put a literal in the diff for a reviewer to notice, and the literal cannot see a case that stopped asserting a 2xx without declaring it. They cover different halves.

Agreed on the general shape — "derive the fact from the run rather than from a field the author remembers to set" is the same move as the exhaustiveness check, and it is the one that survives contact with a future contributor.

---

PR #183, opened 2026-08-23
test: widen the coverage gate's scope, threshold unchanged (#180 item 2)

tomo-chan:
Roadmap item 2 of #180 — gate **G4(a)** in [docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md).

- [x] `COVERAGE_PKGS` に `internal/portforward` / `internal/commandcenter` / `internal/boardmcp` / root パッケージを追加する
- [x] frontend の coverage `include` に `src/utils/**` を追加する
- [x] 追加で 80% を割る箇所を埋める（`board.go` / `main.go` が主対象）
- [x] `main.go` の `parseOptions` / `loadConfig` / `startSessionsFromConfig` を注入可能な単位に切り出す
- [x] 意図的に除外するものは `Makefile` にコメントで理由を残す

## The threshold does not move

Per **decision D1**: raising it works, but the cheapest way to satisfy a higher number is to generate tautological tests, which lowers Khorikov's first *and* second pillars at once. What gets strengthened is the **scope**.

| | Before | After |
|---|---|---|
| Go, gated packages | 88.3% over 5 packages | **86.2% over 9** |
| Frontend | 95.7% over hooks + schemas | **95.0% over hooks + schemas + utils** (utils 92%) |
| Threshold | 80% | 80% |

The Go number going *down* is the point — roughly 2,600 previously ungated lines are now inside the gate.

## Why `main.go` was 0%

Not because nobody wrote tests for it, but because nothing in it could be driven: `parseOptions` read the process-global `flag.CommandLine`, `loadConfig` called `config.Load` directly, and `startSessionsFromConfig` called `session.CreateFromConfig` directly. Each now takes its dependency as an argument — DEVELOPMENT.md's testability rule, principle **P3**:

| Was | Now | What that buys |
|---|---|---|
| `parseOptions()` on `flag.CommandLine` | `parseOptions(args []string) (cliOptions, error)` on its own `FlagSet` | repeatable, and callable with arguments a test chose |
| `loadConfig(opts)` | `loadConfig(opts, loader configLoader)` | both branches and the failure path, without reading the developer's own config |
| `startSessionsFromConfig(cfg, mgr)` | `+ create createSessionFunc` | the best-effort "one pane fails, the others still start" behavior, without a real PTY or SSH host |
| `openChrome` with the per-OS switch inline | pure `browserOpenArgv(goos, url)` + one unconditional `Run` | every OS branch testable — the humble-object shape **P2** asks for |

`board.go`'s helpers that only `setupBoard` reached indirectly now have direct tests, table-driven where the factors form a matrix.

| File | Before | After |
|---|---|---|
| root package | 0% (ungated) | **75%+** |
| `board.go` | 47.8% | **76.9%** |
| `main.go` | 0% | **38.8%** — the remainder is `main()` and `runServer()` themselves |

## Exclusions are reasons, not omissions — and they are checked

- **`internal/session`** — its lifecycle methods drive a real PTY, a live SSH connection and a real tmux server. `make check` stays hermetic (principle 5); the pure decisions extracted out of them are unit-tested in place.
- **`internal/sshconfig`** — a parser over the user's own `~/.ssh/config`, with its own tests and no gate-worthy branching the measured packages don't already drive.
- **Inside** the measured set, `main()` / `runServer()` / `bootstrapWatcher.Run()` stay uncovered: they install signal handlers, start the listener and run poll loops for the life of the process.

`coverage_scope_test.go` derives the expected set from `go list ./...` minus an `ungatedPackages` map naming each skipped package with its reason, so a package added later fails the suite until it is gated or explicitly excluded — the `route_table_test.go` pattern applied to the coverage gate. A second test pins that `COVERAGE_PKGS` and the recipe's own argument list agree.

## Review round (`96318a2`)

Six findings from the review, all reproduced before being acted on. Two are worth calling out:

- **`panemux --help` exited 2 instead of 0.** Moving to `flag.ContinueOnError` made help return `flag.ErrHelp` like any other error. `install.sh` tells users to run it, so a wrapper under `set -e` saw a failure. Now `parseExitCode` — a separate, tested function, because `main` is outside the gate by design.
- **Two tests named `..._UnwritableLocation_DoesNotPanic` never reached the branch they were named after.** `atomicWriteFile` starts with `os.MkdirAll`, so a missing `HOME` is created on demand and the write succeeds. Both were tautological, arriving as gated coverage in the PR that widens the gate. Fixed with a file where the directory has to go (`ENOTDIR`, root-proof), plus an assertion that the save really failed.

The rest: `coverage-go: build-frontend` (the root package embeds `frontend/dist`, so the standalone command failed on a clean tree), the derived scope test above, and a stale comment in `vite.config.ts`.

## Test plan

- [x] `make test-go` — all packages pass
- [x] `make coverage-go` — **86.2%**, above the unchanged 80% threshold, and now working on a clean tree
- [x] `make coverage-frontend` — **95.0%** with `src/utils/**` included
- [x] `make check` — passes end to end
- [x] Gate confirmed to still bite: threshold temporarily raised to 90% failed as expected, then reverted
- [x] Scope gate confirmed by perturbation: a new `internal/zzprobe` package fails by name; desynchronising `COVERAGE_PKGS` from the recipe fails both tests
- [x] Every review finding reproduced before fixing


main.go
```diff
@@ -44,14 +52,19 @@ func main() {
 		return
 	}
 
-	opts := parseOptions()
+	opts, err := parseOptions(os.Args[1:])
+	if err != nil {
+		// flag.ContinueOnError has already printed the error and the usage
+		// text; 2 is the conventional exit code for a usage error.
+		os.Exit(2)
```
tomo-chan: **`panemux --help` now exits 2 instead of 0.**

Switching from `flag.CommandLine` (which is built with `flag.ExitOnError`) to a `flag.NewFlagSet(..., flag.ContinueOnError)` changes the `-h`/`--help` path, not just the injection point. `FlagSet.Parse` special-cases help under `ExitOnError`:

```go
case ExitOnError:
    if err == ErrHelp { os.Exit(0) }
    os.Exit(2)
```

Under `ContinueOnError` it just returns `flag.ErrHelp`, which `parseOptions` wraps into an ordinary error, so `main` falls into the `os.Exit(2)` branch above. Verified against real builds of both revisions:

```
$ ./panemux-base --help ; echo $?   # 913b8a8
...usage...
0
$ ./panemux-pr183 --help ; echo $?  # bd68266
...usage...
2
```

Concrete scenario: `install.sh:200` tells the user `Try: ${PROJECT_NAME} --help`, and any wrapper/verification script running that under `set -e` (or checking `$?`) now sees a failure where the command actually did exactly what was asked. The comment on line 57 also isn't accurate for this path — for `ErrHelp` the flag package prints usage but no error message.

Suggested fix: treat help as success.

```suggestion
	opts, err := parseOptions(os.Args[1:])
	if err != nil {
		// -h/--help is not a usage error: flag has already printed the
		// usage text and there is nothing more to do. Any other parse
		// error has had its message printed too; 2 is the conventional
		// exit code for a usage error.
		if errors.Is(err, flag.ErrHelp) {
			os.Exit(0)
		}
		os.Exit(2)
	}
```

(needs `"errors"` in the import block; `parseOptions` already wraps with `%w`, so `errors.Is` works as-is.)

---

board_test.go
```diff
@@ -292,3 +293,263 @@ func TestAgmsgPresentOnHostLocal(t *testing.T) {
...
+
+// A save failure is a warning, never a crash and never a startup failure:
+// the relay's cursor is an optimisation, and losing it costs a re-read, not
+// correctness. The config directory is deliberately absent here.
+func TestPersistBoardCursors_UnwritableLocation_DoesNotPanic(t *testing.T) {
+	t.Setenv("HOME", filepath.Join(t.TempDir(), "nonexistent"))
+
+	assert.NotPanics(t, func() {
+		persistBoardCursors([]board.CursorEntry{{Host: "local", Team: "demo", Cursor: "abc"}})
+	})
```
tomo-chan: **This test never reaches the failure branch it claims to cover — the write succeeds.**

`persistBoardCursors` → `board.SaveCursorFile` → `atomicWriteFile`, whose first statement is `os.MkdirAll(dir, 0750)` (`internal/board/atomic_write.go:17`). Pointing `HOME` at `<tmp>/nonexistent` is therefore not an unwritable location: `MkdirAll` happily creates `nonexistent/.config/panemux` and the cursor file lands there. Verified by probing the real function with exactly this setup:

```
WROTE SUCCESSFULLY to /tmp/.../001/nonexistent/.config/panemux/board-relay-cursor.json
  -> failure branch NOT exercised
```

So `assert.NotPanics` wraps a call that took the *happy* path — the test is indistinguishable from `TestPersistBoardCursors_WritesAFileLoadCursorFileCanReadBack` above it, and the `log.Printf("Warning: agent board: persisting relay cursor: ...")` branch stays at 0%. Deleting that branch, or changing it to a `log.Fatalf`, would not fail this test. That is exactly the tautological-coverage failure mode this PR's own description (decision D1) and G4(b) red-check argue against, so it's worth fixing here rather than letting it ride in as gated coverage.

A genuinely unwritable location works and is root-proof (`ENOTDIR` is not bypassable by uid 0) — verified, the warning is emitted and no file appears:

```suggestion
func TestPersistBoardCursors_UnwritableLocation_DoesNotPanic(t *testing.T) {
	home := t.TempDir()
	t.Setenv("HOME", home)
	// atomicWriteFile MkdirAll's the whole chain, so an absent directory is
	// not unwritable. A plain FILE where the directory must go is: MkdirAll
	// fails with ENOTDIR, and no uid — including root — can get past it.
	require.NoError(t, os.WriteFile(filepath.Join(home, ".config"), []byte("not a directory"), 0o600))

	assert.NotPanics(t, func() {
		persistBoardCursors([]board.CursorEntry{{Host: "local", Team: "demo", Cursor: "abc"}})
	})

	path, err := board.DefaultCursorFilePath()
	require.NoError(t, err)
	_, statErr := os.Stat(path)
	assert.Error(t, statErr, "the save must actually have failed for this test to mean anything")
}
```

---

board_test.go
```diff
@@ -292,3 +293,263 @@ func TestAgmsgPresentOnHostLocal(t *testing.T) {
...
+	got, err := board.LoadBootstrapState(path)
+	require.NoError(t, err)
+	assert.Equal(t, []string{"pane-a", "pane-b"}, got)
+}
+
+func TestPersistBootstrapState_UnwritableLocation_DoesNotPanic(t *testing.T) {
+	t.Setenv("HOME", filepath.Join(t.TempDir(), "nonexistent"))
+
+	assert.NotPanics(t, func() { persistBootstrapState([]string{"pane-a"}) })
+}
```
tomo-chan, line 485: **Same problem as the cursor test above: the "unwritable location" is writable, so the save-failure branch is never exercised.**

`persistBootstrapState` → `board.SaveBootstrapState` goes through the same `atomicWriteFile`, which starts with `os.MkdirAll(dir, 0750)`. `HOME = <tmp>/nonexistent` just gets created on demand, the state file is written, and `assert.NotPanics` passes on the happy path. The `log.Printf("Warning: agent board bootstrap: persisting bootstrap state: ...")` line stays uncovered, and removing it would not fail this test.

Same fix shape as the cursor one — put a regular file where the directory needs to be, then assert the write really did fail:

```suggestion
func TestPersistBootstrapState_UnwritableLocation_DoesNotPanic(t *testing.T) {
	home := t.TempDir()
	t.Setenv("HOME", home)
	// See the cursor test above: an absent directory is not unwritable,
	// because atomicWriteFile MkdirAll's it. A file in its place is.
	require.NoError(t, os.WriteFile(filepath.Join(home, ".config"), []byte("not a directory"), 0o600))

	assert.NotPanics(t, func() { persistBootstrapState([]string{"pane-a"}) })

	path, err := board.DefaultBootstrapStateFilePath()
	require.NoError(t, err)
	_, statErr := os.Stat(path)
	assert.Error(t, statErr, "the save must actually have failed for this test to mean anything")
}
```

---

Makefile
```diff
@@ -76,11 +103,15 @@ coverage-go:
 	  ./internal/ws/... \
 	  ./internal/server/... \
 	  ./internal/board/... \
+	  ./internal/portforward/... \
+	  ./internal/commandcenter/... \
+	  ./internal/boardmcp/... \
+	  . \
```
tomo-chan, line 116: **Adding `.` here makes `make coverage-go` hard-fail on a clean tree, because it pulls `//go:embed frontend/dist` into the target.**

`main.go:39` embeds `frontend/dist`, so compiling the root package requires that directory to exist. Before this PR `coverage-go` only compiled `internal/...` and had no such dependency — and DEVELOPMENT.md lists `make coverage-go` as a standalone command. Reproduced:

```
$ make clean && make coverage-go
main.go:40:12: pattern frontend/dist: no matching files found
FAIL	panemux [setup failed]
make: *** [Makefile:100: coverage-go] Error 1
```

`make check` and CI are unaffected (both run `build-frontend` first), so this only bites someone running the documented standalone command after `make clean` — but the error names an embed pattern, which gives no hint that the fix is "build the frontend first". Cheapest fix is to make the prerequisite explicit on the `coverage-go:` line (`coverage-go: build-frontend`), or to record it in the comment block above and in DEVELOPMENT.md's coverage section if you'd rather not pay an npm build on every `coverage-go`.

---

Makefile
```diff
@@ -56,16 +56,43 @@ test-agmsg-contract:
...
+#     they used to embed has been extracted into the injectable units around
+#     them (parseOptions, loadConfig, startSessionsFromConfig,
+#     browserOpenArgv), which is where the gate applies.
+#
+# Frontend: measured over src/hooks/, src/schemas/ and src/utils/.
 #           UI components (App, SplitContainer, TerminalPane …) require a real
 #           browser renderer and are covered by integration / E2E tests.
 
-COVERAGE_PKGS := ./internal/config/...,./internal/api/...,./internal/ws/...,./internal/server/...,./internal/board/...
+COVERAGE_PKGS := ./internal/config/...,./internal/api/...,./internal/ws/...,./internal/server/...,./internal/board/...,./internal/portforward/...,./internal/commandcenter/...,./internal/boardmcp/...,.
```
tomo-chan, line 95: **The gate's scope is a hand-maintained allowlist, so the next package added to the repo is silently ungated — the same class of drift PR #181 just removed from the route table.**

The package set now exists in three hand-written places that must agree: `COVERAGE_PKGS`, the `go test` argument list below, and the prose in DEVELOPMENT.md. Nothing checks any of them against `go list ./...`.

Concrete scenario: someone adds `internal/foo`, forgets this line, and `make coverage-go` still prints a healthy 86% while `internal/foo` sits at 0% — the gate reports success precisely because the new code is invisible to it. That is a strictly worse failure than the one this PR is closing, because the number goes *up*, not down. It is also the exact failure mode the Makefile comment guards against for `internal/session`/`internal/sshconfig` by writing reasons down: the guard covers today's two exclusions but nothing about tomorrow's additions.

`internal/server/route_table_test.go` already established the pattern for this repo — derive the list from the source of truth (`chi.Walk` there) instead of naming it. The equivalent here is a small Go test that runs `go list ./...`, subtracts an explicitly-named exclusion set (`internal/session`, `internal/sshconfig`), and fails when the remainder isn't exactly what `COVERAGE_PKGS` names — which would then let a new package's absence fail on the day it is added rather than whenever someone next reads this file.

---

frontend/vite.config.ts
```diff
@@ -27,7 +27,7 @@ export default defineConfig({
       // Measure only hooks and schemas — UI components and entry points
       // (App.tsx, main.tsx, components/) require browser rendering and are
       // covered by integration/E2E tests, not unit tests.
-      include: ['src/hooks/**', 'src/schemas/**'],
+      include: ['src/hooks/**', 'src/schemas/**', 'src/utils/**'],
```
tomo-chan, line 31: The comment immediately above this line still says **"Measure only hooks and schemas"**, which now contradicts the line it documents. Every other scope statement in this PR was updated (`Makefile`, `DEVELOPMENT.md`, `docs/quality-gateway.md`); this is the one place a reader is told the opposite of what the config does — and it's the file they'd open first to find out what is gated.

```suggestion
      // Measure hooks, schemas and utils — the modules that hold decisions.
      // UI components and entry points (App.tsx, main.tsx, components/)
      // require browser rendering and are covered by integration/E2E tests,
      // not unit tests.
      include: ['src/hooks/**', 'src/schemas/**', 'src/utils/**'],
```

---

tomo-chan (commented):
Reviewed the full diff (main.go, Makefile, vite.config.ts, board_test.go, main_test.go, docs) against a checked-out copy of `bd68266`, and verified the gate itself end to end: `go vet ./...` clean, `make lint-go` 0 issues, `go test .` green, `make coverage-go` 86.0%, `make coverage-frontend` 95.03/84.39/96.29/97.61 — all the numbers in the PR body reproduce, including the root package's 75.1%.

The extraction work is sound: `parseOptions` / `loadConfig` / `startSessionsFromConfig` / `browserOpenArgv` are the right seams, and `browserOpenArgv` preserves the previous per-OS argv exactly.

Six findings, one of them a real behavior regression:

1. **`main.go`** — `panemux --help` now exits **2** instead of **0**. `ContinueOnError` returns `flag.ErrHelp` instead of `os.Exit(0)`-ing it, and `main` treats that like a usage error. Verified against builds of both revisions. `install.sh` itself tells users to run `panemux --help`.
2. **`board_test.go`** — `TestPersistBoardCursors_UnwritableLocation_DoesNotPanic` never reaches the save-failure branch: `atomicWriteFile` starts with `os.MkdirAll`, so `HOME=<tmp>/nonexistent` is perfectly writable and the file gets written. Confirmed by probing the real function.
3. **`board_test.go`** — same for `TestPersistBootstrapState_UnwritableLocation_DoesNotPanic`.
4. **`Makefile`** — adding `.` makes `make coverage-go` fail outright on a clean tree (`pattern frontend/dist: no matching files found`).
5. **`Makefile`** — `COVERAGE_PKGS` is a hand-maintained allowlist, so a package added later is silently ungated and the reported percentage goes *up*.
6. **`frontend/vite.config.ts`** — the comment still says "Measure only hooks and schemas".

(2) and (3) are worth fixing in this PR specifically because they are the failure mode its own decision D1 argues against: they add gated coverage on the happy path while the error branch they are named after stays at 0%.

---

main.go
tomo-chan: Confirmed and fixed in `96318a2`. Reproduced first: `parseOptions([]string{"--help"})` returns `parsing command line options: flag: help requested` with `errors.Is(err, flag.ErrHelp) == true`, so `main` was indeed taking the `os.Exit(2)` branch.

I took the fix one step further than the suggestion and extracted it rather than inlining the `errors.Is`:

```go
func parseExitCode(err error) int {
	if errors.Is(err, flag.ErrHelp) {
		return 0
	}
	return 2
}
```

The reason is this PR's own argument: `main` is deliberately outside the coverage gate because it installs signal handlers and runs for the life of the process, so an inline check there would be a behavior rule nothing could test. As a function it gets `TestParseExitCode` — both help spellings, two real parse errors, and an unrelated error.

I also added `TestParseOptions_Help_WrapsErrHelp`, because your fix depends on a property that is invisible at the call site: if `parseOptions`' `%w` wrapping were ever changed to `%v`, `parseExitCode` would silently start answering 2 for `--help` again and nothing else would notice.

Your note about the comment on line 57 was right too — for `ErrHelp` the flag package prints usage but no error message — so that comment now lives on `parseExitCode` and distinguishes the two cases.

---

board_test.go
tomo-chan: You're right, and I reproduced it before changing anything — probing the real function with exactly this setup:

```
CLAIM CONFIRMED: file WAS written to /tmp/TestProbeUnwritable.../001/nonexistent/.config/panemux/board-relay-cursor.json
  -> failure branch NOT exercised
```

Applied your suggestion in `96318a2`, for both this test and the bootstrap-state one.

Worth stating plainly rather than quietly fixing: this is a tautological test, and it arrived in the PR whose own description argues against exactly that (D1), inside the change that widens the gate it would have been counted by. The `assert.NotPanics` had nothing to lose — the `log.Printf` branch could have been deleted outright and the test would still have passed. The fix matters less than the shape being on the record.

The comment now says why an absent directory is not unwritable and a file in its place is, so the next person does not re-derive it.

---

board_test.go
tomo-chan: Same fix applied in `96318a2` — same root cause (`atomicWriteFile`'s `os.MkdirAll`), same shape of fix. Its comment points at the cursor test above rather than repeating the explanation.

---

Makefile
tomo-chan: Reproduced (moved `frontend/dist` aside, `go build ./...` → `main.go:40:12: pattern frontend/dist: no matching files found`) and fixed in `96318a2` with the explicit prerequisite: `coverage-go: build-frontend`.

I went with the prerequisite rather than documenting the requirement because DEVELOPMENT.md lists `make coverage-go` as a standalone command — a documented command that needs an undocumented preceding step is the same class of problem in a different place. `make check` and CI already run `build-frontend` first, so they pay nothing for it.

Verified on a genuinely clean tree: `make coverage-go` alone now builds the frontend and reports 86.2%.

---

Makefile
tomo-chan: Agreed, and the asymmetry you name is the part that makes it worse than the failure this PR was closing: an ungated package makes the number go **up**, so the gate reports success *because* the new code is invisible to it.

Implemented in `96318a2` as `coverage_scope_test.go`, following the `route_table_test.go` pattern you point at:

- **`TestCoverageScopeCoversEveryPackage`** — `go list ./...` minus an `ungatedPackages` map that names each skipped package **with its reason**, compared against the expansion of `COVERAGE_PKGS`. Skipping a package stays possible, but only as a visible edit to that map, never by forgetting the Makefile.
- **`TestCoverageRecipeRunsEveryGatedPackage`** — you counted three hand-written places; this pins two of them against each other. They mean different things and both matter: `COVERAGE_PKGS` feeds `-coverpkg` (what is *measured*), the recipe's arguments decide what is *run*. A package in the first but not the second is measured, never exercised, and reports 0% — dragging the gate down for no reason and inviting someone to "fix" it by narrowing the scope.

Verified by perturbation rather than assumed:

| Perturbation | Result |
|---|---|
| Add `internal/zzprobe` and gate nothing | red — the diff names `panemux/internal/zzprobe` |
| Drop `./internal/boardmcp/...` from `COVERAGE_PKGS` only | red — **both** tests |

The third place, DEVELOPMENT.md's prose, is still hand-written. I left it: turning a sentence meant for humans into a parsed artifact costs more than it protects, and it can now only drift about the *reasons*, not the set — the set is checked. Say the word if you'd rather it were derived too.

---

frontend/vite.config.ts
tomo-chan: Applied verbatim in `96318a2`. You're right that this is the worst place to leave a stale scope statement — it's the file someone opens to find out what is gated, and it said the opposite of the line directly beneath it.

---

main.go
```diff
@@ -44,14 +53,17 @@ func main() {
 		return
 	}
 
-	opts := parseOptions()
+	opts, err := parseOptions(os.Args[1:])
+	if err != nil {
+		os.Exit(parseExitCode(err))
+	}
```
tomo-chan, line 59: `--help` now exits 2 instead of 0.

`flag.ErrHelp` is not distinguished from a real parse error here. The old global `flag.CommandLine` used `ExitOnError`, whose `Parse` does `if err == ErrHelp { os.Exit(0) }`; the new `ContinueOnError` FlagSet just returns `ErrHelp`, which lands in this branch.

Verified by building both revisions with the same stub `frontend/dist`:

```
913b8a8 ./bin-old --help  -> usage printed, exit 0
bd68266 ./bin-panemux --help -> same usage printed, exit 2
```

Any wrapper script under `set -e`, a package-manager smoke test, or a `panemux --help >/dev/null` check now fails. The comment is also inaccurate for this path: on `ErrHelp` the flag package printed usage only (no error), and the user made no usage error.

```suggestion
	opts, err := parseOptions(os.Args[1:])
	if err != nil {
		// -h/--help is not a usage error: flag has already printed the
		// usage text, and the conventional exit code for it is 0.
		if errors.Is(err, flag.ErrHelp) {
			os.Exit(0)
		}
		// flag.ContinueOnError has already printed the error and the usage
		// text; 2 is the conventional exit code for a usage error.
		os.Exit(2)
	}
```

(needs `errors` in the import block, and a `TestParseOptions_Help_ReturnsErrHelp` case pinning `errors.Is(err, flag.ErrHelp)` so the distinction stays testable at the `parseOptions` level.)

---

board_test.go
```diff
@@ -292,3 +293,291 @@ func TestAgmsgPresentOnHostLocal(t *testing.T) {
...
+	require.NoError(t, err)
+	got, err := board.LoadCursorFile(path)
+	require.NoError(t, err)
+	assert.Equal(t, entries, got)
+}
+
+// A save failure is a warning, never a crash and never a startup failure:
+// the relay's cursor is an optimisation, and losing it costs a re-read, not
+// correctness.
+//
```
tomo-chan, line 421: This test exercises the success path, not the failure branch it names.

`SaveCursorFile` -> `atomicWriteFile` starts with `os.MkdirAll(filepath.Dir(path), 0750)`, so a `HOME` pointing at a not-yet-existing directory is simply created and the write succeeds. Verified with a probe test on this branch:

```
PROBE: file WAS written at <tmp>/001/nonexistent/.config/panemux/board-relay-cursor.json
       -> save succeeded, failure branch never taken
```

Consequences:

- `persistBoardCursors`' `log.Printf("Warning: agent board: persisting relay cursor: %v", err)` is never reached — `go tool cover` reports `persistBoardCursors 50.0%` on this branch, which is exactly that branch missing.
- `assert.NotPanics` around code that cannot fail can never go red, so a regression turning that warning into `log.Fatalf`, or a nil deref on the error path, passes.
- The comment "The config directory is deliberately absent here" is false: it is created by the code under test.

Since this PR is gate G4(a) and decision D1 is specifically about not buying coverage with tests that cannot fail, this one is worth making real. `internal/config/agent_board_test.go`'s `TestEnsureAuthToken_WriteFailure_NonFatal_LoopbackHost` already has the pattern — make the parent path un-`MkdirAll`-able (a regular file where the directory should be, or a `0500` parent) rather than merely absent.

---

board_test.go
```diff
@@ -292,3 +293,291 @@ func TestAgmsgPresentOnHostLocal(t *testing.T) {
...
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	require.NoError(t, os.WriteFile(filepath.Join(home, ".config"), []byte("not a directory"), 0o600))
+
+	assert.NotPanics(t, func() {
+		persistBoardCursors([]board.CursorEntry{{Host: "local", Team: "demo", Cursor: "abc"}})
+	})
+
+	path, err := board.DefaultCursorFilePath()
+	require.NoError(t, err)
```
tomo-chan, line 457: Same defect as the cursor twin above: `SaveBootstrapState` also goes through `atomicWriteFile`, which `os.MkdirAll`s the parent, so the "nonexistent" `HOME` is created and the write succeeds. Probe on this branch:

```
PROBE2: bootstrap file WAS written at <tmp>/001/nonexistent/.config/panemux/board-bootstrap-state.json
```

`persistBootstrapState` reports `50.0%` in the coverage profile for exactly this reason — the `log.Printf("Warning: agent board bootstrap: persisting bootstrap state: %v", err)` branch is never executed. `assert.NotPanics` over a path with no reachable failure is a test that cannot go red.

Worth fixing alongside the cursor one, ideally through a shared helper that produces a genuinely unwritable target so both stay honest.

---

main.go
```diff
@@ -167,18 +236,36 @@ func startSessionsFromConfig(cfg *config.Config, manager *session.Manager) error
...
-		cmd = exec.Command("google-chrome", "--app="+url) //nolint:gosec // G204: URL is local server host/port
-	case "windows":
-		cmd = exec.Command("cmd", "/c", "start", "chrome", url)
+// browserOpenArgv returns the command that opens url in Chrome on goos, as a
+// program name and discrete arguments — never a shell string. Splitting the
+// decision out from the exec call is the humble-object shape
+// docs/quality-gateway.md's P2 asks for: every per-OS branch is testable
+// here, and what is left below is one unconditional Run.
+//
+// ok is false for an OS panemux has no opener for, in which case nothing is
```
tomo-chan, line 245: This makes `openChrome` a third `exec.Command` site whose first argument is not a compile-time literal, and `docs/security.md` still says there are only two.

docs/security.md:159:

> `internal/commandcenter/runner.go`'s `Runner` is the one other place in this repository, besides `internal/session`, that calls `exec.Command`/`exec.CommandContext` on a value not fully known at compile time.

That sentence is now inaccurate. Related: the suppression widened from the branch-scoped `//nolint:gosec // G204: URL is local server host/port` on the Linux line to an unqualified `//nolint:gosec` covering all three OS branches plus the variable program name.

The value is still derived from three hardcoded literals via `browserOpenArgv`, so I do not think this is exploitable — the concrete cost is that a later security review or CodeQL triage reading that enumeration concludes no other variable-first-argument sink exists and skips this one. DEVELOPMENT.md's "Documentation updates" and "Security-sensitive implementation" rules ask for the security.md update in the same change; a sentence naming `browserOpenArgv` as the literal-only source (and why the taint chain is unbroken) would settle it. Keeping a reason on the nolint itself (`//nolint:gosec // G204: name comes from browserOpenArgv's literals; args are discrete, never a shell`) would also preserve what the old comment carried.

---

frontend/vite.config.ts
```diff
@@ -24,10 +24,11 @@ export default defineConfig({
...
     coverage: {
       provider: 'v8',
-      // Measure only hooks and schemas — UI components and entry points
-      // (App.tsx, main.tsx, components/) require browser rendering and are
-      // covered by integration/E2E tests, not unit tests.
-      include: ['src/hooks/**', 'src/schemas/**'],
+      // Measure hooks, schemas and utils — the modules that hold decisions.
+      // UI components and entry points (App.tsx, main.tsx, components/)
+      // require browser rendering and are covered by integration/E2E tests,
+      // not unit tests.
```
tomo-chan, line 30: The comment three lines above still says "Measure only hooks and schemas", which this line now contradicts.

Concrete cost: someone adding a file under `frontend/src/utils/` reads that comment, concludes utils sit outside the gate the way `components/` does, and ships it untested — then the 80% frontend threshold fails in CI for a reason the comment actively denied. The Makefile and DEVELOPMENT.md both got their scope wording updated in this PR; this one did not.

```suggestion
      // Measure hooks, schemas and utils — the decision-holding modules.
      // UI components and entry points (App.tsx, main.tsx, components/)
      // require browser rendering and are covered by integration/E2E tests,
      // not unit tests.
      include: ['src/hooks/**', 'src/schemas/**', 'src/utils/**'],
```

---

tomo-chan (commented):
Reviewed the coverage-scope widening. The refactor itself is sound — `parseOptions`/`loadConfig`/`startSessionsFromConfig`/`browserOpenArgv` are the right seams, `golangci-lint run ./...` is clean, and I reproduced the gate locally: **86.1%** over the widened set, so the threshold holds.

Five findings, one of them a real behavior regression:

1. **main.go — `--help` now exits 2 instead of 0.** `flag.ErrHelp` isn't distinguished from a parse error. Confirmed by building both revisions (old: exit 0, new: exit 2).
2. **board_test.go — `TestPersistBoardCursors_UnwritableLocation_DoesNotPanic` never reaches the failure branch.** `atomicWriteFile` does `os.MkdirAll`, so the "nonexistent" HOME is created and the save succeeds; verified with a probe. Coverage agrees: `persistBoardCursors 50.0%`.
3. **board_test.go — the bootstrap twin has the same shape**, same probe result, `persistBootstrapState 50.0%`.
4. **main.go — `openChrome` is now a third variable-first-argument `exec.Command` site**, and docs/security.md:159 still says there are only two; the `//nolint:gosec` also lost its reason and widened to all three branches.
5. **frontend/vite.config.ts — the comment above `include` still says "only hooks and schemas".**

(2) and (3) are the ones I'd push on hardest, since this PR's whole thesis (decision D1) is that scope beats threshold precisely because tests that cannot fail buy coverage without buying protection — and these two are that pattern.

---

main.go
tomo-chan: This is the one finding in this round that was genuinely new — the other four had already landed in `96318a2`, which was pushed a few minutes before this review was submitted against the previous head. Fixed in `6773bbb`.

You're right on both halves, and right that it isn't exploitable. What made it a real problem is exactly what you name: the sentence is an **enumeration**, and an enumeration that has silently become wrong is worse than no enumeration, because the next reader stops looking. `browserOpenArgv` did not just move code — before the split each branch passed a literal, so no variable-first-argument sink existed in `main.go` at all. This PR created one and left the document asserting it hadn't.

`docs/security.md` now has a **"Launching the operator's browser (`--open`)"** section stating the argument — `browserOpenArgv` is pure, is the only source of both values, returns one of three compile-time literals or reports the OS unsupported (in which case nothing launches), `url` is panemux's own listen address, every value is a discrete argv element — and the Command center section's opening sentence enumerates all three sites rather than two.

The `//nolint:gosec` carries its reason again, as you suggested:

```go
//nolint:gosec // G204: name is one of browserOpenArgv's three literals; args are discrete, never a shell
cmd := exec.Command(name, args...)
```

It sits on its own line rather than trailing the call only because `lll` rejected the combined length.

Worth noting for the record: DEVELOPMENT.md's "Security-sensitive implementation" rule says to read `docs/security.md` **before** implementing a change that touches `exec.Command` posture. I did not, because the change registered to me as a testability refactor rather than a security-relevant one — which is precisely when that rule matters, and the reason it is written as "when code changes affect command execution" rather than "when you are working on security".

---

board_test.go
tomo-chan: Already fixed in `96318a2` (pushed shortly before this review landed, so it was reviewing the previous head) — the failure is now forced with a regular file where `.config` has to go, plus an assertion that the save really failed.

Your pointer to `internal/config/agent_board_test.go`'s `TestEnsureAuthToken_WriteFailure_NonFatal_LoopbackHost` is useful and I've cited it in the helper's comment: it means this repository already had the pattern and these two tests simply didn't reuse it.

The `50.0%` in the coverage profile is the detail I'd have wanted and missed — the number was visible the whole time and said exactly which branch was unreached. It's now recorded in the comment so the next reader has the evidence rather than the conclusion.

`21d6820` also takes your suggestion from the bootstrap thread and pulls the fixture into a shared `unwritableHome(t)` helper.

---

board_test.go
tomo-chan: Fixed in `96318a2`, and your shared-helper suggestion taken in `21d6820`: both tests now call `unwritableHome(t)` rather than each building the fixture by hand.

That is the better shape for the reason you give — with the setup duplicated, one of the two silently going green again is invisible, which is the failure being guarded against in the first place.

Red-checked rather than assumed: neutering the helper so it no longer creates the blocking file makes **both** tests fail with `the save must actually have failed for this test to mean anything`. They can go red now.

---

main.go
tomo-chan: Already fixed in `96318a2` — this review landed against the previous head, a few minutes after the fix was pushed.

Same diagnosis, and I took your `TestParseOptions_Help_ReturnsErrHelp` suggestion under a slightly different name (`TestParseOptions_Help_WrapsErrHelp`), for the reason you give: the distinction has to stay testable at the `parseOptions` level, since `%w` → `%v` would silently break it and nothing else would notice.

One difference from the suggested patch: the classification is a small function, `parseExitCode`, rather than inline in `main`. `main` is deliberately outside the coverage gate (signal handlers, process-lifetime loops), so an inline `errors.Is` would be a behavior rule nothing could test — the same argument this PR makes for `parseOptions`/`loadConfig`.

---

frontend/vite.config.ts
tomo-chan: Already fixed in `96318a2`, applied close to verbatim.

The concrete cost you spell out is the part I'd underweighted: not "a stale comment" but a comment that actively misleads someone into shipping an untested file under `src/utils/`, who then hits a CI failure the comment told them couldn't apply to them.

---

main.go
tomo-chan: Withdrawing this — it was posted against `bd68266` while the head was already `96318a2`, which had fixed it. Duplicate of the earlier thread, my mistake.

Verified on the current head by building it and running the binary: `--help` → 0, `-h` → 0, an unknown flag → 2, through `parseExitCode` and `errors.Is(err, flag.ErrHelp)`.

---

board_test.go
tomo-chan: Withdrawing this — posted against `bd68266`, already fixed in `96318a2`. Duplicate of the earlier thread.

Verified on the current head that the fix is real, not just present. The coverage profile shows the save-failure branch executing: `board.go:178.60,180.3` has count **1** (it was 0 before), and `persistBoardCursors` moved 50.0% → 66.7% — the remaining uncovered block is the `DefaultCursorFilePath()` error branch, a different, pre-existing gap.

More to the point, the test can now go red: removing the setup line that puts a regular file where `.config` has to go makes it fail with `the save must actually have failed for this test to mean anything`.

---

board_test.go
tomo-chan: Withdrawing this — posted against `bd68266`, already fixed in `96318a2`. Duplicate of the earlier thread.

Same verification as the cursor twin: `board.go:189.64,191.3` (the `persisting bootstrap state` warning) now has count **1** in the coverage profile, `persistBootstrapState` is at 66.7%, and removing the `ENOTDIR` setup line makes the test fail rather than silently pass.

---

frontend/vite.config.ts
tomo-chan: Withdrawing this — posted against `bd68266`, already fixed in `96318a2`, which is the only frontend change in that commit. The comment now reads "Measure hooks, schemas and utils — the modules that hold decisions." Duplicate of the earlier thread.

---

main.go
tomo-chan: Two corrections to my own comment, both checked against the current head.

**The suppression did not lose its reason — I misread it.** There is a three-line G204 rationale directly above the call (`name` is one of the three literals in `browserOpenArgv`, `url` is panemux's own listen address, both travel as discrete argv elements, never through a shell). That is more than the single trailing `// G204: URL is local server host/port` on `main`'s Linux branch carried, not less. The comment block was already there in `bd68266`; the claim was wrong when I made it.

**The docs/security.md point is weaker than I put it.** That sentence says "a value not fully known at compile time", and `name` holds one of three compile-time literals — only the *selection* is runtime. So it is not clearly false as written. What is true is that `gosec` treats the call as non-literal (hence the suppression at all), so a reader triaging G204 sites from that enumeration would not be pointed here.

That leaves a judgement call rather than a defect: whether that paragraph is worth a clause naming `browserOpenArgv` as a literal-only source. Leaving this thread open for you to decide rather than resolving it — the other four threads from that round were stale duplicates and are now closed.

---

tomo-chan: Retracting the comment I just posted above — I wrote it against `96318a2` without noticing the branch had moved to `21d6820`, and both of its claims are wrong.

**The original finding stands as written.** I said the `//nolint:gosec` "did not lose its reason — I misread it". It did: `main` had `//nolint:gosec // G204: URL is local server host/port` on the directive itself, and `bd68266` left a bare `//nolint:gosec` with the rationale in a separate comment block above. Those are not the same thing — a directive's own reason is what `nolintlint` and the next reader see attached to the suppression. `6773bbb` restores it inline, which is exactly what the finding asked for.

**And the docs half was already done, not a judgement call for you.** `6773bbb` adds the "Launching the operator's browser (`--open`)" section and rewrites the Command center section's opening to say "besides `internal/session` and `openChrome` above", so the enumeration is no longer wrong. I described that as still open because I was reading a commit two behind.

Verified on `21d6820`: `go vet` clean, `golangci-lint run ./...` 0 issues, root package green under `-race -count=2`. Also red-checked `unwritableHome` from `21d6820` — neutering the helper fails both persist tests with `the save must actually have failed for this test to mean anything`, as your reply said.

---

PR #184, opened 2026-08-23
ci: add the .claude agent-side gates G1, G2 and G6 (#180 item 3)

tomo-chan:
Roadmap item 3 of #180 — gates **G1** (edit) and **G6** (adversarial) in [docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md), plus G2 at the Stop hook.

- [x] `.claude/settings.json` に `PostToolUse` hook を追加する（編集したファイルだけに `gofmt -s` / `tsc --noEmit` / 対象パッケージの `go vet`）
- [x] `Stop` hook に G1+G2 相当（秒オーダー）を置く。`make check` 全体は置かない
- [x] `.claude/agents/` に差分だけを見るレビュー用サブエージェントを追加する
- [x] `.gitignore` に `.claude/settings.local.json` を追加する
- [x] `AGENTS.md` から新しい強制手段を参照できるようにする

## Why

This repository had **no `.claude/` directory at all**. Every agent-side guard in `AGENTS.md` and `DEVELOPMENT.md` sat at rung **L0** of the enforcement ladder — advisory to humans and agents alike, with nobody able to confirm it was followed. This promotes the mechanically checkable ones to **L2**.

## What lands

| File | Gate | What it does |
|---|---|---|
| `.claude/hooks/post-edit-check.sh` | G1 | `PostToolUse`. Checks **the one file just written**: `gofmt -s` + `go vet` on its package for Go, `tsc --noEmit` for frontend TypeScript. |
| `.claude/hooks/stop-check.sh` | G1 + G2 | `Stop`. Checks **what the turn changed**: formatting, `go test` on the touched packages, `tsc` + `vitest related` for touched frontend modules. Refuses to end the turn while any of it fails. |
| `.claude/agents/diff-reviewer.md` | G6 | Reviews a branch's diff in a fresh context. Reports; does not block. |
| `.claude/hooks/hooks_test.sh` | — | Tests the hooks. Wired as `make test-hooks`, folded into `make test`. |

Scope is the point throughout: `make lint` covers the repository and takes tens of seconds; these run on what actually changed.

## The Stop hook is deliberately not `make check`

**Decision D6.** Running the frontend build and the whole `-race` suite on every turn sacrifices exactly the fast feedback that makes a gate worth having, and a gate that costs too much gets bypassed. G3 onward stays with `.githooks/pre-push` and CI, where the expensive checks belong and cannot be skipped anyway.

That absence is **asserted by a test**, not left to reviewer memory — and the test strips comment lines first, because the script's own header explains why it is not `make check` and matching that sentence would have been a false positive in the very test that guards against them.

## The reviewer does not block, and says so

Per **decision D5**, `diff-reviewer.md` is told outright that it is not scored on how much it finds, and that "the diff is sound" is a complete review. A reviewer rewarded for finding gaps finds them in healthy work too, and the change made to satisfy such a finding is usually worse than the one before it — extra abstraction, defensive code, tests for cases that cannot occur.

It is pointed specifically at the two shapes this repository's existing gates cannot see: **tautological tests** ("what would have to break for this to fail?") and **tests that reconstruct production wiring** instead of travelling through it — the P1 violation that #178 found, where 161 tests passed against a router that did not exist in production.

## The hooks are tested — in both directions

A hook that silently stops working is worse than no hook: it reports the discipline as enforced while enforcing nothing. A hook that blocks a healthy change is worse still — **design principle 4**, and the failure mode that takes every other gate down with it. So `hooks_test.sh` asserts 16 checks covering both:

- blocks: unformatted Go via `--file`, via hook JSON on stdin, and via the `tool_response.filePath` spelling; unformatted Go in a dirty tree at Stop time
- **passes**: formatted Go, non-code files, missing paths, empty paths, malformed hook input, a clean tree, and a healthy change
- wiring: `settings.json` is valid JSON (a malformed one silently disables every setting in it) and actually points at both scripts

Hermetic — temp files and throwaway git repositories, never this checkout.

## Verified by perturbation, not assumed

| Perturbation | Result |
|---|---|
| Append `make check` to the Stop hook | red — D6 assertion fires |
| Point `PostToolUse`'s command at `true` | red — "PostToolUse is not wired to post-edit-check.sh" |

The exact command strings `settings.json` ships were also run **the way Claude Code invokes them** — hook JSON on stdin, `CLAUDE_PROJECT_DIR` set — not just via the test's `--file` path: `post-edit-check.sh` on a clean `main.go` exits 0, and `stop-check.sh` on this tree exits 0.

## Test plan

- [x] `make test-hooks` — all 16 checks pass
- [x] Both perturbations above confirmed red, then reverted
- [x] Real settings.json command strings exercised with hook JSON on stdin
- [x] `make lint` — 0 issues
- [x] `make check` — passes end to end (now including `test-hooks`)

---


.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+# Scope is what the working tree says changed, not the whole repository, so an
+# untouched package is never re-run.
+#
+# Exit codes are the Claude Code hook contract: 0 = the turn may end, 2 =
+# blocking, reason on stderr. Anything that cannot be checked exits 0 (design
+# principle 4).
+
+set -u
+
+changed=$(git status --porcelain 2> /dev/null | awk '{ print $NF }')
```
tomo-chan: **A brand-new package directory is invisible to this gate.**

`git status --porcelain` collapses an untracked *directory* into a single `?? dir/` entry, so none of the files inside it ever reach the `[ -f "$f" ]` check on line 27.

Reproduced in a throwaway repo:

```
$ mkdir newpkg && printf 'package newpkg\n\nfunc S() int {\n    return 1\n}\n' > newpkg/s.go   # space-indented
$ git status --porcelain
?? newpkg/
$ sh stop-check.sh; echo $?
0          # wanted 2

$ git add newpkg
$ git status --porcelain
A  newpkg/s.go
$ sh stop-check.sh; echo $?
2          # correct
```

So the single largest kind of change — adding a whole new package — passes G1+G2 while enforcing nothing, which is exactly the failure mode the Makefile comment on `test-hooks` cites. `--untracked-files=all` fixes it:

```suggestion
changed=$(git status --porcelain --untracked-files=all 2> /dev/null | awk '{ print $NF }')
```

---

.claude/hooks/post-edit-check.sh
```diff
@@ -0,0 +1,95 @@
...
+	# here, so the vet step is skipped rather than guessed at.
+	case "$file" in
+	"$repo_root"/*)
+		command -v go > /dev/null 2>&1 || exit 0
+		pkg_dir=$(dirname "$file")
+		rel=${pkg_dir#"$repo_root"}
+		rel=${rel#/}
+		[ -n "$rel" ] || rel="."
+
+		vet_output=$(cd "$repo_root" && go vet "./$rel" 2>&1)
```
tomo-chan: **`go vet` blocks every test-first edit, which is the workflow `DEVELOPMENT.md` mandates.**

`go vet` fails when the package does not compile, and that failure becomes a blocking `exit 2`. But a non-compiling package is precisely the state after step one of the repo's own TDD rule:

> Always write tests first, confirm they fail, then implement.

Reproduced against a temp module — writing a test that calls an as-yet-unimplemented `B()`:

```
G1: go vet ./pkg failed:
# sample/pkg
vet: pkg/b_test.go:6:5: undefined: B
hook exit=2
```

Every red test the agent writes is rejected at the moment it is written, which pressures it into writing the implementation *before* the test — inverting the rule this gate is supposed to support. The script's own header calls this out as the thing to avoid ("design principle 4: a gate that produces false positives loses its credibility and gets bypassed").

Options: only block on genuine vet diagnostics for an otherwise-buildable package (e.g. run `go build` first and skip vet when it fails), or drop `go vet` from `PostToolUse` entirely and leave it to `.githooks/pre-push`, where the red-green cycle has already completed.

---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+	for f in $go_files; do
+		dir=$(dirname "$f")
+		case " $pkgs " in
+		*" ./$dir "*) ;;
+		*) pkgs="$pkgs ./$dir" ;;
+		esac
+	done
+
+	# shellcheck disable=SC2086 # word splitting is the intent: one arg per package
+	test_output=$(go test $pkgs 2>&1)
```
tomo-chan: **Same TDD conflict at the Stop gate: a deliberately-red test makes the turn unendable.**

`DEVELOPMENT.md` requires "write tests first, **confirm they fail**". An agent asked to do exactly that — add the failing test and stop for review — writes `internal/foo/bar_test.go`, the test fails as intended, and this returns `exit 2` with `G2: go test ./internal/foo failed`.

Combined with the missing `stop_hook_active` handling (separate comment on line 19), there is no way out of that state: the turn is resumed and re-blocked.

Worth noting that G2 in `docs/quality-gateway.md` is scoped to `make test-go`, which until now only ran at pre-push — i.e. after the red-green cycle had completed. Moving it to *every turn boundary* changes which states are legal to be in, not just when the check runs.

---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+# the expensive checks belong and where they cannot be skipped anyway.
+#
+# Scope is what the working tree says changed, not the whole repository, so an
+# untouched package is never re-run.
+#
+# Exit codes are the Claude Code hook contract: 0 = the turn may end, 2 =
+# blocking, reason on stderr. Anything that cannot be checked exits 0 (design
+# principle 4).
+
+set -u
```
tomo-chan, line 23: **`stop_hook_active` is never read, so `exit 2` can loop indefinitely.**

The script reads no stdin at all, so the Stop hook payload — including `stop_hook_active`, which Claude Code sets to `true` when it is *already* continuing because of a previous Stop hook — is discarded. The documented purpose of that field is precisely to prevent an unbounded resume loop.

Concrete sequence: the working tree contains one gofmt-dirty file the turn never touched (or a test the user deliberately left red). Stop fires → `exit 2` → Claude is forced to continue → Claude has nothing more to do → Stop fires again → `exit 2` again. Nothing in the script bounds the retries.

Suggested shape: read the JSON once (guarded by `command -v jq`, as `post-edit-check.sh` already does) and `exit 0` immediately when `.stop_hook_active` is `true`.

---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+	fi
+
+	# vitest related resolves the test files that import each changed module,
+	# so an edited hook runs its own tests without running all 704 of them.
+	rel_files=""
+	for f in $frontend_files; do
+		rel_files="$rel_files ${f#frontend/}"
+	done
+	# shellcheck disable=SC2086 # word splitting is the intent: one arg per file
+	vitest_output=$(cd frontend && npx --no-install vitest related --run $rel_files 2>&1)
```
tomo-chan: **`vitest related` exits non-zero when it finds no related tests, blocking healthy turns.**

Vitest defaults `passWithNoTests` to `false`, and `frontend/vite.config.ts`'s `test` block does not set it — a run that resolves zero test files reports "No test files found" and exits 1. That becomes `G2: vitest related failed` here.

Modules with no importing test exist in the tree today: `frontend/src/main.tsx`, and under `frontend/src/components/`, `PaneHeaderIcons.tsx`, `PaneStatusBar.tsx`, `SessionBadge.tsx`, `SplitDivider.tsx`. A brand-new component file created during a turn has no importer *by construction*. Editing any of them refuses to end an entirely healthy turn — design principle 4 again.

```suggestion
	vitest_output=$(cd frontend && npx --no-install vitest related --run --passWithNoTests $rel_files 2>&1)
```

---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+
+set -u
+
+changed=$(git status --porcelain 2> /dev/null | awk '{ print $NF }')
+[ -n "$changed" ] || exit 0
+
+go_files=""
+frontend_files=""
+for f in $changed; do
+	[ -f "$f" ] || continue
```
tomo-chan: **Deleted `.go` files are skipped, so their package is never tested.**

`[ -f "$f" ] || continue` drops deleted paths before the package set is built, so a turn whose only Go change is a deletion runs no tests at all — even though deleting a file is one of the likeliest ways to break a package's build.

Reproduced:

```
$ git rm -q pkg/a_test.go
$ git status --porcelain
D  pkg/a_test.go
$ sh stop-check.sh; echo $?
0          # no test run at all
```

Same holds for deleting a non-test file that other files in the package reference: the package no longer compiles and G2 reports success.

The existence check is doing double duty. `gofmt` genuinely needs the file to exist; the *package* set does not. Deriving `pkgs` from `dirname "$f"` before filtering, and applying `[ -f ]` only when building `go_files` for gofmt, covers both.

---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+
+	# shellcheck disable=SC2086 # word splitting is the intent: one arg per package
+	test_output=$(go test $pkgs 2>&1)
+	if [ $? -ne 0 ]; then
+		note "G2: go test$pkgs failed:
+$test_output"
+	fi
+fi
+
+if [ -n "$frontend_files" ] && [ -d frontend/node_modules ]; then
```
tomo-chan: **This script fails *open* when its cwd is not the repo root.**

`post-edit-check.sh` carefully anchors itself (`repo_root=$(... cd -- "$hooks_dir/../.." && pwd)`); this one does not, and uses bare relative paths here (`frontend/node_modules`, `cd frontend`) plus cwd-relative resolution of the porcelain paths on line 27.

The mismatch matters because `git status --porcelain` always prints paths relative to the **repo root**, regardless of cwd. Reproduced with a gofmt-dirty `sub/pkg/s.go`:

```
$ cd $repo/sub && sh stop-check.sh; echo $?
0          # wanted 2 — every [ -f "$f" ] missed
$ cd $repo   && sh stop-check.sh; echo $?
2
```

Silently checking nothing is worse than erroring, and it is the "reports the discipline as enforced while enforcing nothing" mode the `test-hooks` Makefile comment warns about. Adding the same `repo_root` computation and a `cd "$repo_root"` at the top makes it deterministic — and would also let `hooks_test.sh` exercise it from any directory.

---

.claude/hooks/hooks_test.sh
```diff
@@ -0,0 +1,230 @@
...
+	pass "the Stop hook stays inside G1+G2 (decision D6)"
+fi
+
+# --- settings.json wiring ----------------------------------------------------
+#
+# The scripts above are only a gate if settings.json actually points at them.
+settings="$repo_root/.claude/settings.json"
+
+checks=$((checks + 1))
+if jq -e . "$settings" > /dev/null 2>&1; then
```
tomo-chan: **Unguarded `jq` dependency breaks `make check` with a message that blames the wrong file.**

`jq` is not installed by `make install-deps` (it runs `npm install` + `go mod download` only), and `make test-hooks` now sits inside `make test` → `make check`, which `DEVELOPMENT.md` makes mandatory. Every other external tool in these scripts is guarded with `command -v`; `jq` is not.

On a machine without `jq`, six of the sixteen checks go red:

- this line reports `settings.json is not valid JSON — a malformed file silently disables every setting in it`, when the file is perfectly fine
- both `assert_wired` checks fail
- the three stdin-shape checks (lines 101/104/114) fail, because `post-edit-check.sh` deliberately `exit 0`s when `jq` is missing (its own `command -v jq` guard) while these expect exit 2

So a contributor gets a broken mandatory gate plus a diagnosis pointing at a healthy `settings.json`. Either guard with `command -v jq` and skip those checks, or add `jq` to `install-deps` so the dependency is explicit.

---

.claude/hooks/post-edit-check.sh
```diff
@@ -0,0 +1,95 @@
...
+block() {
+	printf '%s\n' "$1" >&2
+	exit 2
+}
+
+case "$file" in
+*.go)
+	command -v gofmt > /dev/null 2>&1 || exit 0
+
+	unformatted=$(gofmt -s -l "$file" 2>&1)
```
tomo-chan: **`2>&1` folds gofmt's parse errors into the "not formatted" verdict.**

`gofmt -s -l` writes syntax errors to stderr and the list of unformatted files to stdout. Merging them means a Go file that does not *parse* is reported as a formatting problem, with advice that cannot possibly fix it.

Hit this while testing:

```
G1: /path/x.go is not gofmt -s clean. Run: make fmt
```

...for a file whose actual problem was a misplaced build tag. `make fmt` runs gofmt too, so the agent is told to run a command that will fail the same way, and the real diagnostic is swallowed.

```suggestion
	unformatted=$(gofmt -s -l "$file" 2> /dev/null)
```

(or capture stderr into its own variable and include it verbatim in the `block` message — the parse error is the useful part).

---

.claude/agents/diff-reviewer.md
```diff
@@ -0,0 +1,82 @@
...
+
+## What to read, in this order
+
+1. `git diff main...HEAD` — the change under review. Read all of it.
+2. `git log main..HEAD` — what the author says they did.
+3. `AGENTS.md`, then whichever documents it indexes are relevant to the files
+   the diff touches. `DEVELOPMENT.md` always is; `docs/security.md` is whenever
+   the diff touches command execution, shell arguments, SSH paths, host keys,
+   or `gosec` posture.
+4. The surrounding code the diff does not touch, wherever a finding depends on
```
tomo-chan, line 37: **`git diff main...HEAD` silently reviews the wrong thing when the local `main` ref is stale.**

`main` here is the *local* branch, not `origin/main`. In a fresh worktree, a CI checkout, or any branch that has been open for a few merges, it is behind — and the three-dot merge base is then computed against an old commit, so the reviewer reads commits that are not part of the change.

I hit this reviewing this very PR: local `main` was 7 commits behind origin, and

```
$ git diff --stat main...HEAD | tail -1
130 files changed, 13477 insertions(+), 519 deletions(-)
```

for a PR that actually changes **9 files / +573**. An agent following this instruction literally would burn its entire fresh context on already-merged work and report findings against code nobody touched — the exact opposite of what D5 is buying.

Worth pinning the ref explicitly, e.g. `git fetch origin main` first and then `git diff origin/main...HEAD` (and correspondingly `git log origin/main..HEAD` on the next line).

---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+# principle 4).
+
+set -u
+
+changed=$(git status --porcelain 2> /dev/null | awk '{ print $NF }')
+[ -n "$changed" ] || exit 0
+
+go_files=""
+frontend_files=""
+for f in $changed; do
```
tomo-chan: **Second, independent parsing bug on the same pipeline: paths containing spaces.**

`awk '{ print $NF }'` takes the last *whitespace-separated* field, and git additionally double-quotes such paths in porcelain output. Both halves break, and then `for f in $changed` word-splits whatever survives.

```
$ git status --porcelain
?? go.mod
?? "my file.go"
$ git status --porcelain | awk '{ print $NF }'
go.mod
file.go"          # <- not a real path
$ sh stop-check.sh; echo $?
0                 # wanted 2; "my file.go" is gofmt-dirty
```

`[ -f ]` then drops it, so the file passes the gate unchecked — fails open, same as the untracked-directory case above.

Both are fixed together by reading the porcelain output NUL-delimited and taking everything from column 4 on, rather than field-splitting:

```sh
changed=$(git status --porcelain=v1 -z --untracked-files=all 2>/dev/null | tr '\0' '\n' | cut -c4-)
```

(note this also needs the `for f in $changed` loop to iterate over lines rather than words — e.g. via `printf '%s\n' "$changed" | while IFS= read -r f`).

---

.claude/hooks/stop-check.sh
tomo-chan: **The Stop gate silently passes everything when the session cwd is a subdirectory.** (confirmed)

`git status --porcelain` always prints paths relative to the *repository root*, but this script never moves there — unlike `post-edit-check.sh`, which carefully derives `repo_root` from `$0`. So `[ -f "$f" ]` on line 27 fails for every entry and the whole gate no-ops. `[ -d frontend/node_modules ]` (73) and `cd frontend` (74, 87) are cwd-relative for the same reason.

Reproduced: temp repo, tracked `sub/ok.go` modified to be space-indented.

```
--- from root ---   exit=2   (blocks, correct)
--- from sub/  ---  exit=0   (passes, silently)
```

Note the fix can't be `cd "$repo_root"` derived from `$0` — `hooks_test.sh`'s temp-repo fixtures depend on cwd. `cd "$(git rev-parse --show-toplevel)" || exit 0` before this line keeps both working.


---

.claude/hooks/stop-check.sh
tomo-chan: **An entirely new directory is invisible to this gate.** (confirmed)

`git status --porcelain` collapses a wholly-untracked directory into one entry ending in `/`, which then fails `[ -f "$f" ]` on the next line.

Reproduced: repo with a committed baseline, plus a new `newpkg/n.go` containing space-indented Go.

```
$ git status --porcelain
?? newpkg/
$ stop-check.sh ; echo $?
0
```

So a turn that creates a brand-new Go package with unformatted, vet-failing, untested code ends cleanly. That is exactly the "reports the discipline as enforced while enforcing nothing" failure this PR argues against, and it is the common shape for an agent turn that adds a package.

```suggestion
changed=$(git status --porcelain --untracked-files=all 2> /dev/null | awk '{ print $NF }')
```
(applied to line 21)


---

.claude/hooks/hooks_test.sh
tomo-chan: **`jq` is now an undeclared hard dependency of `make check`, and therefore of `git push`.** (confirmed)

`make test` gained `test-hooks`, `make check` runs `make test`, and `.githooks/pre-push` runs `make check`. This file hard-depends on `jq`, which `make install-deps` does not install and which is listed as a prerequisite nowhere in `DEVELOPMENT.md`.

Reproduced by running the suite with `jq` removed from `PATH`:

```
FAIL: settings.json is not valid JSON — a malformed file silently disables every setting in it
FAIL: PostToolUse is not wired to post-edit-check.sh
FAIL: Stop is not wired to stop-check.sh
5 of 16 hook checks failed
```

Two problems. A contributor or CI image without `jq` cannot push at all. And the messages are actively misleading — they accuse `settings.json` of being malformed and the hooks of being unwired when the file is fine and the wiring is fine. That is the false-positive shape design principle 4 warns about, landing in the very suite that guards against it.

Suggest gating the three jq-based checks on `command -v jq` (reporting them as skipped, not passed), and/or documenting `jq` as a prerequisite.


---

.claude/hooks/post-edit-check.sh
```diff
@@ -0,0 +1,95 @@
...
+
+file=""
+if [ "${1:-}" = "--file" ]; then
+	file=${2:-}
+else
+	# Claude Code reports the written path under tool_input.file_path for Edit
+	# and under tool_response.filePath for some other tools. Read both, and
+	# treat unparseable input as "nothing to check" rather than as a failure —
+	# this hook must never be the reason a turn stops.
+	if command -v jq > /dev/null 2>&1; then
```
tomo-chan, line 35: **Without `jq`, the entire G1 edit gate disables itself silently.** (confirmed)

If `jq` is absent, `$file` stays empty and line 40 exits 0 with no output. Every edit passes G1 forever and nothing ever says so.

Reproduced with `jq` off `PATH` — the two stdin-driven fixtures that are supposed to block both return 0:

```
FAIL: hook JSON on stdin reaches the same verdict (unformatted): wanted exit 2, got 0
FAIL: the tool_response.filePath spelling is read too: wanted exit 2, got 0
```

The header argues "a hook that silently stops working reports the discipline as enforced while enforcing nothing" — this is that case. Exiting 0 is right (principle 4), but it should say why:

```sh
	if command -v jq > /dev/null 2>&1; then
		file=$(jq -r '.tool_input.file_path // .tool_response.filePath // empty' 2> /dev/null)
	else
		printf 'G1: jq not found; the edit gate is not running\n' >&2
	fi
```


---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+	vitest_output=$(cd frontend && npx --no-install vitest related --run $rel_files 2>&1)
+	if [ $? -ne 0 ]; then
+		note "G2: vitest related failed:
+$vitest_output"
+	fi
+fi
+
+if [ -n "$problems" ]; then
+	printf 'The turn cannot end with these unresolved (docs/quality-gateway.md gates G1/G2):%s\n' "$problems" >&2
+	exit 2
```
tomo-chan: **`stop_hook_active` is never consulted, so a turn can be blocked forever.**

This script never reads its stdin, so it never sees `stop_hook_active` — the field Claude Code supplies specifically so a Stop hook can tell "first block" from "I already blocked this turn and Claude is continuing because of me".

Combined with the scope on line 21 — the header says "verify what this turn changed", but `git status --porcelain` reports the whole dirty working tree — the concrete loop is ordinary: a pre-existing WIP edit in an unrelated package, or a test that was already failing before the session started, makes this exit 2 on *every* stop attempt, including for a turn that only touched a doc. Claude cannot end the turn and cannot fix a problem it did not create.

Two independent fixes, both worth having: read stdin and `exit 0` when `.stop_hook_active` is true, and narrow the scope to what the turn actually changed.


---

.claude/hooks/stop-check.sh
tomo-chan: **A deleted Go file drops its package out of G2 entirely.**

The `-f` guard is needed for the `gofmt` list, but it runs before the package list is built, so a deletion contributes nothing to `$pkgs`.

Scenario: a turn removes `internal/foo/bar.go` and changes nothing else in that package. `go_files` is empty, lines 52–71 never run, and the Stop gate reports success even though `internal/foo` no longer compiles and its tests fail. Deleting a file is a normal refactoring step and is exactly the kind of change whose blast radius the package's own tests exist to catch.

The dirname of a deleted `.go` path should still reach `$pkgs` on lines 57–63, even though the file itself must stay out of the `gofmt` argument list.


---

.claude/hooks/stop-check.sh
tomo-chan: **`vitest related` without `--passWithNoTests` blocks the turn when the edited module has no related test.** (suspected — not reproduced, `frontend/node_modules` is not installed in my checkout)

Vitest exits non-zero when no test files match, and `frontend/vite.config.ts` does not set `passWithNoTests`. `related` resolves tests by module graph, so any module no test imports produces zero matches.

`frontend/src/main.tsx` is the concrete case in this repository — it matches the `frontend/src/*.ts|*.tsx` arm on line 30 (in a `case` pattern `*` matches `/`, so nested paths match too), and nothing imports it; `App.test.tsx` tests `App`, not the entry point. Editing it would produce `G2: vitest related failed` on a change that is entirely healthy — the design-principle-4 false positive this PR sets out to avoid.

```suggestion
	vitest_output=$(cd frontend && npx --no-install vitest related --run --passWithNoTests $rel_files 2>&1)
```


---

.claude/hooks/hooks_test.sh
```diff
@@ -0,0 +1,230 @@
...
+response_shape_status() {
+	"$edit_hook" < "$work/response-shape.json"
+}
+expect_status 2 "the tool_response.filePath spelling is read too" response_shape_status
+
+# Malformed input is not a reason to block a turn.
+malformed_status() {
+	printf 'not json at all' | "$edit_hook"
+}
+expect_status 0 "malformed hook input does not block" malformed_status
```
tomo-chan, line 125: **Neither the `go vet` branch nor the `tsc` branch of the edit gate is covered by any of the 16 checks.**

Every G1 fixture lives under `mktemp -d`, and `post-edit-check.sh` deliberately skips `go vet` for files outside `$repo_root` and `tsc` for files outside `$repo_root/frontend`. So all 16 checks exercise only the `gofmt` path plus the stdin plumbing.

Both branches do work today — verified by hand, a file under `internal/` containing `fmt.Printf("%d\n", "not an int")` makes the hook exit 2 with the vet output. But that is precisely the point: a regression in the `rel` computation (post-edit-check.sh lines 71–75), in the `cd "$repo_root"` invocation, or anywhere in the tsc arm would leave all 16 checks green while the gate checked nothing — the failure mode this file's own header calls "worse than no hook at all".

One fixture per branch placed inside the repository tree (created and removed by the test) would close it.


---

Makefile
```diff
@@ -44,6 +44,15 @@ test-frontend:
 test-e2e:
 	cd frontend && npm run test:e2e
 
+# The agent-side gates themselves (docs/quality-gateway.md's G1 and G2, run as
+# Claude Code hooks from .claude/). Included in `make test` because a hook that
+# silently stops working reports the discipline as enforced while enforcing
+# nothing — and because a hook that blocks a healthy change is worse still, so
+# both directions are asserted. Hermetic: it drives the scripts against temp
+# files and throwaway git repositories, never this checkout.
+test-hooks:
```
tomo-chan, line 53: **`test-hooks` never runs in CI.**

`.github/workflows/ci.yml` does not invoke `make test`. It runs the Go and frontend suites directly:

```yaml
- name: Test (Go)
  run: go test -v ./... -race 2>&1 | go-junit-report ...
- name: Test (Frontend)
  run: cd frontend && npm run test:ci
```

So folding `test-hooks` into `make test` gets it as far as `.githooks/pre-push` (L3 on this repo's own ladder) and no further — never L4. A change that breaks `post-edit-check.sh` or `stop-check.sh` merges green for anyone who pushes with `--no-verify` or from an environment without the hooks path configured.

`.claude/**` is also absent from `ci.yml`'s `paths` filter, so a PR touching only the hooks does not trigger the `ci` workflow at all. Both halves need fixing together: a `make test-hooks` step plus `.claude/**` in the filter.


---

.claude/agents/diff-reviewer.md
```diff
@@ -0,0 +1,82 @@
...
+`docs/quality-gateway.md`. The session that wrote this code carries the context
+of every approach it tried and discarded, which makes it the worst possible
+judge of whether the approach it kept is sound. You have none of that context,
+and you must not go looking for it: do not read the conversation that produced
+the change, and do not accept an explanation of intent that is not written down
+in the repository or in the diff itself.
+
+## What to read, in this order
+
+1. `git diff main...HEAD` — the change under review. Read all of it.
```
tomo-chan: **`main` here is the *local* ref, which in an agent worktree is routinely stale — the reviewer then reviews the wrong scope.** (confirmed, in this session)

Reviewing this very PR, the local `main` sat 3 commits behind `origin/main`:

```
$ git diff main...<pr head> --stat | tail -1
130 files changed, 13477 insertions(+), 519 deletions(-)

$ git diff origin/main...<pr head> --stat | tail -1
9 files changed, 573 insertions(+), 8 deletions(-)
```

A reviewer following step 1 literally would have spent its whole budget on already-merged code and could report findings against commits that are on `main`. This matters more than usual here because `DEVELOPMENT.md` prescribes a fresh `git worktree add` per feature, and this agent is told to read the diff and nothing else — it has no way to notice the scope is wrong.

Same applies to `git log main..HEAD` on line 21. Suggest `git fetch origin main` followed by `git diff origin/main...HEAD`, or `@{upstream}` where one is set.


---

.claude/hooks/stop-check.sh
```diff
@@ -0,0 +1,99 @@
...
+# untouched package is never re-run.
+#
+# Exit codes are the Claude Code hook contract: 0 = the turn may end, 2 =
+# blocking, reason on stderr. Anything that cannot be checked exits 0 (design
+# principle 4).
+
+set -u
+
+changed=$(git status --porcelain 2> /dev/null | awk '{ print $NF }')
+[ -n "$changed" ] || exit 0
```
tomo-chan, line 71: Minor, same silent-skip family as the two above: `awk '{ print $NF }'` on line 21 takes the last whitespace-separated field, but git quotes paths containing spaces. Confirmed:

```
$ git status --porcelain
 M "old name.go"
$ git status --porcelain | awk '{print $NF}'
name.go"
```

`[ -f ]` then fails and the unformatted file is skipped, exit 0. No such path exists in this repository today, so this is low priority — noting it only because the failure is silent rather than loud, and `git status --porcelain -z` read NUL-delimited (or stripping the fixed 3-character status prefix) avoids it along with the rename form.


---

tomo-chan (commented):
Code review of the 9-file diff (`origin/main...6c39cb2`). 11 findings, most of them verified by running the scripts rather than reading them.

The two that matter most are both in `stop-check.sh`, and both are the failure this PR's own header calls out — the gate reporting success while checking nothing:

- it never moves to the repo root, so from any subdirectory `git status --porcelain`'s root-relative paths all fail `[ -f ]` and everything passes (root: exit 2, `sub/`: exit 0, same tree);
- a wholly-new directory collapses to `?? newpkg/`, which also fails `[ -f ]`, so a turn that adds a new Go package is never checked at all.

Third: `make test` -> `make check` -> `pre-push` now hard-depends on `jq`, which nothing installs or documents. Without it 5 of 16 checks fail, and they fail with messages blaming `settings.json` and the hook wiring rather than the missing tool.

Everything else is smaller: `stop_hook_active` unread (loop risk), deleted `.go` files dropping their package out of G2, `vitest related` missing `--passWithNoTests`, `test-hooks` never running in CI, the `go vet`/`tsc` branches being uncovered by the 16 checks, and the reviewer subagent's `git diff main...HEAD` reading a stale local `main` — which cost me 130 files instead of 9 when I started this review.

The `hooks_test.sh` suite passes as shipped here (16/16, 6s), and the design calls — D6, the non-blocking reviewer, asserting the absence of `make check` — read as sound.


---

tomo-chan:
All eleven findings addressed in `7687a8b`. Every one reproduced first; the suite goes from 16 checks to **26**.

## The four fail-open paths

These are the ones that matter, and they share a shape worth naming: each returned **0 having examined no files at all**. That is strictly worse than having no gate, because it reports the discipline as enforced — the exact failure this PR's own header claims to guard against, shipped inside the guard.

| Path | Before | After |
|---|---|---|
| cwd in a subdirectory | root `exit 2`, `sub/` `exit 0`, same tree | `cd "$(git rev-parse --show-toplevel)"` |
| wholly-new directory | `?? newpkg/` fails `[ -f ]` → unchecked | `--untracked-files=all` |
| deleted `.go` file | package dropped from the test set | package derived *before* the existence filter |
| path containing a space | `awk '{print $NF}'` mangles it → skipped | `-z` porcelain + `cut -c4-` |

## A bug I introduced fixing another one

Honouring `stop_hook_active` means reading stdin — and a plain `cat` on an inherited idle pipe **blocks forever**. A Stop hook that hangs is worse than one that answers wrongly: the turn cannot end at all, and there is no recovery. It hung this repository's own test suite, which is how I found it. The read is now bounded by `timeout 5`, and there is a test asserting the hook survives an idle pipe.

## `go vet` vs the TDD rule

You were right that this inverted the rule the gate exists to support. The resolution keeps `go vet` but distinguishes the two outputs, verified both ways:

```
vet: sub/b_test.go:6:5: undefined: B            <- mid-TDD, does not block
sub/bad.go:5:26: fmt.Printf format %d has ...   <- genuine diagnostic, blocks
```

A load failure is reported on a line starting `vet: `; a real diagnostic is a bare `file:line:col:`. Both are now fixtures **inside** the repository, since the vet and tsc branches skip files outside it — which is why all 16 previous checks, every one under `mktemp -d`, exercised only the gofmt path.

At the Stop gate the same conflict is bounded by `stop_hook_active` rather than removed: the first block still blocks, the second does not. Your point that moving G2 to every turn boundary changes *which states are legal*, not just when the check runs, is correct and I have not tried to argue otherwise — the loop bound is what makes it survivable.

## `jq`

No longer a hard dependency of `make check`, and therefore of `git push`. Its checks report themselves **skipped** — never passed, which would be the other failure mode. Verified by moving `/usr/bin/jq` aside: 18 of 26 run, the rest skip, suite passes. Previously 5 of 16 failed while blaming a healthy `settings.json`.

## The rest

- `make test-hooks` is now its own CI step and `.claude/**` triggers the workflow — you were right that `ci.yml` never calls `make test`, so folding it in reached L3 and stopped.
- `vitest related --passWithNoTests`.
- `gofmt` stderr kept separate, so a file that does not *parse* is reported as a parse error rather than a formatting problem with advice that fails identically.
- `diff-reviewer.md` fetches and diffs `origin/main`. Your 130-files-vs-9 measurement is quoted in the file, because the number is the argument.

## One test was passing for the wrong reason

Worth flagging on its own. With the cwd fix reverted, the subdirectory case **still returned 2** — because `go test ./sub` was handed a path that does not exist from that cwd. Right answer, wrong cause, and it would have let the fix regress silently. It now asserts the blocking *reason*, and fails with `blocked, but not for the expected reason` when perturbed.

Perturbation results, all confirmed:

| Reverted | Result |
|---|---|
| `--untracked-files=all` | red — "a brand-new package directory is checked" |
| `cd` to repo root | red — "blocked, but not for the expected reason" |
| deleted-file package fix | red — "deleting a .go file still tests its package" |

---

.claude/hooks/stop-check.sh
tomo-chan: Confirmed and fixed in `7687a8b` — reproduced exactly as you describe (root `exit 2`, `sub/` `exit 0`, same tree). Now `cd "$(git rev-parse --show-toplevel)"`, which as you note keeps `hooks_test.sh`'s temp-repo fixtures working where a `$0`-derived root would not.

The follow-on is worth recording, because it says something about the test rather than the script: with the fix reverted, the new subdirectory test **still returned 2** — `go test ./sub` was handed a path that does not exist from that cwd. Right answer, wrong cause, and it would have let the fix regress with the suite green. The test now asserts the blocking *reason* and fails with `blocked, but not for the expected reason`.

---

.claude/hooks/stop-check.sh
tomo-chan: Confirmed and fixed in `7687a8b` with `--untracked-files=all`. Reproduced first: `?? newpkg/`, `exit 0`.

You are right that this is the common shape for an agent turn — adding a package is the single largest kind of change, and it was the one the gate could not see at all. There is now a test for it (`a brand-new package directory is checked`), and reverting the flag turns it red.

---

.claude/hooks/stop-check.sh
tomo-chan: Confirmed and fixed in `7687a8b`, exactly as you prescribe: the package is derived from `dirname` **before** the existence filter, and `[ -f ]` now applies only to the gofmt argument list.

Your framing — "the existence check is doing double duty" — is the whole bug in one sentence. Test added (`deleting a .go file still tests its package`), using a two-file package where removing one breaks the other's build, so it fails for the reason it names.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b`, with a caveat worth reporting rather than burying.

Honouring `stop_hook_active` means reading stdin, and a plain `cat` on an **inherited idle pipe blocks forever**. A Stop hook that hangs is worse than one that answers wrongly — the turn cannot end at all and there is no recovery path. It hung this repository's own test suite, which is how I found it, and it would have hung `make test-hooks` for everyone.

The read is now bounded (`timeout 5 cat`, falling back to plain `cat` where `timeout` is absent), and there is a test that pipes an idle `sleep 30` into the hook and fails if it does not return.

Two tests cover the field itself: `stop_hook_active bounds the block instead of looping` and `the first block still blocks`.

---

.claude/hooks/post-edit-check.sh
tomo-chan: You are right that this inverted the rule the gate exists to support, and that is the worst kind of false positive here — not noise, but pressure toward writing the implementation first.

Fixed in `7687a8b` by taking your first option, but with a sharper discriminator than "run `go build` first". `go build` ignores `_test.go` files, so it succeeds for exactly the case that matters. What separates the two is the *shape of vet's own output*, verified both ways:

```
vet: sub/b_test.go:6:5: undefined: B            <- load failure, mid-TDD, does not block
sub/bad.go:5:26: fmt.Printf format %d has ...   <- genuine diagnostic, blocks
```

A load/type failure is reported on a line starting `vet: `; a real finding is a bare `file:line:col: message`. The hook now says `./pkg does not compile yet, so go vet was skipped` and exits 0 in the first case.

Both are fixtures now — `a real go vet diagnostic blocks the edit` and `a package that does not compile yet does not block (TDD red)` — and they live **inside** the repository, because as you noted in a separate thread the vet branch skips files outside it, so no `mktemp -d` fixture could ever have reached this code.

---

.claude/hooks/stop-check.sh
tomo-chan: Bounded rather than removed, in `7687a8b`: with `stop_hook_active` honoured the first block still blocks and the second does not, so the state is reachable and escapable instead of being a trap.

Your closing observation is the one I want to acknowledge directly rather than argue with:

> Moving it to *every turn boundary* changes which states are legal to be in, not just when the check runs.

That is correct, and it is a real cost the design document does not discuss — G2 is specified as `make test-go` without saying anything about *when* a red suite is a legitimate state to stop in. The loop bound makes it survivable; it does not make the question go away. I have not tried to settle it here, because it is a change to what the gate means rather than a bug in this script, and it deserves a decision in `quality-gateway.md` rather than a quiet choice in a shell script.

---

.claude/hooks/hooks_test.sh
tomo-chan: Fixed in `7687a8b`. The jq-dependent checks now report themselves **skipped** — never passed, which would be the other failure mode you would rightly have flagged next.

Verified the way you did, by making jq genuinely absent (moving `/usr/bin/jq` aside rather than trimming `PATH`, since `/usr/bin` cannot be removed without taking `sh` and `git` with it):

```
jq absent
skip jq not installed — the settings.json and stdin checks below are skipped
all 18 hook checks passed
```

18 of 26 run, the rest skip, exit 0. `jq` is documented in `DEVELOPMENT.md` and `AGENTS.md` as optional, with what is lost when it is missing.

I went with guarding rather than adding jq to `install-deps`: that target installs language toolchain dependencies, and reaching for a system package manager there is a bigger commitment than this feature justifies.

---

.claude/hooks/post-edit-check.sh
tomo-chan: Applied verbatim in `7687a8b` — the hook now prints `G1: jq not found; the edit gate is not running` on stderr and exits 0.

Exiting 0 was right and staying silent was not, which is the distinction your suggestion draws and the one the header should have implied to me when I wrote it.

---

.claude/hooks/post-edit-check.sh
tomo-chan: Fixed in `7687a8b`, taking the second of your two options: stderr is captured separately and reported verbatim, so a file that does not parse blocks with `G1: <file> does not parse:` followed by the actual diagnostic.

The parse error being the useful part is exactly it — the old message sent the agent to `make fmt`, which runs gofmt and fails identically, so the loop had no exit. Same fix applied in `stop-check.sh`, which had the same `2>&1`.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b`, taking your combined fix: `git status --porcelain=v1 -z --untracked-files=all | tr '\0' '\n' | cut -c4-`, with the loop reading lines rather than word-splitting.

You were right that the two bugs share one pipeline and fall to one change. Test added (`a path containing a space is not silently skipped`) — no such path exists in the repository today, as you say, but the test does not depend on one existing.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b` — `--passWithNoTests` added.

Your list of modules with no importing test is the part that made this concrete rather than theoretical: `main.tsx`, `PaneHeaderIcons.tsx`, `PaneStatusBar.tsx`, `SessionBadge.tsx`, `SplitDivider.tsx` are all in the tree today, and as you say a newly created component has no importer by construction — so the gate would have fired hardest on exactly the turns that create new UI.

---

.claude/agents/diff-reviewer.md
tomo-chan: Fixed in `7687a8b`: `git fetch origin main`, then `git diff origin/main...HEAD` and `git log origin/main..HEAD`.

Your 130-files-vs-9 measurement is quoted in the file itself, because the number is the argument — an instruction saying "use origin/main" is easy to soften later, one carrying the cost of not doing so is not. The file also now names why this agent in particular cannot self-correct: it is given the diff and told to read nothing else, so a wrong scope is invisible to it.

That it happened while reviewing *this* PR, in the worktree layout `DEVELOPMENT.md` itself prescribes, is what makes it a real defect rather than a hypothetical.

---

.claude/hooks/hooks_test.sh
tomo-chan: Fixed in `7687a8b`, exactly as suggested: one fixture per branch, inside the repository, created and removed by the test (with the cleanup on a `trap` so a failure cannot leave them behind).

Three checks now, not two — the vet arm has both outcomes:

- `a real go vet diagnostic blocks the edit`
- `a package that does not compile yet does not block (TDD red)`
- `a TypeScript type error blocks the edit`

Your point is the one I would have missed on my own: both branches *worked*, and that was precisely the problem. Sixteen green checks over a gate where two of three arms were unreachable is the same "reports the discipline as enforced" failure, one level up — the tests had it about the hooks.

---

Makefile
tomo-chan: Both halves fixed together in `7687a8b`, as you said they needed to be: a `Test (agent hooks)` step running `make test-hooks`, and `.claude/**` added to both `paths` filters.

You are right that this was the difference between L3 and L4 on the ladder this PR's own documentation sets out — and the gap was invisible from the Makefile, because `make test` looks like it reaches CI until you check what `ci.yml` actually calls.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b` — and I took your parenthetical, which was the part I would have got wrong: the root comes from `git rev-parse --show-toplevel`, not from `$0`, precisely so the temp-repo fixtures keep working. See the main thread on this file for the follow-on about the test that was passing for the wrong reason.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b` with `--untracked-files=all`, plus the `-z`/`cut -c4-` change from the spaces thread, since both live on the same pipeline.

---

.claude/hooks/hooks_test.sh
tomo-chan: Fixed in `7687a8b`. Verified with jq genuinely absent: 18 of 26 checks run, the rest report **skipped**, suite exits 0. Documented in `DEVELOPMENT.md` and `AGENTS.md` as optional, with what is lost without it.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b`. Your second half — that the scope is the whole dirty tree, not what the turn changed — I have **not** fixed, deliberately: `git status` cannot distinguish them, and narrowing would need per-turn state this hook does not have. The loop bound is what makes a pre-existing problem survivable rather than a trap. Worth its own decision in `quality-gateway.md` if the scope should genuinely change.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b` — the `dirname` of a deleted `.go` path now reaches `$pkgs` while the file itself stays out of the gofmt list, as you describe.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b`. Marked "suspected" on your side — confirmed on mine, with `node_modules` present: `--passWithNoTests` is required and is now passed.

---

.claude/agents/diff-reviewer.md
tomo-chan: Fixed in `7687a8b` — `git fetch origin main` then `origin/main...HEAD`, and the same for `git log`. Your measurement is quoted in the file.

---

.claude/hooks/stop-check.sh
tomo-chan: Fixed in `7687a8b` with the `-z` / `cut -c4-` form you suggest, which as you note also handles the rename shape. Low priority was the right call — but "silent rather than loud" is why it was worth doing anyway, and it came free with the untracked-directory fix on the same pipeline.

---

PR #185, opened 2026-08-23
ci: add red-check (make efficacy) as a pull-request gate (#180 item 4)

tomo-chan:
Roadmap item 4 of #180 — gate **G4(b)** in [docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md).

- [x] 変更されたテストを特定し、実装差分を戻した状態で走らせて**落ちること**を検証するターゲットを作る
- [x] PR 専用の CI ジョブとして追加する（`make check` には入れない）
- [x] 実装差分が無い PR（ドキュメントのみ等）では skip する
- [x] 意図的に落ちないケースの免除方法を決める

## The rule

DEVELOPMENT.md requires tests to be written first and confirmed failing. That rule is stuck at rung **L0** because nothing in a merged branch records the order its lines were written in. The *result* can be checked, and it is the part that matters — **decision D4**:

> A test this branch changed must **fail** when this branch's implementation diff is reverted.

That is the strongest possible mutation — remove the implementation entirely — and it catches the two shapes the existing gates are blind to: a **tautological test**, and a test **written after the fact** to describe code that already worked.

#181 ran exactly this by hand, four times, and each perturbation went red as intended. This automates that.

## How it works

`scripts/efficacy.sh` builds a scratch worktree at HEAD and checks **each changed test on its own, in two phases**:

1. **at HEAD, nothing reverted** — it must pass;
2. **with the implementation diff reverted** — it must fail.

Anything that survives phase 2 is named individually. The real checkout is never touched — a half-reverted working tree would be a nasty thing to leave behind on a failure — and a test asserts that, including the worktree list.

Both phases are load-bearing, and so is "each on its own". Judging a whole set with one command and asking only "did it go red" is a materially weaker rule, because one honest test then covers for every tautology beside it. And without phase 1, a test that was never going to run, a package that was never going to build in the worktree, or a test that was already failing all report as "red" for reasons that have nothing to do with the revert.

## Five details that are decisions, not implementation

Recorded because each one was a way the gate could have reported green on exactly what it was built to catch — and most of them *were*, until review reproduced them.

**1. The verdict is per test, and it takes two runs.** On the Go side it comes from the test's own `--- PASS:`/`--- FAIL:`/`--- SKIP:` line rather than exit status, because `go test` also exits 0 when `-run` selects nothing. On the frontend side it comes from vitest's **JSON reporter**, read per case: `-t` matches as an *unanchored* regex over the full `describe`-joined name (so `-t 'renders'` also selects `renders empty state`, and `renders$` still matches `always renders`), and the `Tests …` summary is an aggregate. Benchmarks are excluded at collection — `-run` never selects them, and a benchmark asserts nothing that could go red.

**2. Scope is the changed test *functions* and *cases*, and there is no whole-file fallback.** The diff's touched line numbers are mapped onto function ranges, so editing an assertion brings that test into scope and appending a new test does *not* drag the untouched one above it in. The frontend does the same with `it(...)`/`test(...)` blocks — and where a name is not a literal (`it.each`), the case is resolved *after* the run from the source lines vitest emits under `--includeTaskLocation`, rather than by running the file whole. Running whole and reading the aggregate would be the last place a sibling's red could vouch for the changed case; "over-wide" is not the safe direction for a gate that reads red as success.

**3. Skipping is as important as failing** (principle 4):

| Branch shape | Verdict |
|---|---|
| No implementation changed (docs, or tests only) | skip — nothing to revert |
| Implementation changed, no test changed | **warn**, exit 0 — no result to check, and failing here would fire on every pure refactor |
| Implementation on one stack, tests only on the other | that stack is not judged — nothing was reverted under it |
| Some changed tests skipped or unresolvable, others checkable | those are reported; the rest are judged |
| **Every** changed test skipped or unrunnable | **fail** — see (4) |
| Implementation + test changed, every changed test goes red | pass |
| Any changed test still green | **fail**, naming it |

**4. "Could not check" is a failure, never a skip.** A missing base ref, a scratch worktree that could not be created, a missing `frontend/node_modules`, or every changed test turning out to be unrunnable — each exits 1 with a message saying what to fix. A green required check that checked nothing is invisible, which is the one failure mode being required is supposed to rule out.

**5. Two escape hatches, at two scales.** `//efficacy:exempt` in the comment above one test takes that test out of scope — for a test in a package whose implementation this branch never touched, which was never going to go red. The PR-wide `efficacy-exempt` label (locally `EFFICACY_EXEMPT=1`) covers a branch whose tests genuinely should not go red at all, a pure refactor. Both are visible where the reviewer already looks; the per-test one exists because using the label to get past one unrelated test also exempts a genuine tautology elsewhere in the same branch.

## Not in `make check`

`make efficacy` needs the base branch, a scratch worktree and a second test run — a per-pull-request cost, not a per-turn one. Same reasoning that keeps mutation testing (D2) and the agmsg contract out of the local gate. It stays cheap when it does run: measured on the branch for rollout item 2, 34 changed tests over both phases took ~30s with a warm Go build cache.

`efficacy.yml` runs it on pull requests with `fetch-depth: 0` (a shallow clone has no merge-base to revert to), with **no `paths:` filter** (a path-filtered workflow reports no status at all on the pull requests it skips, and a required check that never reports blocks every one of them), and with `types: [opened, synchronize, reopened, labeled, unlabeled]` so the `efficacy-exempt` label can actually be applied to a failing run.

## Its own tests

`scripts/efficacy_test.sh` (`make test-efficacy`, inside `make test` since it is hermetic) drives the script against **real throwaway git repositories with real `go test` runs** — 29 checks. Three of them additionally drive real vitest against the repository's own `node_modules`, and report themselves skipped if they are absent, because the bugs they guard are invisible to `--changed-tests`.

The tautology cases cannot be faked: what *makes* a tautology is that the compiler and the test runner are both perfectly happy with it, so the only way to test the gate is to build one and watch it survive.

## Review rounds

Four rounds, eighteen findings, every one reproduced before being fixed and re-verified after. The pattern worth naming: **rounds 2 and 4 each found the previous round's own fix reintroducing the shape it had just closed, one level down** — and in both cases for the same reason, so it is written down rather than left as bad luck: the fixture written alongside a fix was in the one shape where the bug it guards cannot appear.

**Round 1 (f628223, 8a4b380) — eight findings.** The verdict was one `go test` exit status over the union of all changed tests, so a genuinely-red test masked a tautology in the same branch. The scratch worktree could never build the root package (`main.go` embeds the gitignored `frontend/dist`), so every root-package test "went red" with zero files reverted. Neither stack's run was conditioned on that stack's implementation having changed. Plus: `sed -i` is GNU-only and this suite runs inside `make check`; `[no tests to run]` was reported as a survivor; frontend scope was whole files; a missing `node_modules` dropped the frontend half in silence; two paths failed open. A follow-up fixed the `efficacy-exempt` label being unappliable.

**Round 2 (5de6abc) — four findings.** Round 1 gave the frontend *scope* but not *verdict*: `-t` is unanchored and the summary is an aggregate, so a tautology was reported red because its sibling went red — and D4's text had started claiming the property. Fixed with the JSON reporter, per case. `run == 0` exited 0, reachable through `internal/board`'s six environment-`t.Skip`s. A Go test in a package whose implementation the branch never touched is still failed — a real false positive that per-package conditioning would *not* fix (`internal/server`'s tests legitimately cover `internal/api`'s), so `//efficacy:exempt` was added instead.

**Round 3 (7a6c318) — four items.** The marker added in round 2 **inverted itself** on the frontend: when every touched case in a file carried it, `changed_fe_test_names` printed nothing, which the caller read as "could not narrow" and fell back to the whole file — so applying the marker widened scope and ran the exempted case anyway. It now emits `__ALL_EXEMPT__`, distinct from `__UNMAPPED__`. The gap that let it through was that the marker's only fixture was Go-side; there are now three, one per shape. The residual whole-file aggregate is gone rather than documented, via `--includeTaskLocation` (see detail 2). Plus a dead `label`, and both SURVIVOR messages now naming the marker, since that message — not the `run == 0` error — is what someone hitting the cross-package false positive actually reads.

**Round 4 (dd4de2d) — two findings.** Round 3's location resolution had the same off-by-one-block bug the static mapper documents at length. vitest reports a case at the line of its `it(...)`/`])(...)` **call**, so for a multi-line `it.each([` the block's own opening lines fell inside the case above it, and touching the block reported `SURVIVOR: … > adds` for a test the branch never touched — the gate manufacturing the exact false positive `//efficacy:exempt` exists for, against a test whose author has no reason to mark it. Block starts now come from `fe_block_starts`, running the same scanner and doc-comment walk as the static path, so the two cannot disagree; the reported line only assigns a title to a block. Round 3's fixture missed it by writing its `it.each` on one line — the single shape where the reported line *is* the block's first line, while all five real occurrences in this repository are multi-line. Second finding: `--changed-tests` still printed `(whole file)` for a file whose real fate is post-run resolution, and the fixture pinning that string was still named after the removed fallback — passing while describing behaviour that no longer exists. It now reads `(resolved after the run)`, and a companion check runs the shape for real: a setup-only frontend change ends in "no case could be resolved" and exit 1, a verdict flip from the old whole-file pass.

## Test plan

- [x] `make test-efficacy` — 29/29
- [x] Every check confirmed to fail against the version of the script it was written for — 8 perturbations in round 1, 3 in round 2, 2 in round 3, 2 in round 4
- [x] The frontend per-case fix and the location resolution both verified end to end against real vitest, reproducing the reviewer's own fixtures
- [x] `make efficacy` run end to end against a branch with real Go changes (rollout item 2's, 34 changed root-package tests): phase 1 green for all of them, ~30s total
- [x] `make efficacy` run on this branch — correctly reports "no implementation changed against origin/main", since this PR is scripts, CI and docs
- [x] `make lint` — 0 issues
- [x] `make check` — passes end to end (now including `test-efficacy`)

> Note: this PR's own CI run of the new job exercises the skip path, not the gate — confirmed on the branch head: `efficacy: no implementation changed against origin/main — nothing to revert. Skipping.` The gate's failing path is covered by `make test-efficacy` above.


scripts/efficacy.sh
```diff
@@ -0,0 +1,256 @@
...
+status=0
+
+if [ -n "$names" ]; then
+	pattern=$(printf '%s' "$names" | sed 's/^ //; s/ /|/g')
+	echo "efficacy: running the changed Go tests with the implementation reverted"
+	echo "  packages: $pkgs"
+	echo "  tests:    $pattern"
+
+	# shellcheck disable=SC2086 # word splitting is the intent: one arg per package
+	if (cd "$worktree" && go test $pkgs -run "^($pattern)$" -count=1 > "$tmp/go.log" 2>&1); then
```
tomo-chan: **The scratch worktree can never build the root `panemux` package, so the gate is vacuous for every root-package test.**

`main.go:32` has `//go:embed frontend/dist`, and `.gitignore:4` excludes `frontend/dist/`. A `git worktree add` checkout therefore has no `frontend/dist` at all. Reproduced against this branch:

```
$ wt=$(mktemp -d); rm -rf "$wt"; git worktree add --detach "$wt" HEAD
$ cd "$wt" && go test . -run '^(TestNothingZZZ)$' -count=1
# panemux
main.go:32:12: pattern frontend/dist: no matching files found
FAIL	panemux [setup failed]
```

That failure has nothing to do with the revert — it happens with *zero* files reverted. So for any PR touching `board_test.go`, `bootstrap_test.go`, `command_center_test.go`, `main_test.go` or `url_open_main_test.go` (i.e. most Agent Board / bootstrap work in this repo), `./.` is always red and the gate always passes, tautology or not.

It compounds with the union-`go test` issue below: because `./.` is in the same invocation as every other package, its unconditional build failure also masks a tautological test in `internal/...` in the same run.

`efficacy.yml` never runs `make build-frontend`, so CI hits this on every such PR. Either create a placeholder in the worktree before running (`mkdir -p "$worktree/frontend/dist" && touch "$worktree/frontend/dist/.keep"`), or build the frontend there, or run packages separately and treat a build failure unrelated to the reverted files as an error rather than a pass.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,256 @@
...
+		fi
+	done
+}
+
+revert "$go_impl"
+revert "$fe_impl"
+
+status=0
+
+if [ -n "$names" ]; then
```
tomo-chan: **One `go test` invocation over the union of all changed tests: a genuinely-red test masks a tautological one.**

The gate's verdict is the exit status of a single command (line 223) covering *every* changed test function in *every* changed package. `go test` exits nonzero if **any** one of them fails — so "the set went red" does not mean each member went red.

The PR description says the scope-narrowing fix removed this ("an over-wide set lets an unrelated failure **mask** a tautological test"). Narrowing to changed *functions* only removed the intra-file-neighbour instance; the cross-test / cross-package instance is still there, and it is the common one — essentially every PR in this repo changes more than one test.

Reproduced with a two-package fixture: package `a` gets `Mul` + a real `TestMul`; package `b` gets `Bye` + a `TestBye` that never mentions `Bye` (a pure tautology):

```
efficacy: running the changed Go tests with the implementation reverted
  packages:  ./a ./b
  tests:    TestMul|TestBye
  ok — they went red, as a test written before its implementation must.
EXIT=0
```

`TestBye` is exactly the shape D4 exists to catch, and the gate reports green. Running each changed test function individually (or at least per-package, requiring *every* member to be red) is what the stated rule actually asks for.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,256 @@
...
+		echo "      or it asserts nothing at all. See decision D4 in docs/quality-gateway.md."
+		echo
+		sed 's/^/      /' "$tmp/go.log"
+		status=1
+	else
+		echo "  ok — they went red, as a test written before its implementation must."
+	fi
+fi
+
+if [ -n "$fe_tests" ] && [ -d "$repo_root/frontend/node_modules" ]; then
```
tomo-chan: **False failure: each stack's tests run even when that stack's implementation wasn't reverted.**

Line 216 gates the Go run on `[ -n "$names" ]` only, and this line gates the frontend run on `[ -n "$fe_tests" ]` only. Neither checks that the *matching* implementation list is non-empty — the skip at line 177 is satisfied if *either* `$go_impl` or `$fe_impl` is set.

So a cross-stack branch gets failed for a revert that never happened. Reproduced: a branch that changes `frontend/src/x.ts` and appends `TestAddMore` to a Go test file, with no `.go` implementation changed at all:

```
efficacy: running the changed Go tests with the implementation reverted
  packages:  ./.
  tests:    TestAddMore

FAIL: these tests still passed with this branch's implementation reverted.
      ok  	sample	0.003s
EXIT=1
```

Nothing Go-side was reverted, so of course it passed — and the message asserts the opposite of what happened. The mirror case bites more often here: a PR that changes Go implementation and also touches any `frontend/src/**/*.test.ts` without changing a `frontend/src` implementation file (e.g. fixing a flaky frontend test alongside a backend change) fails the frontend half the same way. Given this job is meant to be a required check, that blocks the PR.

Suggest `[ -n "$names" ] && [ -n "$go_impl" ]` and `[ -n "$fe_tests" ] && [ -n "$fe_impl" ]`, with the existing warn path covering "impl changed on one stack, tests only on the other".

---

scripts/efficacy_test.sh
```diff
@@ -0,0 +1,319 @@
...
+#
+# The mapping is by touched LINE, not by "+func Test": strengthening an
+# assertion inside an existing test must bring that test into scope, or the
+# gate would be trivially avoided by never adding a new test function.
+edited=$(new_fixture)
+(
+	cd "$edited" || exit 1
+	# Add a bug-fix to Add, and tighten the existing test to notice it.
+	sed -i 's/return a + b/return a + b + 0/' sample.go
+	sed -i 's/if Add(1, 2) != 3 {/if Add(1, 2) != 3 || Add(0, 0) != 0 {/' sample_test.go
```
tomo-chan: **`sed -i` with no backup suffix is a GNU extension — this breaks `make check` (and the pre-push hook) on macOS.**

BSD/macOS `sed` reads the argument after `-i` as the backup suffix, so `sed -i 's/…/…/' sample.go` consumes the script as the suffix and then treats `sample.go` as the script, erroring out. `.goreleaser.yaml` builds `darwin`, so macOS is a supported developer platform.

The failure cascades: both `sed` calls no-op, `git add -A && git commit -q` finds nothing to commit, `work` stays identical to `base`, `go test funcs:` comes back empty, and the "an edited assertion brings its test into scope" check fails. Since Makefile line 36 now makes `test` depend on `test-efficacy`, and `check: build-frontend lint test coverage`, that is `make check` — and therefore `.githooks/pre-push` — broken for every macOS contributor.

```suggestion
	sed -i.bak 's/return a + b/return a + b + 0/' sample.go && rm -f sample.go.bak
	sed -i.bak 's/if Add(1, 2) != 3 {/if Add(1, 2) != 3 || Add(0, 0) != 0 {/' sample_test.go && rm -f sample_test.go.bak
```

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,256 @@
...
+		{ line[NR] = $0 }
+		END {
+			n = NR
+			cnt = 0
+			for (i = 1; i <= n; i++) {
+				if (line[i] !~ /^func /) continue
+				cnt++
+				fstart[cnt] = i
+				fname[cnt] = ""
+				if (line[i] ~ /^func (Test|Benchmark|Fuzz|Example)[A-Za-z0-9_]*\(/) {
```
tomo-chan: **`Benchmark` is collected here, but `-run` never selects benchmarks — the run then selects nothing, exits 0, and is reported as a gate failure.**

Line 223 treats "`go test` exited 0" as "the tests passed", without checking that any test actually *ran*. Reproduced: a branch adding `Mul` plus only `BenchmarkAdd` (which still compiles after the revert):

```
efficacy: running the changed Go tests with the implementation reverted
  packages:  ./.
  tests:    BenchmarkAdd

FAIL: these tests still passed with this branch's implementation reverted.
      ok  	sample	0.002s [no tests to run]
EXIT=1
```

The quoted evidence literally says `[no tests to run]` while the verdict claims they passed. Build-tag-gated tests and `Example` functions with no `// Output:` comment reach the same state, and any future `-run` mismatch will too — a false red that a contributor has no obvious way to diagnose.

Worth distinguishing "ran and passed" from "selected nothing", e.g. `go test -list` first, or `-v` plus a check for at least one `--- PASS`/`--- FAIL` line.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,256 @@
...
+
+if [ -n "$fe_tests" ] && [ -d "$repo_root/frontend/node_modules" ]; then
+	ln -s "$repo_root/frontend/node_modules" "$worktree/frontend/node_modules" 2> /dev/null
+	echo "efficacy: running the changed frontend tests with the implementation reverted"
+	rel=""
+	for f in $fe_tests; do rel="$rel ${f#frontend/}"; done
+	echo "  tests: $rel"
+
+	# shellcheck disable=SC2086 # word splitting is the intent: one arg per file
+	if (cd "$worktree/frontend" && npx --no-install vitest run $rel > "$tmp/fe.log" 2>&1); then
```
tomo-chan: **Frontend scope is whole test *files*, not changed test cases — the exact masking the Go half was fixed to avoid.**

`$rel` is built from `$fe_tests`, i.e. every changed `*.test.ts(x)` file in full. A PR that edits `frontend/src/hooks/useBoardStatus.ts` and appends one tautological `it(...)` to `useBoardStatus.test.ts` runs that whole file; the pre-existing cases in it legitimately go red once the hook is reverted, vitest exits nonzero, and the script prints "ok — they went red". The new case is never distinguished.

docs/quality-gateway.md's D4 note and the PR description both present changed-function scoping as *the* decision that makes this gate meaningful ("an over-wide set lets an unrelated failure mask a tautological test"). That reasoning applies identically here, but the implementation only covers Go. Vitest's `-t`/`--testNamePattern` would give the same granularity if the changed `it`/`test` names can be extracted the way `changed_test_funcs` does for Go — at minimum this asymmetry is worth recording in the doc as a known gap rather than leaving the doc claiming a property the frontend half doesn't have.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,256 @@
...
+		echo
+		echo "FAIL: these frontend tests still passed with the implementation reverted."
+		echo
+		sed 's/^/      /' "$tmp/fe.log"
+		status=1
+	else
+		echo "  ok — they went red."
+	fi
+fi
```
tomo-chan, line 795: **Silent skip: when `frontend/node_modules` is missing the whole frontend red-check is dropped with no output at all.**

The `if` at line 237 has no `else`. If a developer runs `make efficacy` before `npm ci`, or a CI job ever reorders/omits `make install-deps-ci`, a PR whose only implementation change is under `frontend/src/` and whose only tests are frontend tests prints nothing for the frontend and exits 0 — green, having checked nothing.

Every other skip in this script says why it skipped (lines 39, 168, 178, 183, 195). This one is the odd exception, and it's the one where the silence hides a vacuous pass. A one-line `else echo "efficacy: frontend/node_modules missing — frontend red-check skipped."` would make it visible in the job log.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,256 @@
...
+set -u
+
+base=${EFFICACY_BASE:-origin/main}
+repo_root=$(git rev-parse --show-toplevel)
+mode=${1:-run}
+
+if ! git rev-parse --verify --quiet "$base" > /dev/null; then
+	echo "efficacy: base ref '$base' not found; set EFFICACY_BASE to a ref that exists"
+	exit 0
+fi
```
tomo-chan, line 65: **Fail-open on infrastructure errors: a required check that passes vacuously.**

Two paths exit 0 on what is an error, not a skip:

- here, when `$base` doesn't resolve;
- lines 194-198, when `git worktree add` fails.

`efficacy.yml` sets `EFFICACY_BASE: origin/${{ github.event.pull_request.base.ref }}` and relies on `fetch-depth: 0` for `origin/main` to exist. If that `fetch-depth` is ever dropped, the base branch is renamed, or the runner's disk fills so the worktree can't be created, the job prints one line and goes green. The PR proposes marking this job required — a fail-open means the gate can quietly stop protecting anything and nobody notices, which is the one failure mode a required check is supposed to rule out.

Worth distinguishing "nothing to check" (the deliberate skips at 178/183, correctly exit 0) from "couldn't check" (these two), and making the latter exit nonzero.

---

tomo-chan (commented):
Code review of the red-check gate. 8 findings, 6 reproduced locally against `scripts/efficacy.sh` as written on this branch.

The headline ones are the three that determine whether the gate does what D4 says it does:

1. **The gate is vacuous for every root-package test.** The scratch worktree has no `frontend/dist` (gitignored), so `//go:embed frontend/dist` makes `go test .` fail to build with zero files reverted. `board_test.go`, `bootstrap_test.go`, `command_center_test.go`, `main_test.go` and `url_open_main_test.go` always "go red".
2. **The verdict is one `go test` exit status over the union of all changed tests**, so a genuinely-red test masks a tautological one in the same PR. Verified with a two-package fixture where a pure tautology sails through. This is the same masking the PR says the scope-narrowing fix removed — narrowing to changed *functions* only removed the intra-file case.
3. **A cross-stack branch is failed for a revert that never happened**: the Go run isn't conditioned on `$go_impl` and the frontend run isn't conditioned on `$fe_impl`.

Plus a portability break (`sed -i` is GNU-only, and `test-efficacy` is now inside `make check`/pre-push), a false-red on `[no tests to run]`, file-level rather than test-level frontend scope, a silent frontend skip, and two fail-open exits.

The design and the documentation around it are genuinely good — these are all in the implementation, and (1)-(3) are worth fixing before this is marked required, since each one makes the check report green on something it was built to catch.

---

scripts/efficacy.sh
tomo-chan: Fixed in f628223, and confirmed the way you reproduced it.

Two changes, because the placeholder alone would only have fixed the instance you found:

- The scratch worktree now gets `frontend/dist/index.html`. It has to be a real filename — `go:embed` on a directory skips dotfiles, so a lone `.keep` still fails with `contains no embeddable files`.
- A new phase 1 runs every changed test at HEAD, with nothing reverted, and requires it to pass. That is the general backstop: any package that cannot build in the worktree is now an error naming the package, not a red.

`scripts/efficacy_test.sh` grows a fixture that is this exact shape — a module with `//go:embed frontend/dist`, `.gitignore` excluding it, and a tautology in the root package. It asserts the tautology is *named* (`SURVIVOR: ./. TestSub`), not just that the exit is 1, since a build failure exits 1 too. Deleting the placeholder line turns that check red for the right reason.

Also ran the real gate end to end against the branch for rollout item 2 (34 changed root-package tests): phase 1 passes for all of them now, and the whole run takes ~30s.

---

scripts/efficacy.sh
tomo-chan: You're right, and the PR description was wrong to claim the scope fix had removed this. Scope and verdict are separate decisions; narrowing the set never stops one member masking another.

Fixed in f628223: each `(package, test function)` pair is now run on its own, in both phases. Survivors are counted and named individually, so the output lists `red: ./a TestMul` and then `SURVIVOR: ./b TestBye`.

Your two-package fixture is now a test — `a` with a real `TestMul`, `b` with a `TestBye` that never mentions `Bye` — asserting the exact string `SURVIVOR: ./b TestBye`. It goes red if survivor accounting is removed.

D4's write-up in `docs/quality-gateway.md` is rewritten to say this outright, since the old text stated the weaker rule as if it were the intended one.

---

scripts/efficacy.sh
tomo-chan: Fixed in f628223, with the conditions you suggested: `go_count` is zeroed when `$go_impl` is empty and `fe_count` when `$fe_impl` is, so each stack is judged only against its own revert. The existing warn path then covers "implementation on one stack, tests on the other" and says which stack was left uncovered.

Both directions are now fixtures: frontend implementation + a Go test (expects the warning, exit 0), and the mirror — Go implementation + a touched frontend test — which asserts the Go half still ran (`red: ./. TestMul`) while the frontend half was skipped without needing node_modules. Removing the two conditions turns both red.

---

scripts/efficacy_test.sh
tomo-chan: Fixed in f628223. I went with a `rewrite <file> <old> <new>` helper over `sed -i.bak` — there are now several fixtures doing fixed-string edits, and one helper keeps the `.bak` cleanup from being repeated at each call site (and from being forgotten, which would leave `sample.go.bak` in the fixture for `git add -A` to commit). It shells out to `awk` with the strings passed through the environment, so no shell or regex metacharacter in either argument is reinterpreted.

Thanks for tracing the cascade — the failure being "the fixture silently commits nothing, so scope comes back empty" rather than an error is what would have made it slow to diagnose.

---

scripts/efficacy.sh
tomo-chan: Fixed in f628223, along the lines you suggested but from the `-v` output rather than `-list`.

Verdicts now come from the test's own line — `--- PASS: <fn> `, `--- FAIL: <fn> `, `--- SKIP: <fn> ` — with exit status used only to tell "the package did not build" from "`-run` selected nothing". So `pass`, `fail`, `skip`, `notrun` and `broken` are five distinct outcomes, and only `fail`/`broken` count as red.

Benchmarks are additionally dropped at collection: `-run` never selects them, and a benchmark asserts nothing that "must go red when the implementation is reverted" can meaningfully be applied to. Reaching phase 1 as `notrun` would have been correct but a worse message.

The `notrun` and `skip` cases are reported and counted, not failed — a test skipped at HEAD, or an `Example` with no `// Output:` comment, cannot be red-checked, and saying so is the honest answer. Both are fixtures now (benchmark-only branch, `t.Skip`ped test), as is the `--- FAIL` at HEAD case, which is an error: "it failed after the revert" only means something if it passed before.

I checked the `-list` route first — worth noting it does *not* list an `Example` with no `// Output:` comment, so it would not have covered that case on its own.

---

scripts/efficacy.sh
tomo-chan: Implemented rather than documented as a gap, in f628223 — the reasoning does apply identically, and leaving the doc claiming a property half the gate lacked was the worse option.

`changed_fe_test_names` maps touched lines onto `it(...)`/`test(...)` blocks the way `changed_test_funcs` maps them onto Go functions, and each name is run on its own with `-t` (regex-escaped).

One thing that only showed up once it was written: attributing a touched line outside every case as "unmappable" sends nearly every real change back to whole-file scope, because adding an `it` almost always edits the import line above the `describe`. So lines outside every case are ignored when at least one case was hit; the whole-file fallback is reserved for the cases where nothing can be narrowed safely — no `it`/`test` declaration in the file, a name that is not a literal (`it.each`, a template literal), or no touched line inside any case at all. Over-wide, never wrong.

Two fixtures pin both halves, driven through `--changed-tests` so they stay hermetic: one asserting the appended case is named and the untouched one is not, one asserting a change to `beforeEach` falls back to the whole file.

---

scripts/efficacy.sh
tomo-chan: Fixed in f628223, one step further than the suggested `else echo`: it exits 1.

This is the same category as the two fail-open paths below — the gate was asked to check something and could not — and it only fires when there is actually a frontend half to check (`fe_count > 0`), so a Go-only branch is unaffected and never touches node_modules. The message names the fix (`make install-deps`).

`efficacy.yml` already runs `make install-deps-ci`, so CI is unaffected; I've added a comment there recording that the step is now load-bearing rather than just convenient, so it doesn't get dropped as an optimisation later.

Fixture: a frontend-only branch run in a repo with no node_modules, asserting exit 1 and the message.

---

scripts/efficacy.sh
tomo-chan: Agreed, and fixed in f628223. Both paths now exit 1 with a message that says what to fix — the base-ref one names `fetch-depth: 0` explicitly, since that is the way it would actually happen.

The distinction you drew is now written into the script and into D4: "nothing to check" (no implementation changed, no test changed) exits 0 and says so; "could not check" exits 1. The missing-node_modules case above moved to the second category for the same reason.

Fixture asserts `EFFICACY_BASE=origin/no-such-ref` exits 1; reverting that one line to `exit 0` turns it red.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,555 @@
...
+# fe_verdict <relfile> <name|""> <log> — pass | fail | notrun | broken.
+fe_verdict() {
+	if [ -n "$2" ]; then
+		fv_pat=$(printf '%s' "$2" | sed 's/[][\\.*+?^$(){}|\/]/\\&/g')
+		(cd "$worktree/frontend" && NO_COLOR=1 npx --no-install vitest run "$1" -t "$fv_pat") > "$3" 2>&1
+	else
+		(cd "$worktree/frontend" && NO_COLOR=1 npx --no-install vitest run "$1") > "$3" 2>&1
+	fi
+	fv_rc=$?
+	fv_line=$(sed -n 's/^ *Tests  //p' "$3" | tail -n 1)
```
tomo-chan: **The frontend verdict is still not per case: `-t` is an unanchored match, so a sibling case's genuine red is read as the changed case's red.**

The Go half was fixed to take its verdict from the individual test's own `--- PASS:`/`--- FAIL:` line. This line takes it from the **aggregate summary** instead, and `-t` does not select a single case: vitest matches the pattern as an unanchored regex against the full `describe`-joined name. Measured against vitest 4.1.0 with three cases in one file:

```
-t 'renders'            -> 3 passed (3)      # renders / renders empty state / always renders
-t 'renders$'           -> 2 passed | 1 skipped
-t '^widget renders$'   -> 1 passed | 3 skipped   # the full name is "widget renders"
```

End to end, a branch adding `renderAll` with two cases — `it('renders')` a pure tautology, `it('renders empty state')` genuinely calling it:

```
efficacy: phase 2 — the same tests, with this branch's implementation reverted
  red: src/widget.test.ts -t "renders"
  red: src/widget.test.ts -t "renders empty state"
efficacy: ok — all 2 changed test(s) went red
real exit: 0
```

The first `red:` line is the tautology. It passes with the implementation entirely reverted; what went red in that run was its sibling. This is the same masking as the union-`go test` finding, surviving in the frontend half — and `docs/quality-gateway.md`'s new D4 text ("The frontend half does the same thing with `it(...)`/`test(...)` blocks and vitest's `-t`") states the property it doesn't yet have.

It is reachable in this repository today. Cases whose names are substrings of a sibling in the same file:

- `BoardDashboardPanel.test.tsx`, `CommandHistoryPanel.test.tsx`, `CommandPalette.test.tsx`: `"closes on Escape"` ⊂ `"closes on Escape even when the focused element stops propagation before window"`
- `schemas/index.test.ts`: `"accepts valid response"` ⊂ `"accepts valid response with hosts"`, `"rejects empty name"` ⊂ `"rejects empty names and missing flags"`

Suggest closing it the way the Go half was closed — read each case's own result rather than the summary (`--reporter=json`, or the per-test lines) and take the verdict for the exact name. Anchoring alone is not enough: `renders$` still pulls in `always renders`, per the second row above.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,555 @@
...
+	echo "  Use this only for a change whose tests are not expected to go red without it:"
+	echo "  a pure refactor, or a test-only rename. Say which in the pull request."
+	exit 0
+fi
+
+# Each stack is judged only against its own reverted implementation. A branch
+# that changes Go code and also touches a frontend test has nothing to revert
+# under that frontend test, so failing it there would be a verdict on a mutation
+# that never happened.
+[ -n "$go_impl" ] || go_count=0
```
tomo-chan, line 336: **Same false-verdict shape one level down: a Go test in a package this branch's implementation never touched is still judged against the revert.**

These two lines fix the cross-*stack* case, and the reasoning in the comment above them — "failing it there would be a verdict on a mutation that never happened" — applies unchanged to a Go test in an unrelated package. Reproduced: package `a` gets a real implementation plus its test; package `b` gets **no implementation change at all**, only a tightened assertion in an existing test:

```
  red: ./a TestMul

SURVIVOR: ./b TestBye still passes with this branch's implementation reverted.
efficacy: FAIL — 1 of 2 changed test(s) survived the revert.
```

Nothing under `TestBye` was reverted, so it was never going to go red. Tightening an assertion in a test next to the one you are changing, or fixing a flake alongside a feature, is an ordinary thing for a branch here to do, and this job is meant to be required.

I don't think per-package conditioning is the fix — that would be a real hole, since `internal/server`'s tests legitimately cover `internal/api`'s implementation, and narrowing by package would stop the gate seeing exactly that. The problem is more that the only remedy available is `efficacy-exempt`, which is PR-wide: applying it to get past one unrelated test also exempts a genuine tautology elsewhere in the same branch. Worth either a per-test exemption marker, or an explicit note in D4 that this is a known, accepted false-positive and the label is the answer — right now the doc's "each stack is judged only against its own revert" reads as if the class were closed.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,555 @@
...
+	esac
+done < "$tmp/checkable_fe"
+
+echo
+if [ "$survivors" -gt 0 ]; then
+	echo "efficacy: FAIL — $survivors of $run changed test(s) survived the revert."
+	exit 1
+fi
+
+if [ "$run" -eq 0 ]; then
```
tomo-chan, line 781: **This is the "green having checked nothing" case, and it exits 0 — the same shape the base-ref and node_modules paths were just changed to exit 1 for.**

`run == 0` means every changed test was `skip` or `notrun` at HEAD, so nothing was red-checked at all. The script says so, which is the honest half, but the exit status says the gate passed. Principle 4's argument for exiting 1 elsewhere in this file — "a required check that goes green having checked nothing is invisible: the job is green, so nobody looks" — applies to this line exactly as written.

It is reachable here rather than theoretical: `internal/board/agmsg_contract_test.go` and `internal/board/wrapper_script_integration_test.go` carry six `t.Skip` calls between them, skipping when a real agmsg install is absent — which is the state of every runner this job runs on. A branch that changes `internal/board` implementation and touches only tests in those two files gets a green efficacy check having red-checked nothing.

The distinction the rest of the script draws is "nothing to check" (no implementation changed, no test changed → exit 0) versus "could not check" (→ exit 1). All-unrunnable belongs in the second category. If exiting 1 feels too aggressive for a `t.Skip` that is simply environmental, the middle option is to keep 0 but make the line say it loudly enough to be greppable in a job log — though that is the same argument that was rejected for the missing-`node_modules` case, which is why I'd lean to exit 1.

---

scripts/efficacy_test.sh
```diff
@@ -0,0 +1,753 @@
...
+    expect(add(1, 2)).toBe(3)
+  })
+})
+TS
+	git add -A && git commit -q -m "base"
+	git checkout -q -b work
+
+	printf 'export const mul = (a: number, b: number) => a * b\n' >> frontend/src/math.ts
+	rewrite frontend/src/math.test.ts "import { add } from './math'" "import { add, mul } from './math'"
+	rewrite frontend/src/math.test.ts '})' '})'
```
tomo-chan: Nit: these two `rewrite` calls have no effect — the `cat > frontend/src/math.test.ts <<'TS'` three lines below rewrites the whole file, discarding both edits. This one is additionally a no-op in its own right (`'})'` → `'})'`).

Worth deleting both, since a reader comparing this fixture with the `fefallback` one below will otherwise take the `rewrite` calls as the thing that produces the diff being scoped, when it is the heredoc.

---

tomo-chan (commented):
Re-review of the fix round, against `8a4b380`.

**All eight findings from the previous round are fixed**, and each fix is the right shape rather than a patch over the symptom: per-`(package, test)` runs with survivors named individually, verdicts read from the test's own `--- PASS:`/`--- FAIL:`/`--- SKIP:` line, the `frontend/dist` placeholder plus phase 1 as the general backstop, both stacks conditioned on their own implementation diff, and `rewrite` replacing `sed -i` via awk with the strings passed through the environment (no metacharacter reinterpretation, and no `.bak` left for `git add -A` to commit). `sh scripts/efficacy_test.sh` is 20/20 in ~14s here.

Four new findings, three reproduced against the script as it stands on this branch:

1. **The frontend half's verdict is still not per case.** `-t` is an unanchored regex over the full `describe`-joined name, and `fe_verdict` reads the aggregate `Tests …` summary rather than the individual case's result. A tautological `it('renders')` is reported as `red:` because its sibling `it('renders empty state')` went red in the same invocation, and the gate exits 0. Same masking class as the union-`go test` finding, and reachable in this repository today (`"closes on Escape"` ⊂ `"closes on Escape even when …"` in three component test files, plus two in `schemas/index.test.ts`).
2. **A Go test in a package whose implementation this branch never touched is still failed as a SURVIVOR.** The stack-level conditioning closed the cross-stack case; the cross-package case has the same "a verdict on a mutation that never happened" shape, and the only remedy is the PR-wide `efficacy-exempt` label.
3. **`run == 0` exits 0** — the all-unrunnable case is "could not check", not "nothing to check", and the same Principle 4 argument that moved the base-ref and `node_modules` paths to exit 1 applies to it. Reachable via the six `t.Skip`s in `internal/board`'s agmsg tests.
4. A nit: two dead `rewrite` calls in the `fescope` fixture.

(1) is the one I'd want fixed before this is marked required — it is the exact property the fix round set out to give the frontend half, and the D4 text now claims it.

---

scripts/efficacy.sh
tomo-chan: You're right, and the sharpest part is that D4's text had already started claiming the property. Fixed in 5de6abc by closing it the way you suggest.

`-t` is gone entirely. The file now runs **once** with `--reporter=default --reporter=json --outputFile=…` and each case's status is looked up by name from `assertionResults`. That removes the anchoring problem rather than working around it — the filter was never the right tool, since neither `renders` nor `renders$` nor a `^…$` form can be built without the `describe` prefix, which the line-mapping does not have. It is also cheaper: one invocation per file instead of one per case.

Confirmed both reporters coexist in vitest 4.1.0 (human-readable output still goes to the log, JSON to the file), so failure detail is still there for the error paths.

One case the JSON cannot resolve: two cases sharing a title in the same file. There is nothing in a diff that distinguishes them, so a split result is reported as a **survivor** with "rename one of them", not guessed at — fail-closed is the only safe direction for "did the test this branch changed go red".

Your end-to-end fixture is now a test, run against real vitest with the repository's own `node_modules` (it reports itself skipped if they are absent, since `--changed-tests` could never have caught this):

```
SURVIVOR: src/widget.test.ts > renders still passes with this branch's implementation reverted.
  red: src/widget.test.ts > renders empty state
```

It fails against the previous head, which reported both as `red:` and exited 0.

---

scripts/efficacy.sh
tomo-chan: Agreed, and took exit 1 rather than the louder-message middle option, for the reason you name: it is the same argument that was rejected for missing `node_modules`, and accepting it here would leave the file drawing the distinction inconsistently.

Fixed in 5de6abc. The message names the way out, since an environmental `t.Skip` is a legitimate reason to be in this state:

```
efficacy: ERROR — nothing was red-checked. All 1 changed test(s) were skipped,
  absent or unrunnable here, so this branch's tests have not been checked
  against anything.
  ...
  If every changed test is legitimately environment-skipped — the agmsg contract
  tests are, without a real agmsg install — mark them with 'efficacy:exempt' or
  apply the efficacy-exempt label, and say which in the pull request.
```

Two fixtures, because the distinction matters: a skipped test **beside a checkable one** still reports the skip and exits 0 (that is the common shape and it is not an error), while a branch where every changed test is unrunnable exits 1. The second fails against the previous head.

Your `internal/board` example is what made this concrete rather than theoretical — six `t.Skip`s that fire on every runner this job will ever run on.

---

scripts/efficacy.sh
tomo-chan: Took your first suggestion in 5de6abc — a per-test marker — rather than only documenting the class as accepted.

`//efficacy:exempt` in the comment directly above a test takes that one test out of scope, on both stacks (the doc-comment walk was already there for the line mapping, so it was a two-line change in each awk). Your reasoning for why the label is the wrong instrument is what decided it: applying `efficacy-exempt` to get past one unrelated test also exempts a genuine tautology elsewhere in the same branch, so the blunt remedy actively costs coverage. The marker sits one line above the test it excuses, in the diff a reviewer reads.

You were also right that per-package conditioning would be a real hole, and D4 now says so explicitly rather than leaving it as an unstated reason — `internal/server`'s tests covering `internal/api`'s implementation is exactly the case the gate exists for, and narrowing by package would blind it.

The doc no longer reads as if the class were closed. D4 now names it as a deliberately-open false positive, with the marker as the answer.

Test: a two-package fixture where `b`'s implementation is untouched and only an assertion is tightened. Without the marker it reports `SURVIVOR: ./b TestBye` and exits 1; with the marker added it exits 0 — and package `a`'s genuine check still runs, so the marker takes out one test and nothing else.

---

scripts/efficacy_test.sh
tomo-chan: Both deleted in 5de6abc, and you're right about which half matters — not the wasted calls but that a reader comparing this fixture with `fefallback` would take them as the thing producing the diff.

Left a one-line comment saying the heredoc is what does it, so the next person does not add them back reaching for symmetry with the fixture below.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,667 @@
...
+done
+sort -u "$tmp/go_targets" -o "$tmp/go_targets"
+
+: > "$tmp/fe_targets"
+for f in $fe_tests; do
+	[ -f "$f" ] || continue
+	rel=${f#frontend/}
+	extracted=$(changed_fe_test_names "$f")
+	case "$extracted" in
+	"" | *__UNMAPPED__*) printf '%s%s\n' "$rel" "$tab" >> "$tmp/fe_targets" ;;
```
tomo-chan, line 297: **`//efficacy:exempt` inverts itself on the frontend side: when every hit case in a file is exempt, scope *widens* to the whole file instead of dropping the case.**

`changed_fe_test_names` skips exempt cases at print time (line 250), so a file whose only hit case is exempt produces no output at all. This line cannot tell that from "nothing could be narrowed safely", so it falls back to the whole file — and the exempt case runs anyway, as part of it. Reproduced against this branch, with `//efficacy:exempt` on the single changed case:

```
--- scope ---
frontend cases:
  src/widget.test.ts	(whole file)      <- the case that was just exempted
--- gate ---
SURVIVOR: src/widget.test.ts (whole file) still passes with this branch's implementation reverted.
efficacy: FAIL — 1 of 1 changed test(s) survived the revert.
exit=1
```

Applying the marker leaves the branch worse off than not applying it: without it the case is judged on its own; with it the whole file is. The partial case is fine — two changed cases with one exempt correctly narrows to the other (`red: src/widget.test.ts > renders empty state`, exit 0) — so this fires only when *all* hit cases in a file carry the marker, which is also the shape a frontend-only branch with one exempt case has.

The Go half has no equivalent problem: `changed_test_funcs` returning nothing just adds no targets, and there is no whole-file fallback to land in.

The gap that let it through is testable: `scripts/efficacy_test.sh`'s only marker fixture is Go-side (`b/b_test.go`), so `exempt[k]` in `changed_fe_test_names` has no test at all. Suggest having the function distinguish the two states — e.g. print `__ALL_EXEMPT__` when at least one case was hit but every hit was exempt — and have this `case` treat that as zero targets rather than as the fallback, with a fixture for each of the three shapes (all exempt, some exempt, unmappable).

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,667 @@
...
+# that as a survivor: fail-closed is the only safe direction when the question
+# is "did the test this branch changed go red".
+fe_status() {
+	FE_JSON="$tmp/fe_results.json" FE_TITLE="${1-}" FE_WANT_TITLE=$([ $# -gt 0 ] && echo 1 || echo 0) \
+		node -e '
+			const j = JSON.parse(require("fs").readFileSync(process.env.FE_JSON, "utf8"))
+			const cases = []
+			for (const f of j.testResults || []) for (const a of f.assertionResults || []) cases.push(a)
+
+			if (process.env.FE_WANT_TITLE !== "1") {
```
tomo-chan: Residual, worth recording rather than necessarily fixing now: this branch — the no-title, whole-file verdict — is the one place left that still judges by aggregate ("any case failed ⇒ failed"), which is the masking the per-case path was just restructured to remove. A tautological case inside a file that fell back to whole-file scope is still vouched for by its siblings.

The code comment above `changed_fe_test_names` calls the fallback "over-wide, never wrong". Over-wide is precisely the direction this gate *is* wrong in: a wider run is more likely to go red for someone else's reason, and going red is what the gate reads as success.

It is reachable here today. `qname` returns `""` for `it.each([`, so any touched line inside one of those blocks sends the whole file to this branch — `frontend/src/hooks/useTerminal.test.ts` has four `it.each` blocks and `frontend/src/schemas/index.test.ts` one, and both are large files with plenty of siblings to mask behind.

If it is worth closing later, vitest's JSON reporter can carry each case's source location (`includeTaskLocation`), which would let even an unmappable block be resolved by line range instead of by name. Documenting it as a known gap in D4 alongside the cross-package case would also be a reasonable answer — what I would avoid is leaving the "never wrong" phrasing, since it reads as if the fallback were the safe direction.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,667 @@
...
+while IFS="$tab" read -r pkg fn; do
+	[ -n "$fn" ] || continue
+	run=$((run + 1))
+	case $(go_verdict "$pkg" "$fn" "$tmp/rev.log") in
+	fail | broken)
+		echo "  red: $pkg $fn"
+		;;
+	pass)
+		echo
+		echo "SURVIVOR: $pkg $fn still passes with this branch's implementation reverted."
```
tomo-chan, line 707: Nit: this message doesn't mention `//efficacy:exempt`, and it is the message someone hitting the cross-package false positive actually reads — the `run == 0` error names both hatches, but a branch with one genuinely-red test beside one unrelated changed test lands here instead. A line along the lines of "if this test's implementation is not part of this branch, mark it `//efficacy:exempt` and say so in the pull request" would close the loop; the alternative reading of the current text is that the test is broken, which for this case it is not.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,667 @@
...
+			sed 's/^/         /' "$tmp/base.log"
+			status=1
+			continue
+		fi
+
+		awk -F"$tab" -v f="$fefile" '$1 == f { print $2 }' "$tmp/fe_targets" |
+			while IFS= read -r fename; do
+				if [ -n "$fename" ]; then
+					fe_state=$(fe_status "$fename")
+					label="$fefile > $fename"
```
tomo-chan: Nit: `label` is dead in this producer — it is assigned in both branches, the loop only `printf`s the state/file/name triple, and the consumer loop below recomputes `label` from those fields anyway (this loop is on the right-hand side of a pipe, so nothing set here survives it either).

---

tomo-chan (commented):
Fix verification for round 2, against `5de6abc`. **All four findings are fixed**, each confirmed by re-running my own fixtures rather than by reading the diff:

- **Frontend per-case verdict.** Dropping `-t` entirely and reading vitest's JSON reporter per case is a better answer than the anchoring I suggested — it sidesteps the describe-prefix problem completely, and one run per *file* is cheaper than one per case. My original fixture now reports `SURVIVOR: src/widget.test.ts > renders` with `red: src/widget.test.ts > renders empty state` beside it, exit 1. Treating a duplicated title as `ambiguous` and counting it as a survivor in phase 2 is the right direction for a question phrased as "did the test this branch changed go red".
- **`run == 0`.** Exits 1, and the message names both escape hatches.
- **Cross-package false positive.** `//efficacy:exempt` behaves as designed on the Go side: without it my fixture still reports `SURVIVOR: ./b TestBye` and exits 1; with it, scope narrows to `TestMul` alone and the branch passes. Recording the case as deliberately open in D4, rather than conditioning per package, is the right call for the reason the comment gives.
- Dead `rewrite` calls removed.

`sh scripts/efficacy_test.sh` is 23/23 here (~18s with `frontend/node_modules` linked in so the new per-case check really runs — note it reports itself skipped without them, which is correct but means a bare checkout does not exercise it).

**One new bug, in the new escape hatch**: `//efficacy:exempt` on the frontend widens scope to the whole file instead of removing the case, whenever every hit case in that file is exempt — so the exempted case runs anyway and the branch ends up worse off than if the marker were absent. Reproduced; details inline. The frontend half of `exempt[k]` has no fixture, which is how it got through — the marker's only test is Go-side.

Plus one residual and two nits inline. The residual is that the whole-file fallback still judges by aggregate, which is reachable today through `it.each` in `useTerminal.test.ts` and `schemas/index.test.ts`; it is a "document it or fix it later" call rather than a blocker, but the "over-wide, never wrong" phrasing in the code should not survive either way.

---

scripts/efficacy.sh
tomo-chan: Confirmed and fixed in 7a6c318, with your `__ALL_EXEMPT__` shape.

This is the worst kind of bug to have shipped in an escape hatch: applying the marker made the branch strictly worse off than leaving it alone. `changed_fe_test_names` now prints `__ALL_EXEMPT__` when at least one case was hit but every hit carried the marker, and the collection `case` adds no target for it — distinct from `__UNMAPPED__`, with a comment saying why the two must not collapse.

Your diagnosis of *why it got through* is the more useful half, and I took it as the fix: the marker's only fixture was Go-side, so `exempt[k]` had no test at all. There are now three, one per shape, driven through `--changed-tests`:

- all exempt → no cases in scope (fails against the previous head)
- partly exempt → narrows to the unmarked case only
- not a literal → left for resolution, does not silently vanish

The Go/frontend asymmetry you point out — `changed_test_funcs` returning nothing just adds no targets, with no fallback to land in — is why the Go half was fine, and it is also why the fallback itself was worth removing rather than patching. See the residual thread.

---

scripts/efficacy.sh
tomo-chan: Closed rather than documented, in 7a6c318 — `--includeTaskLocation` works exactly as you suggested, so the whole-file fallback is gone entirely.

A file the static extraction cannot map is now narrowed *after* the run, from the per-case source lines the reporter emits. One detail worth recording because it bit on the first attempt: the ranges need trailing blank and comment lines trimmed off, exactly as the static mapping does. Without that, appending a case touches the blank line above it and the location mapping drags the previous case in — my end-to-end fixture caught it reporting `SURVIVOR: … > adds` for a change that only added an `it.each` block.

`it.each` cases share one location, so all of them come back and all must go red. That is the right reading: the whole block was touched.

A file where no touched line lands in any case at all — a shared helper edit — is now reported as unresolvable and counted as unchecked, rather than run whole. That is the honest answer; any per-case verdict there would be arbitrary, and requiring every case in the file to go red would be far too strict.

Your phrasing point stands on its own and is applied in both places: neither the code nor D4 calls the fallback "over-wide, never wrong" any more. D4 now says outright that over-wide is the direction this gate is wrong in, since a wider run is likelier to go red for someone else's reason and going red is what it reads as success.

Fixture: an `it.each` block covering `renderAll`, with an untouched `adds` beside it. It resolves to `red: … > renders []` / `renders [1]`, never touches `adds`, and never prints `whole file`. It fails against the previous head.

---

scripts/efficacy.sh
tomo-chan: Applied in 7a6c318, to both survivor messages (Go and frontend) rather than only the one you flagged — the frontend has the same false positive for the same reason.

Your point about the alternative reading is the one that decided the wording: the old text says the test asserts nothing, which for this case is untrue and sends the contributor to fix a test that is fine. It now offers both readings and names the marker:

```
SURVIVOR: ./b TestBye still passes with this branch's implementation reverted.
      A test that passes without the code it covers protects nothing.
      ...
      If instead this test's implementation is simply not part of this
      branch — an assertion tightened beside the change, a flake fixed
      alongside a feature — nothing under it was reverted and it was
      never going to go red. Mark it '//efficacy:exempt' in the comment
      above it, and say so in the pull request.
```

---

scripts/efficacy.sh
tomo-chan: Gone in 7a6c318. The producer loop was rewritten anyway for the location resolution, and it now emits only the state/name pair — `label` exists solely in the two consumer loops that use it.

You were right that it was dead twice over: unused in the producer, and unable to escape the pipe even if it had been.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,758 @@
...
+		for (const f of j.testResults || []) for (const a of f.assertionResults || []) {
+			if (a.location && a.location.line) cases.push({ title: a.title, line: a.location.line })
+		}
+		if (cases.length === 0) process.exit(0)
+
+		const src = fs.readFileSync(process.env.FE_SRC, "utf8").split("\n")
+		const blank = (i) => i < 1 || i > src.length || /^\s*(\/\/.*)?$/.test(src[i - 1])
+
+		const starts = [...new Set(cases.map((c) => c.line))].sort((a, b) => a - b)
+		const ranges = starts.map((start, i) => {
```
tomo-chan, line 532: **Ranges built from the reported line put a case's own leading lines inside the *previous* case, so location resolution drags an untouched test into scope.**

vitest reports a case at the line of the `it(...)`/`])(...)` **call**, not at the first line of the block. For a multi-line `it.each` those are different lines, and everything between them lands in the preceding range. Measured against vitest 4.1.0:

```
 9:  it.each([          <- block starts here
10:    [2, 3],
11:    [4, 5],
12:  ])('tautological row %i,%i', (a, b) => {

reported location ->  5 | adds
                     12 | tautological row 2,3
                     12 | tautological row 4,5
                     16 | multiplies
```

`starts` is therefore `[5, 12, 16]`, so `adds` spans 5–11 and swallows lines 9–11 of the block below it. End to end on a branch that adds `mul` plus that `it.each` block:

```
SURVIVOR: src/math.test.ts > adds still passes with this branch's implementation reverted.
  red: src/math.test.ts > multiplies
SURVIVOR: src/math.test.ts > tautological row 2,3 still passes ...
SURVIVOR: src/math.test.ts > tautological row 4,5 still passes ...
efficacy: FAIL — 3 of 4 changed test(s) survived the revert.
```

The two `tautological row` lines are the fix working. `> adds` is the bug: this branch never touched it, nothing under it was reverted, and it was never going to go red — the same false positive the `//efficacy:exempt` marker was added for, now produced by the gate itself against a test the author has no reason to mark. It is the rule the static mapper documents at length ("the blank line and the doc comment BETWEEN two functions belong to the one below"), reappearing in the location path. A multi-line `it(` call or a doc-commented case reaches it the same way.

The new fixture cannot catch it because its `it.each` is written on one line (`it.each([[[], ''], [[1], '1']])('renders %j', …)`), where the reported line *is* the block's first line and nothing sits above it. All five real occurrences in this repository are the multi-line form — `useTerminal.test.ts:1196,1212,1225,1252` and `schemas/index.test.ts:781` all open with `it.each([` and put rows on the following lines — so the mechanism this was built for is exactly where it misfires.

Suggest not deriving the starts from the report at all: `changed_fe_test_names`'s scanner already finds each block's first line with `^[ \t]*(it|test)(\.[A-Za-z]+)?[ \t]*\(`, including the ones whose names are not literals. Passing those lines in as the range starts and using vitest's reported line only to assign each title to a block would make both paths agree by construction, and a multi-line `it.each` fixture would pin it.

---

scripts/efficacy.sh
```diff
@@ -0,0 +1,758 @@
...
+if [ "$mode" = "--changed-tests" ]; then
+	echo "base:            $base ($merge_base)"
+	echo "go impl:        $go_impl"
+	echo "go tests:       $go_tests"
+	echo "go test funcs:  $(cut -f2 "$tmp/go_targets" | sort -u | tr '\n' ' ' | sed 's/ *$//')"
+	echo "go packages:    $(cut -f1 "$tmp/go_targets" | sort -u | tr '\n' ' ' | sed 's/ *$//')"
+	echo "frontend impl:  $fe_impl"
+	echo "frontend tests: $fe_tests"
+	echo "frontend cases:"
+	sed 's/^/  /; s/'"$tab"'$/'"$tab"'(whole file)/' "$tmp/fe_targets"
```
tomo-chan: `(whole file)` is stale now — nothing runs whole-file any more. A `__UNMAPPED__` file's real fate is "resolved by source location after the run", or, when no touched line lands inside any case, `no case could be resolved, so not red-checked`.

That second outcome is a behaviour change worth stating explicitly somewhere, because it flips a verdict: a branch whose only frontend test change is describe-level setup — a `beforeEach` gaining a mock the new implementation needs, with no case body touched — used to run the file whole and pass, and now exits 1 with "nothing was red-checked". Reproduced on this branch with exactly that shape. It follows from Principle 4 and the table's "unresolvable" row covers it, so I read it as intended rather than a bug; it just isn't obvious from the table that the common setup-only edit lands there.

Something like `(resolved after the run)` would keep the listing honest about what it will do.

---

scripts/efficacy_test.sh
```diff
@@ -0,0 +1,1053 @@
...
+	git checkout -q -b work
+
+	printf 'export const mul = (a: number, b: number) => a * b\n' >> frontend/src/math.ts
+	rewrite frontend/src/math.test.ts 'beforeEach(() => {})' 'beforeEach(() => { /* reset */ })'
+	git add -A && git commit -q -m "touch describe-level setup"
+)
+checks=$((checks + 1))
+listing=$(run_efficacy "$fefallback" --changed-tests)
+if printf '%s' "$listing" | grep -q '(whole file)'; then
+	pass "a touched line outside every case falls back to the whole file"
```
tomo-chan: This check's name — and the comment at the top of the fixture, "so the whole file runs" — describe behaviour this commit removed. It still passes because it asserts only the `--changed-tests` listing string, which is the stale `(whole file)` label; run the same fixture for real and the outcome is `no case could be resolved, so not red-checked` followed by exit 1.

Worth renaming to what it now pins (a touched line inside no case yields no static name) and, since this suite's whole point is that a green check has to mean what it says, asserting the run's outcome rather than the listing alone.

---

tomo-chan (commented):
Fix verification for round 3, against `7a6c318`. **All four items are fixed**, each re-run against my own fixtures:

- **The exempt marker no longer inverts.** `__ALL_EXEMPT__` as a token distinct from `__UNMAPPED__` is the right shape. My repro — the single changed frontend case carrying the marker — now yields an empty `frontend cases:` list and the WARNING path at exit 0, instead of widening to the file. Three fixtures, one per shape, is the coverage that was missing.
- **The whole-file aggregate is gone**, implemented rather than documented. `--includeTaskLocation` resolving `it.each` per case is a better answer than the "document it as a gap" half of what I suggested: an `it.each` block of two tautologies is now named case by case while its real sibling shows as `red:`.
- Both SURVIVOR messages name the marker, and the dead `label` is gone.

`sh scripts/efficacy_test.sh` is 27/27 here with `frontend/node_modules` linked in so both vitest-driven checks really run.

**One new bug in the location resolution**, and it is visible in the output of the fixture above: alongside the two correctly-named tautologies, the run also reports `SURVIVOR: … > adds` — a case this branch never touched. vitest reports a case at its *call* line, so for a multi-line `it.each` the block's opening lines fall inside the preceding case's range. That is the "lines above a case belong to the case below" rule the static mapper is careful about, reappearing in the new path, and it produces a false failure on a test the author has no reason to mark exempt. The new fixture writes its `it.each` on one line, where the bug cannot appear; every real `it.each` in this repository is the multi-line form. Details and a suggested fix inline.

Not a false green — it blocks rather than passes — so it is less severe than the last two rounds' findings, but it fires precisely in the files the location path was added for.

Plus two stale-label items inline: `--changed-tests` still prints `(whole file)`, and the fixture that asserts that string is still named after the removed fallback. The second one matters a little more than a wording nit, since the check passes while describing behaviour that no longer exists — and the same fixture's real outcome is now exit 1 rather than a whole-file run.

---

scripts/efficacy.sh
tomo-chan: Reproduced your exact output against `7a6c318` before changing anything, `> adds` and all, and fixed in dd4de2d by taking your suggestion: the starts no longer come from the report.

`fe_block_starts` runs the same scanner and the same doc-comment walk `changed_fe_test_names` uses, so the two paths agree about where a case begins by construction rather than by two implementations happening to match. The reported line is now used for one thing only — deciding which block each title belongs to. After the fix:

```
  red: src/math.test.ts > multiplies
SURVIVOR: src/math.test.ts > tautological row 2,3 still passes ...
SURVIVOR: src/math.test.ts > tautological row 4,5 still passes ...
efficacy: FAIL — 2 of 3 changed test(s) survived the revert.
```

The tautologies are still caught; `adds` is gone.

Your point about the fixture is the one that stings, and it is the second time in this PR that a fixture has been written in the one shape where the bug it should have caught cannot appear. A one-line `it.each` has its reported line *at* the block's first line, so it could never see this — and all five real occurrences here are multi-line. There is now a multi-line fixture, and it fails against `7a6c318`.

I had this exact rule written out at length in the static mapper's comment and reintroduced the bug one function away. The shared scanner is what stops that recurring, rather than a third copy of the rule in a comment.

---

scripts/efficacy.sh
tomo-chan: Relabelled to `(resolved after the run)` in dd4de2d.

The behaviour change is the more important half and you are right that the table did not make it findable. D4 now states it outright rather than leaving it to be derived from the "unresolvable" row:

> One consequence is worth stating because it flips a verdict: a branch whose only frontend test change is describe-level setup — a `beforeEach` gaining a mock the new implementation needs, with no case body touched — has no case to judge, so it lands in "could not check" and fails, where it previously ran the file whole and passed.

Your reading of it as intended rather than a bug matches mine: any per-case verdict there would be arbitrary, and requiring every case in the file to go red would be far too strict. But "follows from Principle 4" is not the same as "a contributor will work out why their green branch went red", which is what the explicit sentence is for.

It is a fixture now too, run for real rather than asserted through the listing — see the test-file thread.

---

scripts/efficacy_test.sh
tomo-chan: Both done in dd4de2d — renamed to `a touched line inside no case yields no static name`, which is what it actually pins, and the fixture comment no longer claims the file runs whole.

Your second half is the one worth acting on rather than just the name: a check that asserts a listing string is not asserting an outcome, and this suite exists to make green mean something. There is now a companion check that runs the same shape for real, under the `node_modules` guard, and requires exit 1 with both `no case could be resolved` and `nothing was red-checked`.

Suite is 29. Worth noting that the renamed check and the outcome check are complementary rather than redundant: the listing one is hermetic and pins what scope decides, the run one pins what the gate does about it — and the old arrangement had the first standing in for the second, which is how it kept passing while describing removed behaviour.

---

PR #186, opened 2026-08-23
test: make the scenario ledger real (#180 item 5)

tomo-chan:
Roadmap item 5 of #180 — gates **G0** (spec) and **G5** (scenario) in [docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md).

- [x] コアのマルチプレクサ機能のセクションを追加する
- [x] #177 の機能（loopback ポートフォワード / ブラウザオープン傍受）の行を追加する
- [x] `auto` 行が指すテスト名が実在することを検査する仕組みを追加する
- [x] コア機能の E2E を追加する（分割・リサイズ・レイアウト復元・ワークスペース CRUD）
- [x] 差分に `frontend/src` / `internal/api` / `internal/config` を含み `scenarios.md` が無変更なら CI で fail させる（ラベルで免除可）

## 未記入の台帳は台帳ではない

`docs/scenarios.md` says a `manual` row is a legitimate answer and a silently absent row is not. That rule was written down and then not followed: **the core multiplexer — the feature panemux exists to provide — had no rows at all**, while Agent Board and the command center were mapped in detail, and #177 shipped a whole user-facing feature without adding one either.

- **Section H** — panes, splits, layout persistence, workspaces, session types, reconnect (15 rows).
- **Section I** — #177's two mechanisms: the browser-open shim and loopback port forwarding (14 rows).

Both sit below G, so the section letters read A through I in order.

## The E2E that lets H say `auto`

`frontend/e2e/core-multiplexer.spec.ts`: splitting horizontally and vertically, closing, dragging a divider to resize, restoring a layout across a reload, and workspace add/rename/delete.

It gets **its own panemux process and fixture** (port 4177), following `agent-board.yml`'s precedent, because it mutates the layout and the suite runs serially against long-lived servers. Sharing the default server was tried first and made `pane-url-open.spec.ts` fail intermittently — the worst way for cross-test coupling to surface, since it points at the wrong test. Baseline was verified stable 3/3 before and the new arrangement 3/3 after, plus the full suite.

Four things the spec had to learn, each recorded in a comment because each first appeared as a product bug that wasn't one:

| Symptom | Actual cause |
|---|---|
| "the layout is not persisted" — a reload after a change showed the old layout | `useLayout`'s save is debounced 500ms; the test has to await the layout `PUT` |
| "delete workspace does nothing" — no `DELETE` request at all | it goes through `window.confirm`, which Playwright dismisses by default |
| the resize test could reload before the debounced `PUT` fired | it subscribed to the `PUT` *after* the split, so it could resolve on the split's own `PUT` (found in review) |
| the post-reload poll never actually waited | it polled the bound an un-restored layout also satisfies, leaving the discriminating assertion un-retried (found in review) |

## The cross-check is the half that matters

`scripts/scenarios_check.sh` (`make check-scenarios`, inside `make check`) resolves every path and Go test name an `auto` row claims — 139 tokens across 80 rows, in under a second.

Requiring a row is the obvious half of G0 and the weaker one. A ledger's likeliest failure is not an **absent** row but a **stale** one — naming a test since renamed, moved or deleted. Such a row reads as coverage and is worth nothing, and nobody grepping for a test name expects to find it in a markdown table.

**It found one the day it was written**: C7 named `..._TransportError_DistinctFromNo`; the test is `TestBootstrapWatcher_RemotePresenceCheckTransportError_DistinctFromNo`. Fixed here.

Because it reads prose, its own risk is false positives — and all three of these were, in the first version:

| Looks like a path | Actually |
|---|---|
| `Cmd/Ctrl+Shift+B` | a keystroke |
| `/ws/board-command` | an HTTP route |
| `bin/panemux` | a build artifact that does not exist on a clean checkout |

All three are regression-tested in `scripts/scenarios_check_test.sh` (`make test-scenarios-check`, 22 checks).

### Three ways it could have reported all clear having checked nothing

All three are guarded, and each guard has a fixture. They are worth listing together because they are the same failure wearing different clothes, and two of the three were found by review rather than by writing the script:

| Shape | Guard |
|---|---|
| There were no rows to read (the table format changed) | the `auto`-row grep must match something |
| The loop stopped part-way (an unopenable `$TMPDIR` redirect) | `__CHECKED__` reports rows processed; it must equal the row count |
| Every row was read and **none** was recognised (a column that stopped backticking its values) | `$resolved` must be non-zero |

The third needs no infrastructure failure at all — a markdown reformat is enough — which makes it the likeliest and the quietest.

## It caught a real stale row the first time it met `main` (750350f)

Not a constructed case. Merging the already-landed sibling PRs into this branch, the first thing `make check-scenarios` did was fail:

```
I9 names internal/server/openurl_routes_test.go, which does not exist
```

#182 consolidated the per-route server tests and that file went with them. Row I9 survived the consolidation still naming it — the exact shape this gate exists for, and one no other check in the repository could see: `make check` was green on both branches independently, because neither the row nor the deletion is a compile or test error.

The coverage itself had not moved: `TestServer_APIIntegration` drives `POST /api/sessions/{id}/open-url` through the real router and asserts I9's own claim for a local pane — `"forwarded":false` **plus the reason**, deliberately rather than the flag alone. The row now names that.

## The CI gate

`.github/workflows/scenarios.yml` fails a pull request that changes `frontend/src`, `internal/api` or `internal/config` without touching `docs/scenarios.md`. Exempt with the **`scenarios-exempt`** label.

The trigger is deliberately those three directories. A change in `internal/session` can be user-visible but far more often is not, and a gate that fires on changes it has no opinion about gets bypassed (principle 4). Test-only changes are excluded. All four paths were verified against simulated branches before the workflow was written:

| Simulated branch | Result |
|---|---|
| this PR (no `frontend/src` change) | skip |
| `frontend/src/App.tsx` + `scenarios.md` | pass |
| `internal/api/handler.go` alone | **fail**, exit 1 |
| `internal/api/handler_test.go` alone | skip |

## Review round 1 (010b9aa)

Nine findings, two of which meant the gate could report green having checked nothing. All fixed; each new assertion was confirmed to fail against the previous version of the script.

- **`scripts/scenarios_check.sh` failed open.** It collected findings in a temp file under `$TMPDIR`; a redirect that could not be opened left them empty, and empty is how the script says "everything resolved" — so a read-only, full or stale `TMPDIR` turned a failing ledger green (reproduced with `TMPDIR=/nonexistent-dir`).
- **The `scenarios-exempt` label could not be applied to a failing run.** `on: pull_request` had no `types:`, so labelling did not re-run the workflow and a re-run replayed the stored payload. The same defect was in #185's `efficacy.yml`; fixed there too.
- **A test name without a trailing `*` still matched by prefix**, so the most common kind of rename — appending a suffix — left the row resolving to a function it no longer names.
- **stderr was folded into the findings**, so any tool error was announced under "these auto rows name something that does not exist".
- **`echo "$rows"` expands backslash escapes under dash**, which is what `sh` is on Debian and Ubuntu, rewriting a row before it was tokenised.
- `find` filtered `node_modules` rather than pruning it; `test-scenarios-check` never ran in CI (`ci.yml` runs neither `make test` nor `make check`); the two E2E issues in the table above; and sections H/I were out of alphabetical order.

## Review round 2 (4d4c81a)

One finding, on the guard round 1 added rather than on anything from round 1: `__CHECKED__` proved the loop *finished*, but nothing proved it *recognised* anything. `$resolved` was computed, printed, and never asserted on, so a ledger whose Verification column stopped backticking its values reported `ok — 0 path and test names across 2 auto rows all resolve` and exited 0. Reproduced verbatim, then guarded — the third row of the table above.

The guard sits **after** the findings check on purpose: a ledger whose rows all name nonexistent tests also resolves nothing, and there the format is fine and the missing names are the real answer. A second fixture fails if the guard is ever moved ahead of them.

## Test plan

- [x] `make check-scenarios` — 139 tokens across 80 auto rows all resolve, in ~0.8s (after fixing the stale C7 row, the stale I9 row, and after exact-match was tightened)
- [x] `make test-scenarios-check` — 22/22
- [x] Each of the five new checks confirmed to fail against the version of the script it was written for
- [x] `make test-e2e` (core-multiplexer project) — 5/5, run 3× consecutively after the review fixes
- [x] Full `make test-e2e` — 28/28
- [x] Baseline suite confirmed stable 3× before, to establish that the contamination was this spec's
- [x] CI gate logic exercised on four simulated branches
- [x] `make lint` — 0 issues
- [x] `make check` — passes end to end on the branch with `main` merged in (now including `check-scenarios` and `test-scenarios-check`, alongside #184's `test-hooks` and #185's `test-efficacy`)


scripts/scenarios_check.sh
```diff
@@ -0,0 +1,128 @@
...
+				resolved=$((resolved + 1))
+			else
+				report_missing "$id names Go test ...$suffix, which no _test.go file defines"
+			fi
+			;;
+		esac
+	done
+done > "${TMPDIR:-/tmp}/scenarios_check.$$" 2>&1
+
+output=$(cat "${TMPDIR:-/tmp}/scenarios_check.$$")
```
tomo-chan: **The gate fails open when the temp file cannot be created.** `set -e` is not enabled, so if the `done > "${TMPDIR:-/tmp}/scenarios_check.$$"` redirect on line 113 fails, the whole `while` loop never runs, `cat` on line 115 fails, `$output` is empty, and the script prints `ok — every path and test name in the auto rows resolves` and exits `0` having verified nothing.

Reproduced on this branch:

```
$ TMPDIR=/nonexistent-dir sh scripts/scenarios_check.sh
scripts/scenarios_check.sh: 53: cannot create /nonexistent-dir/scenarios_check.17584: Directory nonexistent
cat: /nonexistent-dir/scenarios_check.17584: No such file or directory
  ok — every path and test name in the auto rows resolves
EXIT=0
```

A read-only or full `/tmp`, or a stale `TMPDIR` exported by a wrapper, is enough. This is exactly the failure mode `scenarios_check_test.sh` guards against for the empty-ledger case ("the way a gate quietly stops working") — the same guard is missing here.

The simplest structural fix is to drop the temp file entirely and keep the loop in the current shell, e.g. `while ... done <<EOF ... EOF` with a here-doc, or capture with `output=$(echo "$rows" | while ...; done)`, so a redirect that cannot be opened aborts the script instead of emptying its result. Failing that, check the file exists before trusting it:

```suggestion
if [ ! -f "${TMPDIR:-/tmp}/scenarios_check.$$" ]; then
	echo "scenarios: could not write to ${TMPDIR:-/tmp} — the check did not run"
	exit 1
fi
output=$(cat "${TMPDIR:-/tmp}/scenarios_check.$$")
```


---

.github/workflows/scenarios.yml
```diff
@@ -0,0 +1,78 @@
...
+# changes it has no opinion about gets bypassed (design principle 4).
+#
+# No `paths:` filter on the workflow itself, for the reason
+# agmsg-contract.yml records: a path-filtered workflow reports no status at all
+# on the pull requests it skips, so a required check that never reports blocks
+# every one of them. The job decides in seconds instead.
+
+on:
+  pull_request:
+    branches:
```
tomo-chan, line 32: **The documented `scenarios-exempt` escape hatch cannot actually be applied to a failing PR.** With no `types:`, `on: pull_request` defaults to `opened, synchronize, reopened` — labelling a PR does not re-run this workflow, and re-running the failed job replays the *stored* event payload, in which `github.event.pull_request.labels` still does not contain the new label. So the `if:` on line 48 keeps evaluating to true.

Scenario: a contributor changes `internal/config/validate.go` only (a rename with no use-case impact), the job fails, they follow the message on line 77 and apply `scenarios-exempt` — nothing happens, and the only way out is an empty commit to fire `synchronize`.

```suggestion
on:
  pull_request:
    types: [opened, synchronize, reopened, labeled, unlabeled]
    branches:
```


---

frontend/e2e/core-multiplexer.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  const divider = page.locator('[data-divider-drop-zone="horizontal"]').first()
+  const box = await divider.boundingBox()
+  expect(box).toBeTruthy()
+
+  // The layout PUT is debounced by 500ms (useLayout's updateSizes), so the
+  // reload below has to wait for it. Reloading on the drag alone reads the
+  // config as it was before the drag and looks exactly like "resize is not
+  // persisted" — which is what this test reported before the wait was added.
+  const saved = layoutSaved(page)
```
tomo-chan, line 135: **This subscribes to the layout `PUT` *after* the split, so it can resolve on the split's own `PUT` instead of the debounced resize one** — reintroducing precisely the flake the comment above says it fixes.

`useLayout.splitPane` ends with `await saveLayout(newLayout)`, i.e. it issues `PUT /api/workspaces/{id}/layout` itself. `page.waitForResponse` only sees responses that arrive after it is called, so this works only while that response lands during the three CDP round-trips between the split (line 106) and line 120. `PutLayout` → `Config.SaveLayout` writes the config file synchronously; under CI I/O contention that can easily exceed those few milliseconds.

When it does, `saved` resolves on the split's `PUT`, `expect((await saved).status()).toBe(200)` on line 138 returns immediately, and `page.reload()` runs before the 500 ms `updateSizes` debounce has fired — the pending `setTimeout` dies with the page, the resize is never persisted, and line 143 fails with "a resize is a layout change, so it must be persisted like any other". A failure pointing at the product rather than the test, which is the outcome the comment set out to avoid.

Either await the split's own `PUT` explicitly before line 120, or move `const saved = layoutSaved(page)` to just after `page.mouse.up()` — the debounced `PUT` fires ~500 ms after the last `mouse.move`, so it is still comfortably ahead of the subscription.

---

frontend/e2e/core-multiplexer.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+  expect((await saved).status()).toBe(200)
+
+  await page.reload()
+  await expect(panes(page).first()).toBeVisible()
+  await expect
+    .poll(async () => (await panes(page).first().boundingBox())?.width ?? 0, {
+      message: 'a resize is a layout change, so it must be persisted like any other',
+    })
+    .toBeGreaterThan(widthAfter - 30)
+  expect((await panes(page).first().boundingBox())?.width ?? 0).toBeLessThan(widthAfter + 30)
```
tomo-chan: **The polled bound is satisfied by the un-restored width, so the poll does no waiting and the real assertion runs exactly once.**

After the split the pane is ~50% wide (`widthBefore`); the drag narrows it to `widthAfter ≈ widthBefore - 120`. Post-reload, a layout that was *not* restored comes back at `widthBefore`, and `widthBefore > widthAfter - 30` — so `expect.poll(...).toBeGreaterThan(widthAfter - 30)` passes on its first sample in both the good and the bad case. The only assertion that can actually detect the failure is `toBeLessThan(widthAfter + 30)` on this line, and it has no retry.

Two consequences: the poll's stated purpose (waiting for the restored layout to settle) is not achieved, and if the first paint after reload is transiently 50/50 before the persisted sizes are applied, the poll returns on that transient value and this line fails — a flake in the "product is broken" direction.

Poll the bound that actually discriminates, e.g.

```suggestion
  await expect
    .poll(async () => (await panes(page).first().boundingBox())?.width ?? 0, {
      message: 'a resize is a layout change, so it must be persisted like any other',
    })
    .toBeLessThan(widthAfter + 30)
  expect((await panes(page).first().boundingBox())?.width ?? 0).toBeGreaterThan(widthAfter - 30)
```


---

scripts/scenarios_check.sh
```diff
@@ -0,0 +1,128 @@
...
+				resolved=$((resolved + 1))
+			else
+				report_missing "$id names $token, which does not exist"
+			fi
+			;;
+		# A Go test function named in full, possibly with a trailing wildcard
+		# for a family of them (TestBuildBootstrapInstruction_*).
+		Test*)
+			pattern=${token%\*}
+			if grep -rqE "^func ${pattern}[A-Za-z0-9_]*\(" --include='*_test.go' "$repo_root"; then
```
tomo-chan: **A test name written without a trailing `*` still matches by prefix, so the most common kind of rename slips through the gate.** The `[A-Za-z0-9_]*` suffix is needed for the wildcard branch, but it is applied to exact names too.

Reproduced on this branch — `TestValidate_AgentBoardMode_InvalidValue` does not exist (the real test is `..._InvalidValue_Error`), and the checker passes it:

```
| Z1 | x | y | `auto`: `TestValidate_AgentBoardMode_InvalidValue` |
→ ok — every path and test name in the auto rows resolves   (exit 0)
```

Renames overwhelmingly *append* (`TestFoo` → `TestFoo_RemoteVariant`, `TestBootstrapWatcher_X` → `TestBootstrapWatcher_X_AndY`), which is exactly the case this cannot catch — including the stale C7 row the PR description cites, had it been shortened rather than restructured.

Anchor the exact form and keep the loose match only for the wildcard:

```suggestion
			if [ "$token" != "$pattern" ]; then
				expr="^func ${pattern}[A-Za-z0-9_]*\("
			else
				expr="^func ${pattern}\("
			fi
			if grep -rqE "$expr" --include='*_test.go' "$repo_root"; then
```


---

scripts/scenarios_check.sh
```diff
@@ -0,0 +1,128 @@
...
+			suffix=${token#...}
+			if grep -rqE "^func Test[A-Za-z0-9_]*${suffix}\(" --include='*_test.go' "$repo_root"; then
+				resolved=$((resolved + 1))
+			else
+				report_missing "$id names Go test ...$suffix, which no _test.go file defines"
+			fi
+			;;
+		esac
+	done
+done > "${TMPDIR:-/tmp}/scenarios_check.$$" 2>&1
```
tomo-chan: **`2>&1` folds every tool error into the missing-row report**, so an infrastructure failure is announced as "these auto rows name something that does not exist" — the one message that is guaranteed to send the reader looking in the wrong place.

Anything the loop body writes to stderr lands in `$output` and therefore fails the gate under that headline: `find` hitting an unreadable directory during the whole-repo fallback scan on line 79, or `grep` rejecting a pattern. Demonstrated on this branch:

```
$ sh scripts/scenarios_check.sh /tmp/l.md
FAIL: these auto rows name something that does not exist.
      ...
grep: Unmatched ( or \(
  Z1 names Go test Test(bad, which no _test.go file defines
```

Drop the `2>&1` so stderr goes to the terminal as diagnostics rather than being counted as a finding, and the exit status keeps meaning "a row did not resolve".

Related, on the same construct: `find "$repo_root" -path "*/$target" -not -path '*/node_modules/*'` filters rather than prunes, so it still descends into `frontend/node_modules` and `.git`. On a checkout with dependencies installed that is tens of thousands of stat calls per unresolved token, inside a target the Makefile advertises as "hermetic and fast" and puts in `make check`. `-name node_modules -prune -o ...` would keep it fast.

---

scripts/scenarios_check.sh
```diff
@@ -0,0 +1,128 @@
...
+if [ -z "$rows" ]; then
+	echo "scenarios: no \`auto\` rows found in $ledger — is the table format still what this expects?"
+	exit 1
+fi
+
+echo "scenarios: checking what $ledger's auto rows name"
+
+# Each row is checked on its own so a failure can name the scenario it came
+# from; a bare "TestFoo not found" would send the reader grepping.
+echo "$rows" | while IFS= read -r row; do
```
tomo-chan: **`echo "$rows"` is not safe for arbitrary markdown here.** `make check-scenarios` runs this with `sh`, which on Debian/Ubuntu (including `ubuntu-latest`) is `dash`, whose builtin `echo` interprets backslash escapes with no `-e` needed.

`docs/scenarios.md` happens to contain no backslashes today, but a literal `|` inside a table cell must be written `\|` in GitHub-flavoured markdown, and `\t` / `\n` / `\\` are all plausible in a Verification column that quotes commands. Any of those is silently rewritten before the row is tokenised — `\|` becomes `|`, which also splits the row into different `awk -F'|'` fields and misreports the scenario ID in a failure message.

`printf` is already used correctly for `$row` a few lines below; the same applies here:

```suggestion
printf '%s\n' "$rows" | while IFS= read -r row; do
```


---

Makefile
```diff
@@ -44,6 +44,24 @@ test-frontend:
...
+# moved or deleted. Such a row reads as coverage and is worth nothing.
+#
+# Hermetic and fast, so it sits inside `make check`.
+check-scenarios:
+	sh scripts/scenarios_check.sh
+
+# The checker's own tests. Its value rests on its false-positive rate — it
+# reads prose, which is full of things shaped like paths that are not — so both
+# directions are asserted.
+test-scenarios-check:
```
tomo-chan, line 63: **`test-scenarios-check` never runs in CI**, so the new checker's own 15 assertions are enforced only by the local pre-push hook.

`.github/workflows/ci.yml` does not invoke `make test` or `make check` — it runs `make lint`, an inline `go test`, `npm run test:ci`, `make test-e2e`, `make coverage-go` and `make coverage-frontend`. The only place `scripts/scenarios_check*.sh` is exercised on a pull request is the new `scenarios.yml`, and that runs `make check-scenarios` (the checker) rather than `make test-scenarios-check` (the checker's tests).

Concretely: a change that breaks the false-positive handling — the `Cmd/Ctrl+Shift+B` / `/ws/board-command` / `bin/panemux` cases the PR description calls out as the gate's main risk — passes CI as long as the current ledger happens to still resolve, and is only caught if the author pushes with the hook installed. Adding a `- run: make test-scenarios-check` step to `scenarios.yml` (it needs no dependencies) closes it for a few seconds of job time.

---

docs/scenarios.md
```diff
@@ -91,6 +91,64 @@ Verification column values:
 | E7 | A real `claude` binary answers a real board question | The reply reflects actual board state | `manual`: enable `command_center`, start panemux with `claude` on PATH, ask "which panes are on the board?" and compare the answer against the dashboard |
 | E8 | Command center disabled | No palette, no history button, and `/ws/board-command` is not registered at all | `auto`: `internal/server/board_routes_test.go`; the UI half is covered by `frontend/src/App.test.tsx` |
 
+## H. The core multiplexer
```
tomo-chan, line 132: Sections **H** and **I** are inserted between **E** and **F**, so the document now reads A, B, C, D, E, H, I, F, G. In a ledger whose rows are referenced by letter-number ID from other docs and from code comments (`H9`, `I14`, `C7`), out-of-order section letters make "find row I3" a scan rather than a lookup, and the next contributor adding a section has no obvious place to put it.

Either move these two below `## G. The agmsg compatibility contract`, or relabel them `F`/`G` and renumber the two existing sections — moving is the smaller change since nothing yet cites `H`/`I` outside this PR.

---

tomo-chan (commented):
Reviewed the full diff. The direction is right and the ledger's test names all resolve exactly (I re-verified every `auto` row with a stricter exact-match script — no stale rows remain). Eight findings, mostly in the two new enforcement mechanisms rather than the ledger content.

The two I would not merge without: `scripts/scenarios_check.sh` **fails open** — it prints `ok` and exits `0` when its temp file cannot be created, having checked nothing (reproduced with `TMPDIR=/nonexistent-dir`) — and the `scenarios-exempt` label **cannot be applied to a failing run**, because `on: pull_request` has no `labeled` type and a re-run replays the old payload.

The rest: a race and a mis-directed assertion in the resize E2E, prefix-matching that lets an appended-suffix rename pass the gate, stderr being reported as a missing row, `echo` vs `printf` under `dash`, `test-scenarios-check` not running in CI, and section letters out of order in `scenarios.md`.

---

scripts/scenarios_check.sh
tomo-chan: Fixed in 010b9aa, taking the structural option: the temp file is gone and the loop reports through a pipeline.

That alone would still be trusting an empty capture, so the loop now ends by printing `__CHECKED__ <rows-seen> <tokens-resolved>` and the caller compares `rows-seen` against the row count it took before the loop started. A scan that stopped early, or never ran, can no longer be mistaken for a clean one — and the ok line now says `ok — 138 path and test names across 80 auto rows all resolve` rather than just `ok`, so the number is visible in the log.

Test added, exactly your reproduction: `TMPDIR=/nonexistent-dir` against a ledger naming a test that does not exist must exit 1. Run against the previous version of the script it fails, which is the point.

---

.github/workflows/scenarios.yml
tomo-chan: Applied in 010b9aa, with a comment recording why the type list is explicit — otherwise it reads like boilerplate and the next person trims it.

The same defect is in `efficacy.yml` on #185, which reads `efficacy-exempt` the same way. Fixing it there too rather than leaving a documented escape hatch that does not work in the other half of the same roadmap.

---

frontend/e2e/core-multiplexer.spec.ts
tomo-chan: Fixed in 010b9aa, doing both things you offered rather than either:

- A `splitSaved = layoutSaved(page)` subscribed *before* the split click, awaited right after the pane count settles. That drains the split's own PUT, so there is no in-flight response left to be mistaken for the drag's.
- `const saved = layoutSaved(page)` moved to just after `page.mouse.up()`, ~500 ms ahead of the debounced PUT.

The first is what makes the second safe: with the split's PUT already awaited, subscribing after `mouse.up()` cannot pick up a stale one no matter how slow `Config.SaveLayout` is.

Verified: the core-multiplexer project 3× alone (5/5 each) and the full `make test-e2e` once (28/28).

---

frontend/e2e/core-multiplexer.spec.ts
tomo-chan: Applied verbatim in 010b9aa — the bounds are swapped, so the poll is on `toBeLessThan(widthAfter + 30)` and the single-shot assertion is the lower bound.

You are right that the poll was doing no waiting at all: `widthBefore > widthAfter - 30` holds by construction, so it returned on its first sample in both the restored and the un-restored case. Left a comment saying so, since "poll the discriminating bound" is the non-obvious half — an assertion pair like this reads as symmetric and isn't.

---

scripts/scenarios_check.sh
tomo-chan: Applied verbatim in 010b9aa. The observation that renames overwhelmingly *append* is what makes this matter — the loose match was blind to the single most likely form of rot the gate exists to catch.

Two tests: `TestValidate_AgentBoardMode_InvalidValue` (no wildcard) must now fail, and `TestValidate_AgentBoardMode_InvalidValue*` must still resolve, so the wildcard branch is pinned alongside the exact one.

Worth recording: after the change all 138 tokens across the ledger's 80 auto rows still resolve, so no row was relying on the loose match.

---

scripts/scenarios_check.sh
tomo-chan: Both fixed in 010b9aa.

The `2>&1` is gone — stderr now goes to the terminal as diagnostics and only the loop's own findings reach the verdict. Test: a row naming `` `Test(bad` `` must still exit 1 with `X17 names Go test ...` on stdout, and stdout must *not* contain grep's `Unmatched (` — it fails against the previous version, which printed the error inside the FAIL block.

`find` now prunes rather than filters: `\( -name node_modules -o -name .git -o -name dist \) -prune -o -path "*/$target" -print -quit`. On this checkout with dependencies installed the whole check runs in ~0.8s.

---

scripts/scenarios_check.sh
tomo-chan: Applied in 010b9aa. Confirmed `/bin/sh` is dash in the environment this was developed in too, so the test discriminates rather than being taken on trust.

The fixture uses a literal `\n` in a cell rather than `\|`: expanded, it breaks one row into two, so the failure comes back naming an empty scenario instead of `X18` — and the new row-count check reports "read 2 of 1 auto rows" on top. Against the previous version it fails; against this one the finding names X18.

(One thing the fix does *not* buy, recorded so it isn't assumed: `awk -F'|'` still splits on the `|` of a `\|`, since the escape is a markdown convention awk knows nothing about. The fix stops the escape being *rewritten* before tokenising, which is the part that changed behaviour silently.)

---

Makefile
tomo-chan: Added in 010b9aa — a `make test-scenarios-check` step in `scenarios.yml`, ahead of `make check-scenarios`, with a comment recording why it is there (ci.yml runs neither `make test` nor `make check`, so the false-positive assertions had no CI enforcement at all).

The suite is now 20 checks rather than 15, and the false-positive cases you name are exactly the ones that had no CI coverage.

---

docs/scenarios.md
tomo-chan: Moved in 010b9aa — H and I now sit below G, so the document reads A through I in order. Took the move rather than the relabel for the reason you give: nothing outside this PR cites H or I yet, and relabelling would have invalidated every existing reference to F and G.

All 80 auto rows still resolve after the move.

---

scripts/scenarios_check.sh
```diff
@@ -0,0 +1,181 @@
...
+	exit 1
+fi
+
+seen=${checked%% *}
+resolved=${checked##* }
+
+if [ "$seen" != "$row_count" ]; then
+	echo "scenarios: read $seen of $row_count auto rows — the check did NOT run to completion."
+	exit 1
+fi
```
tomo-chan, line 169: **`__CHECKED__` guards that the loop *finished*, but not that it *resolved anything* — so a format drift in the Verification column still turns the gate green while checking nothing.**

`$resolved` is computed and printed, and then only ever used to make the ok line read nicer. Nothing asserts it is non-zero. Every row can be read, tokenised, and match no branch at all, and `seen == row_count` with empty findings is how this script says "everything resolved".

Concretely: `grep -o '`[^` ]* `'` is what produces `$tokens`, so a row only ever yields tokens for values written **in backticks**. A row selected by `grep '`auto'` always carries at least the `auto` token, which the first `case` skips — so if the column stops backticking test names and paths, every row contributes exactly one skipped token and zero findings.

Reproduced against this commit, with a ledger naming a test and a file that both do not exist:

```
| Z1 | x | `auto`: TestDoesNotExistAtAll |
| Z2 | y | `auto`: internal/nope/missing.go |

$ sh scripts/scenarios_check.sh /tmp/sample-ledger.md
  ok — 0 path and test names across 2 auto rows all resolve
EXIT=0
```

This is the same failure shape the rest of this commit closed, one level up: the empty-ledger case is guarded (line 40), the didn't-finish case is guarded (line 166), and the ran-but-recognised-nothing case is not. It is also the likeliest of the three to happen by accident, because it needs no infrastructure failure — only someone reformatting a column in a markdown file, which is exactly the kind of edit nobody runs a shell script's tests for.

The guard is one branch, next to the one already here:

```sh
if [ "$resolved" -eq 0 ]; then
        echo "scenarios: $seen auto rows, and not one of them named a path or a Go test."
        echo "           The Verification column's format must have changed — the check did NOT run."
        exit 1
fi
```

(tabs to match the file). A test alongside `a ledger with no auto rows fails rather than silently passing` would pin it — the fixture above fails against this commit and passes with the branch added.

---

tomo-chan (commented):
Re-reviewed `010b9aa` against the nine findings from the previous round. All nine are fixed, and I verified the four that make a behavioural claim rather than a textual one by running them:

- `TMPDIR=/nonexistent-dir` now completes normally (the temp file is gone entirely), and `__CHECKED__` + the row-count comparison replaces the trust that was in the empty capture.
- Exact test names no longer match by prefix: `TestValidate_AgentBoardMode_InvalidValue` fails, and its `*` form still resolves.
- `grep`'s own error on `` `Test(bad` `` no longer appears on stdout under the missing-row headline.
- `find` prunes rather than filters; the whole check is ~0.6s on a checkout with dependencies installed.

`make check-scenarios` resolves 138 tokens across 80 auto rows, `make test-scenarios-check` is 20/20, and the section letters read A through I.

One new finding, on the guard added in this commit rather than on anything from the last round: `__CHECKED__` proves the loop *finished*, but nothing asserts it *recognised* anything, so `resolved == 0` across every row still exits 0. That leaves the same fail-open shape this commit closed twice over reachable through a markdown formatting change alone. Details and a repro inline.

---

scripts/scenarios_check.sh
tomo-chan: Reproduced your ledger verbatim against `010b9aa` before changing anything — `ok — 0 path and test names across 2 auto rows all resolve`, exit 0 — and applied the guard in 4d4c81a.

Your framing is the part worth keeping: three fail-opens of the same shape, and this was the last one standing. The empty-ledger case guards "there were no rows", `__CHECKED__` guards "the loop stopped early", and nothing guarded "it read every row and recognised none of them". You're also right that it is the likeliest of the three, and for the reason that makes it nasty — it needs no infrastructure failure, just a markdown reformat, which is exactly the edit nobody runs a shell script's tests for.

One placement decision, since it is not arbitrary: the branch sits **after** the findings check, not before. A ledger whose rows all name nonexistent tests also resolves nothing, and there the format is fine and the missing names are the real answer — reporting "the format must have changed" would send the reader to the wrong file. Two fixtures now:

- `rows that resolve nothing at all fail rather than reporting all clear` — your ledger; fails against `010b9aa`, passes with the branch.
- `a missing name is reported as missing, not as a format change` — one row naming a real path and a renamed test; fails if the guard is ever moved ahead of the findings check.

Suite is 22 checks; `make check` green; the real ledger still reports 138 tokens across 80 rows.

---

PR #187, opened 2026-08-23
test: measure performance and accessibility, without gating (#180 item 7)

tomo-chan:
Roadmap item 7 of #180. Performance efficiency and interaction capability are the two ISO 25010 characteristics [docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md) records as unprotected. This is the first measurement of both.

- [x] ターミナル出力スループット / 多ペイン時の描画 / リレーのポーリング負荷を計測する手段を作る
- [x] E2E に a11y 検査を追加する
- [x] しばらくは**計測のみ**。閾値を決めるのは実データが溜まってから

## Nothing here asserts a threshold

Deliberately. The roadmap says measure now and choose thresholds from real data, and **design principle 4** is why: a number guessed at today is a number nobody trusts, and an untrusted gate gets routed around — taking the gates that do work with it. `make bench` is outside `make check`; the a11y spec asserts only that the scan ran.

## What `make bench` measures

| Package | Path | Why it is the one worth watching |
|---|---|---|
| `internal/session` | `publish` across chunk size × subscriber count | every byte a pane produces travels it |
| `internal/session` | `Subscribe` | what a workspace switch pays per remounted pane |
| `internal/board` | `AppendMessage`, `MessagesSince` | the relay polls every host every 5s for the life of the process |
| `internal/board` | `StatusSnapshot` | what `GET /api/board/status` copies per poll |

The publish benchmark starts with a **full** replay buffer on purpose. With an empty one the result depends on how far through `b.N` the buffer happened to fill, so the same code measures differently every run — the first version had exactly that problem. The cold path is a separate benchmark, and the comparison between them is the finding.

## The finding

4KB chunk, no subscribers. Medians of five runs, with the observed range:

| Replay buffer | ns/op (median) | range | B/op |
|---|---|---|---|
| Empty (cold) | 1,746 | 1,686 – 2,251 | 4,096 |
| Full (steady state) | 343,034 | 268,696 – 355,473 | ~598,000 |

**The gap — two orders of magnitude — is the trim.** `publish` retains a fixed 256KB replay window by reallocating and copying it on every chunk, so a pane open for more than a few seconds pays a full window copy per 4KB of output, forever. A ring buffer would make it constant.

**Deliberately not fixed here.** Item 7 is measurement; changing the replay buffer belongs with the reliability tests that cover replay (and the Alloy model), not with the benchmark that spotted it.

Related, and reassuring: subscriber fan-out is cheap by comparison — 16 subscribers add roughly 35% to a 4KB chunk — so many-pane cost is dominated by the per-pane buffer, not the fan-out.

## Accessibility baseline

`frontend/e2e/a11y.spec.ts` scans the dashboard and a modal dialog with axe-core and **records** what it finds: a log line in the run plus a full attachment for whoever fixes them.

`.xterm` is excluded — xterm.js owns that canvas, and auditing it would report on someone else's markup every run with nothing actionable here.

| Page state | critical | serious | moderate |
|---|---|---|---|
| Dashboard | 1 (`aria-required-children`) | 1 (`color-contrast`, 2 nodes) | 1 (`region`, 7 nodes) |
| Pane settings dialog open | 2 (`+ select-name`) | 1 (`color-contrast`, 10 nodes) | 1 (`region`, 7 nodes) |

Turning axe on over an existing UI produces a backlog, and a gate that starts red is bypassed on day one. The intended next step is to freeze these counts as a ceiling that may fall but not rise — a threshold chosen from data.

## Review round (221ab92)

Four findings, all reproduced here before being fixed. Two of them meant a benchmark measured something other than its label, which on a measure-only PR is the whole deliverable being wrong.

**`b.StopTimer`/`b.StartTimer` per iteration.** Each does a stop-the-world `runtime.ReadMemStats`. `BenchmarkSessionPublishColdBuffer` alone spent 50s of wall clock on ~1s of measured work — `sys` 52s against `real` 50s — and the repeated pauses reset the GC and allocator state the benchmark exists to measure, reporting **~5,400 ns/op for a path that costs ~1,750**. The reset is now inside the timed region (a single pointer store, cheaper to time than to exclude); 50s became 2.6s. `BenchmarkSessionSubscribe` now unsubscribes in the timed region and says so — collecting `b.N` unsubscribe funcs to drain later would need tens of gigabytes, since each live subscription holds a channel *and* a copy of the replay buffer.

**`AppendMessage/history=0` could not measure an empty history.** `b.N` is millions, so the first 2000 iterations fill the cache and the rest measure the full case — the same `b.N` confound this description credits itself with removing from `publish`, one file over. Demonstrated by the ordering flipping between runs (at `-benchtime 100000x` the "empty" row came out *slower*). Row dropped: there is also nothing for it to find, since the trim is a pointer bump and append's regrow is amortized alike.

**The recorded numbers did not reproduce**, by up to ~3×. Re-measured with `-count 5`; the section now quotes medians with ranges, because the spread turned out to be the finding — the board rows are steady to ~1.1× while the `publish` rows move by up to **2.9× between runs of the same binary on the same container**. So the doc now says these are indicative, and that choosing a threshold from a single run of them would build a gate firing on container noise — principle 4's own failure mode.

The old "~33×" headline came from the inflated cold figure. It is not replaced with "~125×": at these spreads the same two benchmarks support anything from ~120× to ~200×, so the claim is now "two orders of magnitude" plus an explicit note that the precision never existed. The trim finding survives the correction comfortably.

**`benchSession`'s comment** called it a `Session` "whose `Read` never blocks and never ends"; `Read` returns `io.EOF` on its first call. Nothing calls it today — the trap is the next benchmark to go through `Manager.Add`, which would take the EOF branch immediately and report a near-zero cost *rather than failing*, since a benchmark has no red state to fall into. The comment now says what the type is and what to change before using it that way.

## Docs

Every number is recorded in a new **"First measurements"** section of `docs/quality-gateway.md`, because with a measure-only change the numbers *are* the deliverable. The two ISO characteristic rows and rollout row 7 are updated; `docs/scenarios.md`'s "Not covered" explains why these get no scenario rows (there is no expected outcome to state yet).

## New dependency

`@axe-core/playwright` (devDependency). It is the standard implementation and is only loaded by the E2E suite; the alternative — hand-rolled assertions about accessible names — would be materially weaker for the same maintenance.

## Test plan

- [x] `make bench` — 23 benchmarks run; full suite now ~35s at default benchtime, ~3min at `-count 5` (was ~30min-scale before the `StopTimer` fix)
- [x] `make bench BENCH_ARGS='-count 5'` — medians and ranges as recorded above
- [x] Each of the four review findings reproduced before the fix and re-measured after
- [x] `make test-e2e` (default project) — 10/10 including the two a11y scans
- [x] `make lint` — 0 issues
- [x] `make check` — passes end to end (benchmarks are correctly outside it)

---


internal/session/manager_bench_test.go
```diff
@@ -0,0 +1,135 @@
...
+	entry := newBenchEntry()
+	chunk := make([]byte, 4096)
+
+	b.SetBytes(int64(len(chunk)))
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		entry.history = nil
+		b.StartTimer()
```
tomo-chan: **Per-iteration `StopTimer`/`StartTimer` dominates the run and inflates the cold-path number.**

`b.StopTimer()` and `b.StartTimer()` each call `runtime.ReadMemStats`, which is stop-the-world. Calling both every iteration (here and in `BenchmarkSessionSubscribe` below) means the wall-clock cost is the harness, not the code:

```
$ go test ./internal/session -run '^$' -bench 'BenchmarkSessionPublishColdBuffer|BenchmarkSessionSubscribe' -benchmem
BenchmarkSessionPublishColdBuffer-4        552516     1875 ns/op   4096 B/op   1 allocs/op
BenchmarkSessionSubscribe/buffered=0B-4    650745     1572 ns/op   1928 B/op   3 allocs/op
...
ok  panemux/internal/session  126.354s
real 2m7s   user 0m34s   sys 2m16s
```

~4s of measured work, 126s of wall clock, and `sys` time larger than `real` — that is ~95–135µs of `ReadMemStats` per iteration. The Makefile's own documented invocation, `make bench BENCH_ARGS='-benchtime 3s -count 5'`, extrapolates to ~30 minutes spent almost entirely in STW pauses. (`-timeout` doesn't kill it — `testing` stops the alarm before `runBenchmarks` — so it just runs a very long time.)

It also perturbs the result. Repeated STW pauses between iterations reset exactly the GC/allocator state this benchmark is measuring, and the cold path comes out ~1,875 ns/op rather than the **~5,400 ns/op** recorded in `docs/quality-gateway.md`. That matters because the headline of this PR is derived from it: against the 4KB/0-subscriber steady-state figure I measure (235,373 ns/op), the real gap is ~125×, not "~33×".

For this benchmark the fix is free — `entry.history = nil` is a single pointer store (~1ns, a constant offset on a ~2µs op), so it can just live inside the timed region:

```suggestion
	for i := 0; i < b.N; i++ {
		entry.history = nil
		entry.publish(chunk)
```

For `BenchmarkSessionSubscribe` the same applies: either unsubscribe inside the timed region (it is one map delete plus a channel close) or collect the unsubscribe funcs and drain them after the loop.

---

internal/board/cache_bench_test.go
```diff
@@ -0,0 +1,104 @@
...
+	return c
+}
+
+// BenchmarkBoardCacheAppendMessage is what one polled message costs the relay.
+// The full case is the one that matters: at the history limit every append
+// reslices, so a long-running panemux pays that on every message forever.
+func BenchmarkBoardCacheAppendMessage(b *testing.B) {
+	for _, prefilled := range []int{0, defaultBoardCacheHistoryLimit} {
+		b.Run(fmt.Sprintf("history=%d", prefilled), func(b *testing.B) {
+			c := filledCache(prefilled)
```
tomo-chan: **The `history=0` row doesn't measure an empty history — it's the same `b.N`-dependent confound this PR says it removed from the publish benchmark.**

`filledCache(0)` is built once, then `b.N` appends run into it. `defaultBoardCacheHistoryLimit` is 2000 and `b.N` at default benchtime is ~6.5M, so after the first 0.03% of iterations this is at the limit and measuring exactly what `history=2000` measures. The PR body says of `publish`: *"With an empty one the result depends on how far through `b.N` the buffer happened to fill, so the same code measures differently every run"* — that reasoning applies unchanged here.

Two runs on the 4-core Xeon container this PR names:

```
-benchtime 100000x
  history=0      266.4 ns/op
  history=2000   245.2 ns/op     <- "empty" measured SLOWER

default benchtime
  history=0      203.5 ns/op
  history=2000   258.9 ns/op
```

The ordering flips between runs, i.e. there's no signal. So the `~270` vs `~730` contrast in `docs/quality-gateway.md` (lines 284-285) and the claim that *"at the 2000-row limit every append reslices, so a long-running panemux pays that on every message forever"* aren't reproducible from this benchmark. (They also can't be: the reslice is `c.history = c.history[1:]`, a pointer bump; the only real cost is the amortized regrow every ~cap appends, which both rows pay identically.)

Either drop the `history=0` row, or restore the cache to its prefilled state per iteration — and if you do the latter, please avoid `StopTimer` for the reset (see the `manager_bench_test.go` comment).

---

docs/quality-gateway.md
```diff
@@ -249,6 +249,61 @@ block executed*; mutation reports *whether the tests would notice it changing*,
...
+
+### Relay polling cost
+
+| Operation | ns/op |
+|---|---|
+| `AppendMessage`, empty history | ~270 |
+| `AppendMessage`, at the 2000-row limit | ~730 |
+| `MessagesSince`, caught-up cursor | ~10,300 |
+| `MessagesSince`, cold start | ~400,000 |
+| `StatusSnapshot`, 64 panes | ~22,800 |
```
tomo-chan: **These numbers don't reproduce, and with a measure-only change the numbers are the deliverable.**

Re-running `make bench` on the same class of machine this section names (4-core Intel Xeon @ 2.80GHz container):

| Recorded | Measured |
|---|---|
| `AppendMessage`, at the 2000-row limit ~730 | 259 ns/op |
| `MessagesSince`, caught-up ~10,300 | 7,375 ns/op |
| `MessagesSince`, cold start ~400,000 | 532,019 ns/op |
| `StatusSnapshot`, 64 panes ~22,800 | 10,288 ns/op |
| publish, empty (cold) ~5,400 | 1,875 ns/op |
| publish, full (steady state) ~180,000 | 235,373 ns/op |

Up to ~3× off, in both directions. Some of this is genuine container noise, some is the `StopTimer` overhead flagged on `manager_bench_test.go`, and the `AppendMessage` row is measuring something other than what it's labelled (see that comment) — but from a reader's side it's indistinguishable, and the stated next step is *"freeze these counts as a ceiling"*, i.e. someone will run this to check the baseline before choosing a threshold from it.

Suggest either recording these with `-count` and a spread rather than a single run, or adding an explicit note that they're indicative single-run figures from a shared container and not comparable across machines. Otherwise the first person who reruns them concludes the baseline is wrong.

---

internal/session/manager_bench_test.go
```diff
@@ -0,0 +1,135 @@
...
+	chunk []byte
+}
+
+func (b *benchSession) ID() string    { return b.id }
+func (b *benchSession) Type() Type    { return TypeLocal }
+func (b *benchSession) Title() string { return b.id }
+func (b *benchSession) State() State  { return StateConnected }
+func (b *benchSession) Read(p []byte) (int, error) {
+	return copy(p, b.chunk), io.EOF
+}
```
tomo-chan: **The comment and the implementation disagree: this `Read` ends immediately.**

"a `Session` whose `Read` never blocks and never ends" — but `newBenchEntry` builds `&benchSession{id: "bench"}` with `chunk` left nil, so `Read` does `copy(p, nil)` → `(0, io.EOF)` on the first call.

Harmless today because nothing here starts a pump. The risk is that the comment explicitly frames this type as the never-ending PTY stand-in, so the next person adding a benchmark that goes through `Manager.Add` (to measure the pump path, which is the obvious gap) gets a session that hits the `err == io.EOF` branch on the first read, calls `closeSubscribers()`, and returns — a benchmark that measures nothing and reports a near-zero cost rather than failing.

Either give `chunk` a value and return `nil` instead of `io.EOF`, or trim the comment to what the type actually is: a stub that satisfies the interface so `managedSession` can be constructed, and whose I/O methods are never called.

---

tomo-chan (commented):
Reviewed the measure-only change (Makefile `bench` target, the two benchmark files, `a11y.spec.ts`, and the docs). The structure is right — `bench` is correctly outside `make check`, the a11y spec asserts nothing about counts, and the lockfile entry for `@axe-core/playwright` is complete. The `publish` steady-state benchmark's full-buffer setup is correct: `entry.history = make([]byte, sessionReplayLimitBytes)` sits exactly at the limit so every iteration takes the trim path, and the `598019 B/op / 2 allocs/op` I measure matches the "reallocate and copy the whole window per chunk" finding. That finding holds.

Four comments, all about the measurement rather than the mechanism. Two matter:

1. **`BenchmarkSessionPublishColdBuffer` / `BenchmarkSessionSubscribe` call `StopTimer`/`StartTimer` every iteration.** Both do a stop-the-world `ReadMemStats`, so those two benchmarks alone take 126s wall for ~4s of measured work (`sys` 2m16s > `real` 2m7s), and `make bench BENCH_ARGS='-benchtime 3s -count 5'` — the invocation the Makefile itself documents — extrapolates to ~30 minutes. It also inflates the cold number the "~33×" headline is computed from; without the interleaved pauses the cold path is ~1,875 ns/op, making the real gap ~125×.

2. **`BenchmarkBoardCacheAppendMessage/history=0` doesn't measure an empty history** — it fills to the 2000-row limit within the first 0.03% of `b.N` and then measures the same thing as `history=2000`. Reproduced twice with the ordering flipping between runs, so the recorded ~270 vs ~730 contrast isn't a real signal. This is the same `b.N`-dependent confound the PR body describes fixing in the publish benchmark.

Plus a note on the recorded numbers generally (several are ~2–3× off on rerun on the same machine class, in both directions) and a small comment/implementation mismatch on `benchSession.Read`.

Nothing here blocks the direction — the benchmarks measure the right paths and the a11y baseline is a real addition.

---

internal/session/manager_bench_test.go
tomo-chan: Reproduced before changing anything, and the effect is as large as you measured:

```
before: BenchmarkSessionPublishColdBuffer-4   554780   1919 ns/op
        ok  panemux/internal/session  50.146s
        real 0m50.7s   user 0m14.4s   sys 0m52.9s     <- sys > real

after:  BenchmarkSessionPublishColdBuffer-4   714477   1859 ns/op
        ok  panemux/internal/session  1.359s
        real 0m2.6s
```

~1s of measured work either way; 50s of wall clock became 2.6s. Applied your suggestion verbatim for the cold-buffer reset in 221ab92.

For `BenchmarkSessionSubscribe` I took the first of your two options rather than the second, and the reason is worth recording since the second looks cheaper on paper: each live subscription holds a 64-slot channel **and** a copy of the replay buffer, so collecting `b.N` unsubscribe funcs to drain later would need tens of gigabytes on the `buffered=262144B` row. Unsubscribing inside the timed region is now what the benchmark does, and both its comment and the doc rows say the figure covers the pair — which is also what a remounting pane actually pays.

You are right that this fed the headline. `~5,400 ns/op` was the STW-inflated cold figure; the honest median over five runs is 1,746. I have not replaced "~33×" with "~125×" though — at the spreads these benchmarks show (see the docs thread) the same two numbers support anything from ~120× to ~200×, so the section now says "two orders of magnitude" and states outright that the precision never existed. The trim finding itself survives comfortably.

---

internal/board/cache_bench_test.go
tomo-chan: Reproduced, including the ordering flip:

```
-benchtime 100000x   history=0  393.5 ns/op   history=2000  315.7 ns/op   <- "empty" slower
default benchtime    history=0  305.2 ns/op   history=2000  426.6 ns/op
```

Dropped the row in 221ab92 rather than trying to restore per iteration, because your parenthetical is the decisive half: there is nothing for it to find. The trim is a pointer bump and append's regrow is amortized alike, so even a correctly-held-empty cache would measure the same thing — restoring per iteration would buy a slower benchmark and the same answer.

The comment on `BenchmarkBoardCacheAppendMessage` now records both halves — why the row cannot work and why it would not matter if it did — since "why is there no empty case here" is exactly the question that gets one added back.

You caught this being the same confound the PR body claims credit for removing from `publish`. That is the part I would rather have found myself: the reasoning was written down, one file over, and not applied.

---

docs/quality-gateway.md
tomo-chan: Agreed, and the sentence that decided it: *with a measure-only change the numbers are the deliverable.* Re-measured with `-count 5` in 221ab92 and the section now quotes medians with the observed min–max.

The spread turned out to be the finding rather than a caveat on it:

| Benchmark | median | range | spread |
|---|---|---|---|
| publish 4KB/subs=0 (steady) | 343,034 | 268,696 – 355,473 | 1.32× |
| publish 256B/subs=0 | 245,012 | 102,390 – 297,239 | **2.90×** |
| publish cold | 1,746 | 1,686 – 2,251 | 1.34× |
| `AppendMessage` @ limit | 288 | 276 – 303 | 1.10× |
| `MessagesSince` caught-up | 9,171 | 9,102 – 9,258 | 1.02× |
| `MessagesSince` cold start | 602,421 | 567,966 – 625,917 | 1.10× |
| `StatusSnapshot` 64 panes | 9,873 | 9,575 – 10,654 | 1.11× |

The board rows are steady to ~1.1×; the `publish` rows move by up to 2.9× between runs of the same binary on the same container. So the doc now says these are indicative, that rerunning and getting a different number is not a regression, and — the part that actually matters for item 7's next step — that choosing a threshold from a single run of these would produce a gate firing on container noise, which is principle 4's own failure mode. A distribution on dedicated hardware is what that step needs.

Your `MessagesSince`/`StatusSnapshot` deltas were partly the `StopTimer` neighbours perturbing the whole binary; the remaining gap is inside the spread above. The Makefile's `bench` comment now recommends `-count` for anything you intend to read, and records the ~35s / ~3min timings so the `-benchtime 3s -count 5` example is no longer the only one offered.

---

internal/session/manager_bench_test.go
tomo-chan: Took your second option in 221ab92 — trimmed the comment to what the type is — rather than making `Read` genuinely endless. Giving `chunk` a value and returning `nil` would work for a pump benchmark, but with `chunk` still nil anywhere else it turns into `(0, nil)` forever, i.e. a busy loop at full CPU: a different trap in the same place, and speculative until something actually drives the pump.

So the comment now says it is a stub whose I/O methods are never called, and spells out the trap you identified rather than just avoiding the false claim:

> Read returns io.EOF on the first call, so this is NOT a stand-in for a never-ending PTY stream. A benchmark that went through Manager.Add — the obvious next one to write — would take the io.EOF branch immediately, call closeSubscribers, and report a near-zero cost rather than failing. Give chunk a value and return a nil error before using it that way.

"Reports a near-zero cost rather than failing" is why this was worth a comment at all — a benchmark has no red state to fall into, so a wrong one just looks fast.

---

Issue #164, opened 2026-08-09
ci: investigate gating on per-block ("branch-adjacent") coverage, not just statement %


tomo-chan:
## Background

While closing test-coverage gaps for PR #163 (Agent Board Phase 1), a manual analysis found 28 genuinely untested error/edge-case branches (`if err != nil { ... }` bodies, fallback paths, etc.) in `internal/board` and `internal/server/board.go` — none of which `make coverage-go`'s 80% *statement* coverage gate caught, because the surrounding function's happy path already pushed its statement % comfortably past 80%.

The analysis method: run `go test -coverprofile=coverage.out -covermode=count` across the coverage-gated packages, then parse the raw profile (`file:startLine.startCol,endLine.endCol numStmts count`), **summing counts per unique block** (a block can appear multiple times in a merged profile when several `go test` invocations share one `-coverpkg` list — naively checking each raw line undercounts/overcounts), and flagging every block whose summed count is `0`. That surfaced exactly the untested branches; all but 2 were then closed with targeted tests (see PR #163's coverage-gap test commit). The 2 left open are documented as intentionally untestable/low-value in that commit.

## The ask

Investigate whether this "sum-per-block, flag zero-count" technique can be formalized into a small script and wired into `make coverage-go` / CI as an additional gate — catching untested branches that a statement-% threshold alone misses — and if so, propose a concrete design.

## Why not just adopt existing branch-coverage tooling

- **Go's toolchain has no native branch/decision (C1) coverage.** `go test -cover` / `go tool cover` only ever report statement (block) coverage. A proposal for native branch coverage support exists ([golang/go#70306](https://github.com/golang/go/issues/70306), opened Nov 2024) but is still in the "Incoming" / undecided proposal stage — not something to depend on.
- **gobco** (the most visible third-party option, [rillig/gobco](https://github.com/rillig/gobco)) instruments condition/branch coverage, but its documented usage is single-package at a time, with no documented support for `-race` or this repo's multi-package `-coverpkg` setup (`internal/config`, `internal/api`, `internal/ws`, `internal/server`, `internal/board`). Adopting it would mean a second, differently-shaped coverage tool and CI step alongside the existing one, for a repo this size.

## Proposed direction (for the follow-up PR to evaluate, not prescribe)

Reuse `coverage.out` — already produced by the existing `make coverage-go` — and add a small script (Go or a Makefile-invoked script, matching this repo's existing tooling style) that:

[...]

PR #188, opened 2026-08-29
test: add per-block coverage reporting and a diff-scoped gate (#164)

tomo-chan:
`make coverage-go`'s 80% statement threshold cannot see an entire
`if err != nil { ... }` body that no test enters — the happy path around
it carries the function past the threshold on its own. Issue #164 found
28 such branches by hand in internal/board and internal/server/board.go
and asked whether the technique that found them could be formalized.

scripts/coverage_blocks.sh reads the coverage.out that `make coverage-go`
already produces, sums each block's execution count across the profile's
duplicate entries for it, and reports every block whose sum is zero.
The summing is the correctness argument: one `go test` over nine package
patterns with a shared -coverpkg list emits the same block once per test
binary, so reading the raw lines one at a time flags blocks that other
packages' tests do execute.

The gate is scoped to the diff. Measured on this commit the repository
has 275 unexecuted blocks out of 1801; a gate over all of them would
start red, which docs/quality-gateway.md principle 4 says is how a gate
gets routed around. Diff-scoped it starts green, needs no second
exclusion list, and is exempted narrowly with `//coverage:exempt <reason>`
or branch-wide with COVERAGE_BLOCKS_EXEMPT.

Claude-Session: https://claude.ai/code/session_01Xj4dGtuSt3V9PWKhp4bNhC


scripts/coverage_blocks.sh
```diff
@@ -0,0 +1,405 @@
...
+# ── Gate mode ─────────────────────────────────────────────────────────────────
+
+if ! git rev-parse --verify --quiet "$base" > /dev/null 2>&1; then
+	echo "coverage-blocks: ERROR — base ref '$base' does not exist."
+	echo "  Without it there is no diff to scope the gate to, so it cannot run."
+	echo "  In CI this usually means the checkout lost 'fetch-depth: 0'."
+	exit 1
+fi
+
+merge_base=$(git merge-base "$base" HEAD)
```
tomo-chan: **Fail-open when `git merge-base` fails** — the one shape this script says it will never take.

`merge_base` is captured without checking the exit status. When `git merge-base` fails it prints nothing, so `merge_base` is empty, `git diff --name-only "" HEAD` errors out (`fatal: ambiguous argument ''`) and prints no names, `changed` stays empty, and line 271's branch reports:

```
coverage-blocks: no Go implementation changed against <base> — nothing to check.
```

and exits **0**. A required check reports green having measured nothing — exactly what the header comment and the missing-profile / missing-base paths exist to rule out.

Reproduced on a scratch repo with an orphan base branch:

```
=== gate vs unrelated orphan branch 'other' ===
fatal: ambiguous argument '': unknown revision or path not in the working tree.
coverage-blocks: no Go implementation changed against other — nothing to check.
exit=0
```

The realistic trigger is not orphan branches but a clone whose history does not reach the merge base — the same `fetch-depth: 0` regression the base-ref check three lines above already anticipates. `rev-parse --verify` passes there (the remote-tracking ref exists), and only `merge-base` fails, so that guard does not cover this case.

```suggestion
merge_base=$(git merge-base "$base" HEAD 2> /dev/null) || merge_base=""
if [ -z "$merge_base" ]; then
	echo "coverage-blocks: ERROR — no merge base between '$base' and HEAD."
	echo "  Without it there is no diff to scope the gate to, so it cannot run."
	echo "  In CI this usually means the checkout lost 'fetch-depth: 0'."
	exit 1
fi
```

Worth a case in `coverage_blocks_test.sh` too — the suite covers the missing-base ref but not the missing merge base.

---

scripts/coverage_blocks.sh
```diff
@@ -0,0 +1,405 @@
...
+# reason recorded there"), and an exemption nobody has to justify is how an
+# allowlist turns into a place gaps go to be forgotten.
+: > "$tmp/kept"
+while IFS="$tab" read -r file start end stmts; do
+	src="$repo_root/$file"
+	marker=""
+	if [ -f "$src" ]; then
+		above=$((start - 1))
+		[ "$above" -lt 1 ] && above=$start
+		marker=$(sed -n "${above},${start}p" "$src" 2> /dev/null)
```
tomo-chan: **An exemption on one block silently waives the block that opens on the next line.**

The window is `start-1` .. `start`, and the `case`/`grep` then run over both lines joined. A marker written on block A's *opening* line is therefore also found by block B when B's opening line is `A_start + 1` — which is exactly what a nested `if`, an `else` on the following line, or a second statement inside a one-line body looks like.

Reproduced:

```go
func Do(a, b bool) int {
	if a { //coverage:exempt outer cannot happen   // line 4, block start 4
		if b {                                     // line 5, block start 5
			return 1
		}
	}
	return 0
}
```

Both blocks at count 0, both lines changed by the branch:

```
coverage-blocks: nothing to report — every measured block on a line this branch changed was executed (2 exempt).
exit=0
```

The inner block was never exempted by anyone and no reason was ever written for it, but it is waived and the output does not say which blocks the two exemptions applied to — so a reviewer reading the diff cannot see it either. That is the same "place gaps go to be forgotten" the comment directly above this hunk is guarding against.

The fix is to scope each half of the window to the block it belongs to: a marker on `start` counts only if it is on `start`, and a marker on `start - 1` counts only when `start - 1` is not itself the opening line of another block in `$tmp/zero`/the profile. Simplest version that closes the reproduction: take the `start - 1` line only when no block in the profile starts on it.

Either way, a test case for "marker on the opening line of the enclosing block does not exempt a nested block" would pin it.

---

tomo-chan (commented):
Code review of the per-block coverage gate (issue #164). I read every hunk, ran `scripts/coverage_blocks_test.sh` (39/39 pass), and exercised `coverage_blocks.sh` against scratch repositories to check the paths the suite does not cover.

The core of the script holds up: summing per unique block is correct, the diff scoping works on both the "block start was changed" and "a line in the middle was changed" shapes, the `--summary`-wins-over-base ordering is right and matters for the reason the comment says, and the unmeasured-directory reporting behaves as documented. The `.PHONY`, `make test` and CI wiring are consistent.

Two correctness findings, both reproduced, both in `coverage_blocks.sh`:

1. **Fail-open when `git merge-base` fails** (line 229) — the gate exits 0 having checked nothing. This is the failure mode the script's own header calls out, and the existing missing-base-ref guard does not cover it because `rev-parse --verify` succeeds in that case.
2. **The `//coverage:exempt` two-line window leaks onto the next block** (line 352) — an exemption on an outer `if`'s opening line silently waives a nested block starting on the line below, with no reason ever written for it.

Neither is a blocker for the design; both are worth fixing before this becomes a required check, since (1) makes the gate report green without measuring and (2) makes an exemption wider than what a reviewer sees in the diff.

Non-blocking observation: D8 records that three blocks in `internal/ws/board_command.go` flip between runs. Diff scoping removes that noise from the baseline, but not from a PR that touches those lines — that PR gets a flaky gate. Might be worth a sentence in D8 saying so.

---

scripts/coverage_blocks.sh
tomo-chan: Fixed in fac2c54, with your suggestion applied as written.

Reproduced in the suite first, as a case built the way you described — a base ref that exists but shares no history with HEAD, which passes `rev-parse --verify` and fails only at `merge-base`:

```
FAIL: a base ref with no shared history fails rather than passing quietly: wanted exit 1, got 0
      fatal: ambiguous argument '': unknown revision or path not in the working tree.
```

It now exits 1 with the message your suggestion carries, and reverting the guard turns that case (and its message assertion) red again.

Your point about the realistic trigger is the part I had missed: I wrote the `rev-parse --verify` check for the lost `fetch-depth: 0` case and then assumed it covered it. It does not — the remote-tracking ref resolves in a shallow clone and only `merge-base` fails, so the exact regression that check exists for was landing in the fail-open path instead.

---

scripts/coverage_blocks.sh
tomo-chan: Fixed in fac2c54, both halves of it.

**Scoping.** The window is now the block's own opening line always, plus the line above *only when no other block in the profile opens there* — the version you called the simplest that closes the reproduction. The profile already carries every block's start, so this is one extra pass over it and a `grep -qxF` per finding, no new source parsing.

Your reproduction is a test case now, and it fails against the old window:

```
FAIL: a marker on the enclosing block's opening line does not exempt a nested block: wanted exit 1, got 0
      coverage-blocks: nothing to report — ... (2 exempt).
```

Verified on real code too: the same nested shape in `internal/portforward` with the marker on the outer `if` now reports

```
coverage-blocks: 3 block(s) this branch changed never executed (1 exempt).

  Exempt by //coverage:exempt:
    internal/portforward/registry.go:88-89

  internal/portforward/registry.go:87-88  1 stmt(s)
  internal/portforward/registry.go:89-91  1 stmt(s)
```

**Visibility.** "the output does not say which blocks the two exemptions applied to" was the more useful half of the finding, since it is what would have kept the bug invisible after a fix as well. Both exit paths now list the exempted blocks, not just the count.

One thing that came out of writing the assertions: the listing makes a bare `file:line` substring check unable to tell "reported" from "exempted", because both print the line. The nested case now asserts the failure headline instead.

---

scripts/coverage_blocks.sh
```diff
@@ -0,0 +1,456 @@
...
+if [ -z "$changed" ]; then
+	echo "coverage-blocks: no Go implementation changed against $base — nothing to check."
+	exit 0
+fi
+
+# touched_lines <file> — the line numbers this branch added or modified, against
+# the file as it stands now. -U0 keeps the hunk headers exact so no untouched
+# neighbour is swept in. Same helper, same reasoning, as scripts/efficacy.sh.
+touched_lines() {
+	git diff -U0 "$merge_base" HEAD -- "$1" |
```
tomo-chan: **A third fail-open, same family as the two just fixed: the pathspec here resolves against the *current directory*, not the repository root.**

`git diff --name-only` on line 297 prints paths relative to the repository root regardless of where it runs (`diff.relative` is off by default), so `$1` arrives here as e.g. `pkg/a.go`. But a pathspec after `--` is resolved against the caller's cwd. Run the script from a subdirectory and every pathspec misses, `$tmp/touched` comes back empty for every file, and line 308's `[ -s "$tmp/touched" ] || continue` skips them all silently — so the gate prints "nothing to report" and exits **0** having measured nothing.

Reproduced on a scratch repo with one uncovered block on a changed line, same repo and same profile both times:

```
--- from the repository root ---
coverage-blocks: 1 block(s) this branch changed never executed.
  pkg/a.go:4-6  1 stmt(s)
exit=1

--- from pkg/ ---
coverage-blocks: nothing to report — every measured block on a line this branch changed was executed.
exit=0
```

Isolating it, run from `pkg/`:

```
$ git diff --name-only main HEAD
pkg/a.go                       <- repository-relative

$ git diff -U0 main HEAD -- pkg/a.go
                               <- empty: resolved as pkg/pkg/a.go

$ git diff -U0 main HEAD -- ":/pkg/a.go"
diff --git a/pkg/a.go b/pkg/a.go
```

**Not currently breaking the gate**: `coverage-blocks.yml` and `make` both run from the repository root, so CI is unaffected. It is reachable through the invocation this script's own header documents on line 33 (`sh scripts/coverage_blocks.sh --base origin/main --profile coverage.out`) — nothing there says it must be run from the root, and `repo_root` on line 104 shows the script otherwise intends not to care.

Either fix works; the first also covers any other pathspec added later:

```suggestion
	git -C "$repo_root" diff -U0 "$merge_base" HEAD -- "$1" |
```

or `-- ":/$1"`. A test case would need `expect_status` to run from a subdirectory of the fixture repo rather than its root, which is a small change to the helper.

---

tomo-chan (commented):
Re-reviewed `fac2c54`. **Both findings from the previous review are genuinely fixed** — I verified each by running the pre-fix script (`9537c5f`) and the current one against the same scratch repository and the same profile, rather than reading the diff alone.

**1. `git merge-base` fail-open** — fixed.

| | base ref that exists but shares no history with HEAD |
|---|---|
| `9537c5f` | `fatal: ambiguous argument ''` → `no Go implementation changed — nothing to check.` → **exit 0** |
| `fac2c54` | `ERROR — no merge base between 'unrelated' and HEAD` + the `fetch-depth: 0` hint → **exit 1** |

**2. `//coverage:exempt` leaking onto the next line's block** — fixed. Same nested `if a { //coverage:exempt … }` / `if b {` fixture, both blocks at count 0, both lines changed:

| | |
|---|---|
| `9537c5f` | `nothing to report — … (2 exempt).` → **exit 0** |
| `fac2c54` | `1 block(s) this branch changed never executed (1 exempt).` → **exit 1**, with `Exempt by //coverage:exempt: pkg/a.go:4-5` naming the one block actually waived |

Scoping the `start - 1` line to "no other block in the profile opens there" is the right shape, and building `$tmp/block_starts` from the profile keeps it to one extra pass with no new source parsing. Listing the exemptions on both exit paths addresses the half of that finding that mattered more — a reviewer can now see what was waived without re-deriving it from the diff.

`scripts/coverage_blocks_test.sh` passes **45/45**, up from 39, and both new cases go red when their guard is reverted. All 12 CI checks are green on this head.

**One new finding**, filed inline: `touched_lines` passes a repository-relative path as a git pathspec, which git resolves against the caller's cwd, so running the script from a subdirectory reports green having measured nothing. CI is not affected (it runs from the repository root), but it is the same fail-open family as the two above and it is reachable through the direct invocation this script's header documents.

---

scripts/coverage_blocks.sh
tomo-chan: Fixed in 7e4c6d9, with the `git -C "$repo_root"` form — the one that also covers any pathspec added later.

Reproduced in the suite first. The suite could not see this class of bug at all, because every case ran the checker from the fixture repository's root, so it gained an `expect_status_in <subdir>` helper alongside `expect_status`:

```
FAIL: the same finding is reported when run from a subdirectory: wanted exit 1, got 0
      coverage-blocks: nothing to report — every measured block on a line this branch changed was executed.
```

47 checks now, and removing the `-C` turns that case and its message assertion red again.

Verified on this repository too, against `origin/main~12` (24 changed Go implementation files, so there is something real to find). From the root and from `internal/board`, with the same profile:

```
coverage-blocks: 29 block(s) this branch changed never executed.

  Not measured — these changed files are in packages the coverage profile
  does not cover, so this gate says nothing about them:
    internal/session/browseropen.go
    internal/session/local.go
```

byte-identical. Before the fix the second of those printed "nothing to report" and exited 0.

Two things I'll note, since this is the third of the same family. The pathspec is now the only one in the script, and it is the only place that depended on cwd — `git diff --name-only`, `merge-base` and `rev-parse` are all cwd-independent for what they return here, and every source read already goes through `$repo_root`. And your framing is the part I keep getting wrong on my own: I wrote `repo_root` precisely so the script would not care where it runs, then used a pathspec that made it care. All three findings have been "the guard exists, and the code path around it does not go through it."

---

PR #196, opened 2026-08-30
test: pin the boundary values mutation testing found unasserted (#190)

### LINKED ISSUE #190: test: 境界値テストを補強する（mutation 測定が見つけた 45 件）
#180 の項目 6 で入れた mutation 測定（[#189](https://github.com/tomo-chan/panemux/pull/189)、詳細は [#180 のコメント](https://github.com/tomo-chan/panemux/issues/180#issuecomment-5465858769)）が、生存 mutant 108 件のうち **45 件は境界値**だと特定した。その 45 件を潰す。

これは DEVELOPMENT.md の「Test granularity」が**すでに文章で要求している**内容で、L0（文書）に留まったまま誰も強制していない規律にあたる。

> Boundary variants: minimum/maximum values, zero/negative values, size sums, single item vs multiple items, nested structures, and paths with `~/` when paths are involved.

## 一貫したパターン

リポジトリ全体で **「境界の 1 つ外側は試すが、境界そのものは試さない」**。ポート範囲がいちばんわかりやすい。同じ形のチェックが 3 パッケージにあり、3 つとも生存した:

```go
internal/config/validate.go:35   if c.Server.Port < 1 || c.Server.Port > 65535 {
internal/api/handler.go:600      if req.Port < 0 || req.Port > 65535 {
internal/session/loopback.go:35  if port < 1 || port > 65535 {
```

テストが使っている値:

```
internal/portforward/registry_test.go:233  []int{0, -1, 80, 1023, 65536, 99999}
internal/session/loopback_test.go:228      []int{0, -1, 65536}
```

`65536` はあるが `65535` が無い。`> 65535` を `>= 65535` に変えると**有効なポート 65535 が拒否される**が、`65536` で試している限りどちらの版も拒否するのでテストは気づかない。手で確認済み — `validate.go:35` を書き換えて `go test ./internal/config/` は `ok`。

同じ死角が 3 パッケージに、別々の時期に書かれて存在している。1 つのテストの書き漏らしではなく**習慣**なので、まとめて直す価値がある。

## 対象（測定で生存が確認できているもの）

| 箇所 | 境界 |
|---|---|
| `internal/config/validate.go:35` | `port` の `1` と `65535` |
| `internal/api/handler.go:600` | `req.Port` の `0` と `65535` |
| `internal/session/loopback.go:35` | `port` の `1` と `65535` |
| `internal/config/validate.go:180` | `child.Size <= 0` のちょうど `0` |
| `internal/config/validate.go:185` | 合計の `99.9` と `100.1` |
| `internal/portforward/callback.go:132` | `minForwardablePort` / `maxPort` の両端 |
| `internal/portforward/registry.go:103,112,262` | `ttl <= 0` / `SweepInterval >= 0` / `active > 0` |
| `internal/board/cache.go:97` | `overflow > 0`（ちょうど `maxHistory`） |
| `internal/session/manager.go:148` | リプレイ上限ちょうど |
| `internal/api/handler.go:1276`, `internal/board/relay.go:72`, `internal/session/ssh.go:258` | `timeout <= 0` |
| `internal/session/local.go:770`, `internal/session/ssh.go:979` | `pid <= 0` |

残りは `internal/session` の PID 比較まわりなど。全 45 件の一覧は `MUTATION_BASE=origin/main make mutation` か、全体スキャン（`gremlins unleash --timeout-coefficient 10 --workers 2 .`）で再現できる。

## やらないこと

生存 mutant 108 件のうち **B（チューニング定数 12 件）と C（fault injection が必要なエラー分岐 25 件）には手を出さない**。`scanner.Buffer(make([]byte, 0, 64*1024), 16*1024*1024)` を殺すテストはバッファサイズを固定するだけの tautology で、G4 が探している欠陥そのものになる。必要なら `//mutation:exempt ` を付ける。

## 副次的な価値

**`make efficacy`（G4(b)）と `make mutation`（G4(c)）は、実 PR でまだ一度も判定していない。** #185・#186・#187・#189 はいずれも Go 実装を含まないため、両ゲートとも正しく skip し続けている（`no implementation changed` / `no Go implementation changed`）。失敗経路の担保は hermetic なテスト（29 件 / 16 件）だけ。

この issue は**実装とテストを同時に変える**ので、両ゲートの初の実地検証を兼ねる。#180 の項目 6 第 4 段階（mutation をゲート化するか）の判断材料も、ここで最初の 1 件が取れる。

## 完了条件

- [x] 上表の境界値がテストで突かれている（境界そのもの、および必要なら境界±1）
- [x] `MUTATION_BASE=origin/main make mutation` が、この差分の変更行について生存者を報告しない（または報告分に `//mutation:exempt ` の判断が付いている）
- [x] `make check` が通る

**完了** — [#196](https://github.com/tomo-chan/panemux/pull/196)（`6b1d479`）。結果と、当初の見立てと違った 3 点は [このコメント](https://github.com/tomo-chan/panemux/issues/190#issuecomment-5470015526)。要約すると: 45 件のうち 5 件はテスト漏れではなく殺せない生存者だった（第 4 段階の判断材料）／「equivalent」と「unreachable」を混同していた 1 件が実欠陥を隠していた／期待していた「両ゲートの初の実地検証」は実際に両方から不具合を出した。

## 関連

- #180 — 品質ゲートウェイ導入ロードマップ（項目 6）
- #189 — mutation ゲート（この issue の測定元）
- [docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md) の「Surviving mutants: the first measurement」／「Clearing the boundary-value class (issue #190)」

tomo-chan:
Closes #190.

The mutation measurement in #189 identified 45 surviving mutants as boundary values: the same habit in package after package of testing one port past the end of a range and never the end itself. `port > 65535` could become `>= 65535` in internal/config, internal/api and internal/session alike, rejecting a legal port, with every test still green.

Every site #190 named is now exercised at the boundary and at both neighbors — the three port-range checks, `child.Size <= 0` and the ±0.1 sum tolerance, `forwardablePort`'s 1024/65535 ends, the port-forward registry's TTL default, the board cache's history bound, the replay buffer's 256 KiB bound, the three `timeout <= 0` fallbacks and `remoteShellPID`'s `pid <= 0`. Each was confirmed by reverting the comparison by hand and watching the suite go red; a new test that passes proves nothing on its own here.

## What the sweep found

**Not every survivor is a missing test, and the two ways that happens are not interchangeable.** An *equivalent* mutant computes the same thing on every possible input. An *unreachable* one is killable, but only by input the code's own callers cannot produce. Five carry a `//mutation:exempt` naming which they are — four equivalent, one unreachable.

**A sixth was filed as equivalent and was neither**, caught in review. `endConn`'s `active > 0` was exempted on the grounds that `endConn` is only ever called paired with `beginConn` — a correct statement about *reachability* that says nothing about equivalence. `>= 0` drives the counter to −1 on an unpaired call, `reapExpired`'s own `f.active == 0` then never matches again, and the forward outlives its TTL for the life of the process:

```
shipped  `> 0`  → active=0,  ports after reap=[]
mutant  `>= 0`  → active=-1, ports after reap=[36685]
```

That is a real defect behind a load-bearing guard, so it has a test (`TestRegistryEndConnNeverDrivesTheCounterNegative`) and no exemption. The check that separates the two — *if the mutant were reachable, would anything be wrong?* — is now written into DEVELOPMENT.md, because the miss is more useful to stage 4 than the count is.

**One implementation change from the sweep itself.** `Options.SweepInterval`'s disabled/defaulted boundary sits exactly at zero and nothing observable separates the two from outside `New` — a reaper that ticks once a minute cannot be waited on in a test — so the decision moves into the pure `resolveSweepInterval`, the same split `browserOpenArgv` has from `openChrome`.

## The two gates' first real verdicts

Both G4(b) and G4(c) had never judged a Go diff; #185–#189 all skipped correctly. This branch gave them work, and both had something to say.

**`make efficacy` — 15 of 18 changed tests survived the revert, and every one legitimately.** A branch whose purpose is pinning behavior that is *already correct* cannot make those tests go red by reverting an implementation it never changed. Each carries a per-test `//efficacy:exempt` with that reason, rather than the branch-wide label. The three tests around `resolveSweepInterval` go red as they should.

**`make mutation` could not run in CI at all.** gremlins gathers coverage over the whole module before mutating, so the root package must compile and `main.go` embeds `frontend/dist` — the job died on `pattern frontend/dist: no matching files found`. `make mutation` now depends on `build-frontend` exactly as `make coverage-go` does, and mutation.yml gains the Node setup coverage-blocks.yml already carried. It survived #189 because every earlier pull request exited at "no Go implementation changed" first.

## Two gate fixes the review turned up

**`//efficacy:exempt` did not require its reason.** Both sibling markers refuse a bare one (`coverage_blocks.sh`, `mutation.sh`); `efficacy.sh` matched the bare substring in both places it looks, so a marker with no reason silently took a test out of scope — and this is the branch promoting that marker as the first tool to reach for. The `bare_marker` treatment is ported to both the Go and the frontend mapper: a bare marker waives nothing, the test stays in scope, and the run names it. A note rather than an error, because a bare marker is not itself a failing branch — the test under it is still judged on its own merits, which is the verdict the gate exists for.

**The mutation marker waives by line, not by mutant.** `scripts/mutation.sh` matches file and line and never the mutant `$type`, so a reason written about the boundary mutant covers every other mutant on that line. Measured across all five exempt sites: each one's negation mutant is killed by the suite independently, so nothing is hidden today — including `endConn`, whose negation mutant was killed the whole time while the boundary mutant its reason named was the one surviving. Structural gap, not yet load-bearing; recorded in docs/quality-gateway.md and DEVELOPMENT.md, and it starts mattering at stage 4, when a survivor becomes a failure.

## Test plan

- [x] `make check`
- [x] `make efficacy` — ok, all 3 non-exempt changed tests red
- [x] `COVERAGE_BLOCKS_BASE=origin/main make coverage-blocks` — nothing to report
- [x] `MUTATION_BASE=origin/main make mutation` — no survivors on changed lines
- [x] `sh scripts/efficacy_test.sh` — all 31 checks pass, including two new ones written first and confirmed failing
- [x] CI failure reproduced locally (`rm -rf frontend/dist`) and confirmed fixed
- [x] Every boundary mutant hand-applied and confirmed killed; every exemption hand-applied and confirmed to survive for the reason its marker gives
- [x] `internal/board` green under `-race -count=5` after the InDelta change, and still kills `opTimeout < 0`


internal/session/local.go
```diff
@@ -402,6 +402,10 @@ func newestKnownAgentTypeDescendantPID(processes []processInfo, rootPID int) (pi
 		stack = append(stack, children[proc.PID]...)
 
 		if t, matched := detectAgmsgAgentType(proc.Command); matched {
+			// The `>` boundary cannot be pinned by a test: two processes in
+			// one ps snapshot never share a PID, so `>=` can only differ on
+			// input the OS cannot produce. Issue #190.
+			//mutation:exempt equivalent — PIDs are unique within one snapshot, so >= never sees an equal pair
```
tomo-chan: This mutant (and its twin in `newestMatchingDescendantPID`) is **not equivalent** — a test can kill it, which matters because docs/quality-gateway.md counts both toward the "at least six of them cannot be killed by anyone" floor it offers as the evidence for roadmap item 6's stage 4.

`newestKnownAgentTypeDescendantPID` is a pure function taking `[]processInfo` as a parameter, so its test surface is a fabricated slice, not a real `ps` snapshot. Verified by hand-applying the mutant:

```go
procs := []processInfo{
    {PID: 5, PPID: 1, Command: "claude"},
    {PID: 5, PPID: 1, Command: "codex"},
}
newestKnownAgentTypeDescendantPID(procs, 1)
// shipped `>`  -> type "codex"
// mutant  `>=` -> type "claude-code"
```

The reason given ("PIDs are unique within one snapshot") is an argument about what the OS can produce, which is the right argument for *reachability* but not for *equivalence* — and the `//mutation:exempt` reasons plus the docs table row both say "equivalent … no test can kill them".

Two ways to close it, either is fine:
- reword both markers and the `internal/session/local.go (two newest-PID scans)` row in docs/quality-gateway.md from "equivalent" to "unreachable with real `ps` output", and drop these two from the six-mutant floor (it becomes four); or
- add the duplicate-PID case as a test and drop the exemptions entirely.

Everything else on this branch checks out — I hand-applied each boundary mutant the PR claims to pin (`config` port + `child.Size` + the ±0.1 sum tolerance, `forwardablePort`'s two ends, the API port check, `loopback`'s range, the registry TTL default, the board cache bound, the replay-buffer bound, the three `timeout <= 0` fallbacks, `remoteShellPID`) and every one goes red. `resolveSweepInterval` is behaviour-identical to the inlined version, and `make efficacy` passes with 3 tests going red.

---

tomo-chan (commented):
Reviewed the full diff for correctness. One finding, low severity, on a classification claim rather than runtime behaviour.

I verified the PR's central claim rather than taking it on trust: for every boundary this branch says it now pins, I hand-applied the mutant to the shipped source and re-ran the package. All of them go red — the three port-range checks (`internal/config`, `internal/api`, `internal/session/loopback.go`), `child.Size <= 0`, both ends of the ±0.1 sum tolerance, `forwardablePort`'s 1024/65535 ends, the registry TTL default, the board cache's history bound, the replay buffer's 256 KiB bound, the three `timeout <= 0` fallbacks (`prLookupTimeout`, relay `opTimeout`, `dialTransport`) and `remoteShellPID`'s `pid <= 0`. `resolveSweepInterval` is behaviour-identical to the code it replaced, `gofmt -s`/`go vet` are clean, the packages pass under `-race -count=5`, and `EFFICACY_BASE=origin/main scripts/efficacy.sh` passes with 3 tests going red.

Worth noting the float64 detail the tolerance test depends on, since it looks fragile but isn't: `49.95 + 49.95` and `50.05 + 50.05` compare exactly equal to the literals `99.9` and `100.1`, so those two cases really are sitting on the boundary and really do kill the `<=`/`>=` mutants.

---

internal/session/local.go
tomo-chan: Right on the substance, and it turned out to be worse than reported. Fixed in 2f739a0 — I ran your snippet, and two more besides.

**`newestKnownAgentTypeDescendantPID` — you're right.** Reproduced exactly as you wrote it: shipped `>` returns `"codex"`, mutant `>=` returns `"claude-code"`. Killable. Taken the first of your two options: the marker now says `unreachable` rather than `equivalent`, because the only assertion available is "on a duplicate PID, whichever type the traversal reaches first wins" — an accident of stack order, not designed behaviour, and pinning it would freeze the accident.

**The twin is the opposite case, and it stays `equivalent`.** `newestMatchingDescendantPID` returns `(int, bool)` and nothing else, so on an equal PID the guard's only writes are `matched = proc.PID` (the value already there) and `ok = true` (already true). Same fabricated slice, both versions return `5 true`. There is no second value for `>=` to change — the reason it survives is not reachability at all, and its marker now says so.

**And a third you didn't flag, which is the one that mattered.** Applying your own test — *is this equivalence or reachability?* — to the rest of the table caught `endConn`'s `active > 0`, which I had also filed as equivalent on a reachability argument. It is neither:

```
shipped  `> 0`  → active=0,  ports after reap=[]
mutant  `>= 0`  → active=-1, ports after reap=[36685]
```

`reapExpired`'s own `f.active == 0` stops matching at −1, so the forward outlives its TTL for the life of the process. That is a real defect behind a load-bearing guard, so it gets a test (`TestRegistryEndConnNeverDrivesTheCounterNegative`) and no exemption.

Floor is now four equivalent + one unreachable, and the docs table carries a `Kind` column plus a paragraph on the `endConn` miss, since "a reviewer who accepts an exemption's reason without asking which kind it claims will wave through real defects" is the more useful finding than the count. DEVELOPMENT.md gains the check itself: *if the mutant were reachable, would anything be wrong?*

Separately, the red `mutation` job: gremlins gathers coverage over the whole module before mutating, so the root package must compile and `main.go` embeds `frontend/dist`. `make mutation` now depends on `build-frontend` like `make coverage-go` does, and mutation.yml gains the Node setup coverage-blocks.yml already had. It survived #189 because every earlier PR exited at "no Go implementation changed" before gremlins ran — this branch is the first to give the gate real work, which is roughly what the issue hoped it would find out.

---

DEVELOPMENT.md
```diff
@@ -126,7 +128,8 @@ Example: `Config.sshConfigPath` uses `sshconfig.DefaultPath()` only when the ove
 - Run it yourself with `make efficacy` (compares against `origin/main`; override with `EFFICACY_BASE`). CI runs it on every pull request.
 - It is deliberately **not** part of `make check` — it needs the base branch and a second test run, which is a per-pull-request cost, not a per-turn one.
 - A branch that changes no implementation, or changes implementation but no test, is skipped rather than failed.
-- If a change genuinely should not go red without its implementation — a pure refactor, a test-only rename — apply the `efficacy-exempt` label to the pull request and say why in the description. That is the only exemption; do not reach for it to get past a test that turned out not to assert anything.
+- If a change genuinely should not go red without its implementation — a pure refactor, a test-only rename, a test pinning behavior this branch never touched — mark that test `//efficacy:exempt <reason>` in the comment directly above it. **The reason is required**, and the marker is the one to reach for first: it is per-test and it sits in the diff a reviewer reads. The `efficacy-exempt` label is the blunter tool and takes the whole branch out of scope.
```
tomo-chan: **The reason is required** is not true of `//efficacy:exempt` — unlike the two sibling markers, `scripts/efficacy.sh` never checks for one.

Both other markers enforce it:

- `scripts/coverage_blocks.sh:405` — `grep -Eq '//coverage:exempt[[:space:]]+[^[:space:]]'`, and a bare marker sets `bare_marker=1` and is reported as a finding rather than waived.
- `scripts/mutation.sh:370` — the same regex, same fall-through.

`scripts/efficacy.sh` matches the bare substring in both places it looks (`:160` for `--changed-tests`, `:241` for the survivor scan):

```awk
if (line[j] ~ /efficacy:exempt/) exempt = 1
```

So `// efficacy:exempt` with nothing after it silently takes the test out of scope. That is the exact failure mode the other two scripts' own comments call out ("an exemption nobody has to justify is how an exemption stops being reviewable"), and it is the marker this PR is introducing as the one to reach for *first*, so it is worth having it hold.

Either enforce it in `efficacy.sh` — the `bare_marker` treatment from `mutation.sh:366-378` ports over directly, and `scripts/efficacy.sh:715/765/792`'s own guidance text already tells the reader to say why — or drop the bold claim from this bullet.

---

internal/board/relay_test.go
```diff
@@ -430,6 +430,101 @@ func TestRelay_Poll_HostOperationTimeout_DoesNotBlockOtherHosts(t *testing.T) {
...
+				BackfillLimit:    1000,
+				OperationTimeout: tt.configured,
+			})
+
+			require.NoError(t, r.Poll(context.Background()))
+			assert.Len(t, cache.MessagesSince(0), 1)
+			assert.Equal(
+				t, hostOperationTimeout, host.observedBudget(),
+				"the per-host call must get the default budget, not a zero-length one",
+			)
```
tomo-chan, line 482: This asserts a wall-clock measurement for **exact** equality, so it can fail with no defect present.

`record` stores `time.Until(deadline).Round(time.Second)`, and the deadline was set by `Poll`'s own `context.WithTimeout(..., hostOperationTimeout)` a moment earlier. `Duration.Round` is round-to-nearest, so anything that eats more than 500 ms between those two points — a loaded CI runner, a GC pause, `-race` scheduling — rounds 9.4s down to `9s` and this fails with `expected: 10s, actual: 9s`. The margin is generous in practice, but it is a real timing dependence in a test that otherwise has none, and the failure it produces reads as a defect in the timeout plumbing rather than as a stall.

An assertion that keeps the property the comment says it is after ("which budget was chosen, not nanosecond accuracy") without the cliff:

```go
assert.InDelta(
    t, hostOperationTimeout, host.observedBudget(), float64(time.Second),
    "the per-host call must get the default budget, not a zero-length one",
)
```

with `record` storing `remaining` unrounded. That still separates 10s from the 0s this test exists to rule out — the two are ten seconds apart — while tolerating a stall of any size below a second in either direction rather than only below half of one.

Separately, the helper's doc comment at :509 says it "rounds the remaining budget **up** to whole seconds". `Round` does not round up; `Truncate` rounds down and neither rounds up. If the delta form above is taken the sentence goes away with it, but the claim is worth not leaving in place either way.

---

tomo-chan (commented):
Re-reviewed 2f739a0. The three corrections hold up under measurement, and one of them corrects *me*.

**Verified rather than read:**

- `TestRegistryEndConnNeverDrivesTheCounterNegative` kills the mutant it exists for — hand-applying `f.active >= 0` gives `active after an unpaired endConn = -1, want it held at 0`, green unmutated, stable under `-race -count=10`.
- `newestMatchingDescendantPID` really is equivalent and my original comment was wrong to sweep it in with its twin: the guard's only writes are `matched = proc.PID` and `ok = true`, and there is no third value, so `>=` cannot differ on any input. Keeping it `equivalent` while the other becomes `unreachable` is the right split.
- The `mutation` job now reaches gremlins for the first time (`mutation: no surviving mutants on lines this branch changed`). The `internal/portforward/registry.go:133 CONDITIONALS_NEGATION` line under "Not analysed" is gremlins' own SKIPPED, not a survivor — applying `configured != 0` by hand does fail the suite.
- The other five `//mutation:exempt` sites all still have their CONDITIONALS_NEGATION mutants killed by the suite (checked one at a time), so the per-line waiver — `scripts/mutation.sh` matches file+line and never the mutant `$type`, so a reason written for the boundary mutant waives every other mutant on that line — is currently masking nothing. It was masking something before this commit: `endConn`'s negation mutant sat under that same waiver. Worth a note in `docs/quality-gateway.md` next to the `Kind` column, since the marker's reason and the mutant it actually waives are not the same set.
- `go test ./internal/...`, `gofmt -s`, `go vet` all clean at this head.

Two findings below, neither introduced by this commit — they are from the earlier commits on the branch and were never posted, so this is their first airing.

---

DEVELOPMENT.md
tomo-chan: Confirmed, and enforced rather than the claim dropped — in 32126a9.

You're right on the mechanism: both sites match the bare substring, so `//efficacy:exempt` with nothing after it waived a test silently. Given this PR is the thing promoting the marker as the first tool to reach for, an unenforceable "the reason is required" was the worst of the two options.

Ported the `bare_marker` treatment to both mappers (`:160` Go, `:241` frontend). A bare marker now waives nothing, the test stays in scope and is judged like every other, and the run names it:

```
efficacy: NOTE — //efficacy:exempt with no reason waives nothing.
  An exemption nobody has to justify is how an exemption stops being
  reviewable. Write what makes the test unable to go red, e.g.
  '//efficacy:exempt pins pre-existing behavior; no implementation under it changed'.
  Still in scope:
    b/b_test.go:12  TestBye
```

Deliberately a note and not an error: a bare marker is not itself a failing branch, and the test it sits on is judged on its own merits — so the branch fails only if that test really does survive, which is the verdict the gate is for. Failing on the marker alone would make the gate fire on something other than test efficacy.

Two tests in `scripts/efficacy_test.sh`, written first and confirmed failing before the change — one per mapper, since the Go and frontend halves have separate awk and a fix to one would not have covered the other:

```
ok   a bare efficacy:exempt waives nothing and says why
ok   a bare efficacy:exempt leaves a frontend case in scope
```

`all 31 efficacy checks passed`. This branch's own 15 markers all carry reasons, so nothing in the diff changes scope.

---

internal/board/relay_test.go
tomo-chan: Taken as written, in 32126a9 — `assert.InDelta` with `float64(time.Second)`, and `record` storing `remaining` unrounded.

Agreed on the reading: the cliff is at 500 ms because `Round` is round-to-nearest, and a stall that size produces `expected: 10s, actual: 9s`, which reads as broken timeout plumbing rather than as a slow runner. That is a worse failure than no test.

Checked that the tolerance does not cost the assertion anything: hand-applying `opTimeout < 0` still fails the test, since the mutant's budget is 0s and the delta is one second against a target of ten. Green under `-race -count=5`.

The "rounds **up**" sentence is gone with the rounding — you're right that it was wrong either way (`Round` is to-nearest, `Truncate` is down, neither is up). The replacement says what the helper now does and hands the tolerance question to the caller:

```go
// record stores the remaining budget as it stands, unrounded: the caller
// compares it with a tolerance, which is the honest way to assert on a value
// that has already ticked down by the time the client sees it.
```

---

tomo-chan:
All three findings addressed in 32126a9; the two inline threads have replies. This is for the review body's third point, the one about the marker's scope — it needed a correction of its own.

**The mechanism is exactly as described, and it is now recorded.** `scripts/mutation.sh` decides an exemption by file and line; `$type` is carried into the report and never compared. So a reason written about the boundary mutant waives every mutant gremlins produces on that line, `CONDITIONALS_NEGATION` included. Noted in `docs/quality-gateway.md` beside the `Kind` column and as a bullet in DEVELOPMENT.md.

**One correction: `endConn`'s negation mutant was not being masked.** I measured it rather than taking it forward — removed `TestRegistryEndConnNeverDrivesTheCounterNegative` and re-applied both mutants against the suite as it stood before this branch's fix:

```
without the new test:
  KILLED:   if f.active > 0  ->  if f.active <= 0     (CONDITIONALS_NEGATION)
  SURVIVED: if f.active > 0  ->  if f.active >= 0     (CONDITIONALS_BOUNDARY)
```

The negation mutant was killed by the existing suite the whole time. What survived under that waiver was the boundary mutant — the one its reason was written about. So the waiver was covering a mutant that was dying anyway, on that line as on the other five (checked each: all their negation mutants are killed independently, matching what you found).

That makes the gap **structural but not yet load-bearing**, which is how the docs now put it, rather than "it was masking something". The distinction is worth keeping because it is the one that decides urgency: today a line-scoped waiver hides nothing, so this is not a fix owed before merge. It starts mattering at stage 4, when a survivor becomes a failure and a waiver written for one mutant would be the difference between a red gate and a green one, with nothing in the diff to show which mutant it actually covered.

Not fixed here, deliberately: making the marker type-aware is a change to the gate's own contract (`//mutation:exempt[CONDITIONALS_BOUNDARY] <reason>`, or similar) and belongs with stage 4's decision rather than ahead of it. Happy to open an issue for it if you'd rather it not live only in the docs note.

---

PR #197, opened 2026-08-30
test: close G3 — drive /ws through the real router, round-trip Zod against Go (#191)

### LINKED ISSUE #191: test: G3（契約）を閉じる — `/ws` の実ルータ結合と Zod スキーマ往復
[docs/quality-gateway.md](https://github.com/tomo-chan/panemux/blob/main/docs/quality-gateway.md) の **G3（契約 / Resistance to refactoring）** に残っている 2 つ。#180 の項目 1 は (a) の HTTP 側だけを閉じた。

現在の G3 の状態:

| | 内容 | 状態 |
|---|---|---|
| (a) | 実 `server.New()` ルータ経由の HTTP/WS 結合 | **HTTP は present**（`internal/server/api_integration_test.go`）、**`/ws` は absent** |
| (b) | ルート表の網羅検査 | present |
| (c) | Zod スキーマ往復 | **absent** |
| (d) | agmsg 契約 | present |

## (a) `/ws` を実ルータ経由で駆動する

`api_integration_test.go` は `/api` の全ルートを `server.New()` 経由で叩き、ケースの無いルートがあれば fail する。`/ws` にはこれが無い。ルート表テストが「`/ws/board-command` が登録されているか」を報告しているだけで、**ハンドシェイクもフレームのやり取りも本番のルータを通っていない**。

`internal/ws` にはハンドラ単体のテストがあるが、それは #178 が指摘した「本番の配線を再構築したテスト」の形に近い — 161 件のテストが本番に存在しないルータに対して通っていた、というあの失敗と同じクラス。

対象は 2 本:

- `/ws/{sessionID}` — 認証なし。ハンドシェイク、バイナリフレーム、resize 制御メッセージ、切断時の status
- `/ws/board-command` — `Sec-WebSocket-Protocol` によるトークン認証。`command_center.enabled` が false のときは**ルート自体が存在しない**ことも含めて

後者は [docs/security.md](https://github.com/tomo-chan/panemux/blob/main/docs/security.md) が明示的に設計を記録している箇所なので、本番ルータ経由での担保に意味がある。

## (c) Zod スキーマ往復

`frontend/src/schemas/index.test.ts` に 103 件のテストがあるが、**Go が実際に出力した JSON を 1 件も通していない**（`testdata` / golden / fixture への参照ゼロ）。手書きの TypeScript オブジェクトに対する検証なので、

- Go 構造体のフィールド名や型が変わり
- Zod スキーマがそのまま残っても

**両側のテストは通り続ける。** これが「往復が無い」の実際の意味で、フロントエンドが実行時に落ちるまで誰も気づかない。

案として、Go 側のハンドラテストが返す JSON を golden ファイルに書き出し、vitest がそれを読んで Zod で検証する。生成の向きは Go → fixture → TS の一方向でよく、fixture が古くなったら Go 側のテストが更新する。

## 完了条件

- [ ] `/ws/{sessionID}` と `/ws/board-command` が `server.New()` のルータ経由で駆動されている
- [ ] `command_center.enabled: false` で `/ws/board-command` が存在しないことが本番ルータに対して確認されている
- [ ] Go が出力した実 JSON に対して Zod スキーマが検証されている
- [ ] Go 側のスキーマを壊すと、フロントエンドのテストが落ちる（摂動で確認すること）
- [ ] `make check` が通る

## 関連

- #180 — 項目 1 が (a) の HTTP 側を閉じた
- #178 — 本番配線を再構築したテストの調査
- #182 — `api_integration_test.go`（HTTP 側の先例）

tomo-chan:
docs/quality-gateway.md's gate G3 (contract / resistance to refactoring) had
two halves open. #180's item 1 closed (a) for HTTP only.

(a) /ws through the production router
-------------------------------------
route_table_test.go reported that /ws/{sessionID} and /ws/board-command were
registered; nothing completed a handshake or exchanged a frame through the
router server.New() builds. internal/ws's own handler tests mount the handler
on a router they build themselves — the shape #178 found, where 161 tests
passed against a router that does not exist in production.

internal/server/ws_integration_test.go drives both routes over a real TCP
listener (an upgrade needs http.Hijacker, which httptest.ResponseRecorder does
not implement):

- /ws/{sessionID}: handshake, binary frames both directions, output larger
  than the pump's 4096-byte buffer arriving whole, the resize control message,
  the replay bracket around buffered output, the final status frame when a
  pane exits, and 404 for an unknown pane.
- /ws/board-command: absent — chi's 404, not the handler's 401 — when the
  command center is disabled; missing/wrong/empty subprotocol tokens rejected;
  the accepted handshake echoing the subprotocol; a malformed prompt's error
  frame; a query streamed to completion against a `claude` stand-in; and a
  second concurrent query answered `busy`, gated on a file rather than timing.
- That /ws/board-command wins over /ws/{sessionID} even when a pane is named
  "board-command", which nothing else asserted.

Both directions were confirmed by perturbation: dropping the subprotocol echo
and registering the route unconditionally each fail these tests.

(c) Zod schema round-trips
--------------------------
frontend/src/schemas/index.test.ts had 103 cases and none of them saw JSON Go
produced, so a renamed Go field with an unchanged Zod schema kept both suites
green until the dashboard failed to parse a live response.

internal/server/contract_fixture_test.go captures every response the dashboard
parses, plus both WebSocket frame streams, from the real router into
testdata/api-contract/. frontend/src/schemas/contract.test.ts parses each with
the schema that owns it, and additionally requires the parsed value to equal
the captured one — Zod strips undeclared keys, so an optional field renamed in
Go passes safeParse cleanly and only the equality check sees it.

The fixtures are rewritten on every Go test run rather than diffed against.
A golden test that failed on drift would stop at the Go suite, and the frontend
— the side holding the stale schema — would never see the new shape. The Go
suite runs first in make test, make check and ci.yml, so the frontend is what
goes red. Verified by perturbation both ways: renaming is_status (required)
fails the parse; renaming last_tool (optional) fails the equality check.

Only values are normalized, never keys or types: timestamps, the capture's temp
HOME (also DEVELOPMENT.md's path-sanitization rule), the random BoardCache
epoch, and the detected shell. Substitutions are applied longest-key-first, so
the order does not depend on Go's randomized map iteration.
TestAPIContractFixtures_ContainNoMachinePaths re-checks the path half against
the bytes each capture produces — not against the committed files, since for
that failure the committed file is the suspect rather than the reference.

Every capture must also be deterministic: TestAPIContractFixtures_AreDeterministic
runs each 20 times and requires identical bytes, after GET /api/sessions was
found writing its rows in Go's map order (measured: 4 of 20 runs flipped).

Which optional fields a capture leaves unpopulated is derived rather than
declared: contract.test.ts walks each schema against its capture and names
every optional field absent everywhere, against a list carrying a reason per
entry. The discriminated-union checks read the union's own .options, so a
variant no Go code emits has to be classified rather than ignored.

Both sides carry exhaustiveness checks: a fixture with no reader, a reader with
no fixture, and an exported schema classified as neither round-tripped nor
explained all fail.

Also: @types/node as a devDependency, so the frontend test can read the
fixtures; scenario rows E5, E8 and H15 updated and H16/H17 added; G3 marked
present with rollout row 9 and decision D10 in quality-gateway.md; the
regeneration workflow written down in DEVELOPMENT.md.

#198, and the fixture follow-up it asked for
--------------------------------------------
Building this gate found #198 — a pane-only root layout passes Go validation
but is rejected by LayoutNodeSchema — which was split out rather than decided
here, because which side is wrong is a production question. It is now fixed and
merged (#199), so this branch is rebased onto it and the fixture no longer has
to work around it.

The workspaces capture previously seeded a root `pane` *alongside* children,
with a comment calling that a compromise: capturing a genuine pane-only root
produced JSON the frontend rejected, which was the bug itself. It now captures
both shapes #199 settled:

- `scratch` keeps `{pane, children}`. normalizeLayoutNode relocates a root pane
  only when the node has no children, so this reaches the wire exactly as
  written — which is why LayoutNodeSchema still declares `pane`, and why
  renaming that json tag has to fail something.
- `solo` is a hand-written pane-only root, seeded unmigrated and captured after
  WorkspacesView normalizes it, so testdata/api-contract/workspaces.json holds
  the migration's own output: no root `pane`, a filled-in direction, and the
  pane in the single child it means. Seeding the already-migrated shape would
  have produced the same bytes while proving nothing about the migration.

The rebase's one conflict was docs/scenarios.md: main's H14a–H14e rows (added
by #199) are kept, and this branch's rewritten H15 plus H16/H17 follow them.

Closes #191

https://claude.ai/code/session_01Veqb91P9norzcFVuovCQE3


internal/server/contract_fixture_test.go
```diff
@@ -0,0 +1,623 @@
...
+
+	"session-token": {capture: func(t *testing.T) ([]byte, map[string]string) {
+		e := newAPIEnv(t)
+
+		rr := e.do(t, http.MethodGet, "/api/session-token", "")
+		require.Equal(t, http.StatusOK, rr.Code)
+		return rr.Body.Bytes(), nil
+	}},
+
+	"open-url": {capture: func(t *testing.T) ([]byte, map[string]string) {
```
tomo-chan, line 228: **Undeclared gap: `open-url` never exercises the optional `port` field.**

`e.createPane` makes a *local* pane, so `POST /api/sessions/{id}/open-url` always answers `forwarded:false` with `reason` set and no `port` — see the captured `testdata/api-contract/open-url.json`. But this fixture sets no `optionalFieldsUnexercised`, and `documentedContractGaps` lists only `git-info`, so the suite asserts that this capture *does* cover every optional field its schema allows. It does not.

Verified by perturbation. Change `internal/api/openurl.go:29` to `` Port int `json:"forwarded_port,omitempty"` ``, then:

```
go test ./internal/server -run TestAPIContractFixtures   # ok
git diff --stat testdata/                                # empty
cd frontend && npx vitest run src/schemas/contract.test.ts  # 36 passed
```

Both suites stay green while `OpenUrlResponseSchema.port` would silently stop being populated in the dashboard — the exact class of failure G3(c) exists to catch.

Note that `TestAPIContractFixtures_DocumentedGaps` cannot see this: it compares the set of fixtures that *carry* a prose gap string against the hardcoded list of the same names, so it only detects a declared gap going undeclared, never an undeclared one. Either capture through an SSH pane so `port`/`forwarded:true` appear, or declare the gap here.

---

frontend/src/schemas/contract.test.ts
```diff
@@ -0,0 +1,239 @@
...
+// the capture" says nothing about the variants the capture never contained.
+// These two cases pin which variants the server actually produced, and name
+// the ones it does not — the alternative being a union that grows a variant
+// no Go code emits, with nothing to notice.
+describe('discriminated union coverage', () => {
+  function variantsIn(fixture: string): string[] {
+    return [...new Set((readFixture(fixture) as { type: string }[]).map((frame) => frame.type))].sort()
+  }
+
+  it('captures every control frame /ws/{sessionID} emits', () => {
```
tomo-chan: **These two cases never look at the schema, so they cannot do what the block comment says they do.**

`variantsIn` reads only the fixture, and both cases compare its output to a hardcoded array. The stated purpose — "the alternative being a union that grows a variant no Go code emits, with nothing to notice" — is not achieved: add a fifth member to `WSControlMessageSchema` (say `z.object({ type: z.literal('ping') })`) that no Go code emits, and `variantsIn('ws-control-frames')` still returns `['replay','status']`, so both `captures every control frame …` and `names the control variants with no server-side producer` still pass.

What the second case actually asserts today is a tautology over the first case's output: `['resize','error'].filter(v => emitted.includes(v))` is empty precisely because `emitted` is already pinned to `['replay','status']`.

The check that would hold is deriving the union's own variants from the schema and comparing *those* to the emitted set, e.g.

```ts
const declared = WSControlMessageSchema.options
  .map((o) => o.shape.type.value)
  .sort()
expect(declared.filter((v) => !emitted.includes(v))).toEqual(['error', 'resize'])
```

so a variant added to the schema with no producer has to be classified rather than silently ignored.

---

internal/server/contract_fixture_test.go
```diff
@@ -0,0 +1,623 @@
...
+	}},
+}
+
+// seedRichLayout replaces the starting single-pane layout with one that
+// exercises every field PaneConfigSchema and LayoutChildSchema allow: a
+// nested split, all four pane types' distinguishing fields, the optional
+// per-pane header/status-bar overrides, and agent_board. A fixture built
+// from the default layout would validate a schema against three fields and
+// report the contract as covered.
+func seedRichLayout(cfg *config.Config) {
```
tomo-chan, line 352: **Second undeclared gap, same class: `LayoutNode.Pane` is never populated.**

The doc comment scopes the claim to "every field `PaneConfigSchema` and `LayoutChildSchema` allow", and that part holds. But `LayoutNodeSchema` also has an optional `pane`, backed by `config.LayoutNode.Pane` (`internal/config/config.go:63`, `json:"pane,omitempty"`), and neither seeded workspace sets it — every layout here is `direction` + `children`, so `testdata/api-contract/workspaces.json` contains no root-level `pane`.

Concretely: a user whose workspace layout is a single root pane (`layout: {pane: {...}}`, no children) serializes that key. Rename it on the Go side and `workspaces.json` is unchanged, so both suites stay green while `LayoutNodeSchema.pane` stops matching. Because `useLayout` parses the whole tree and PUTs it back wholesale on any edit, the stripped field is then deleted from the user's `config.yaml` — which is the `agent_board` regression `PaneConfigSchema`'s own comment records ("a field missing from this schema is silently deleted from the user's config.yaml by an unrelated action").

Cheapest fix is a third seeded workspace whose `Layout` is `config.LayoutNode{Pane: &config.PaneConfig{...}}`.

---

internal/server/contract_fixture_test.go
```diff
@@ -0,0 +1,623 @@
...
+	)
+
+	out, err := json.Marshal(frames)
+	require.NoError(t, err)
+	return out
+}
+
+// readRawControl reads one text frame and returns it byte for byte, so the
+// fixture records what the server wrote rather than a re-encoding of it.
+func readRawControl(t *testing.T, conn *websocket.Conn) json.RawMessage {
```
tomo-chan, line 525: **No read deadline here, unlike `readFrame`.**

`ws_integration_test.go`'s `readFrame` sets `conn.SetReadDeadline(time.Now().Add(wsReadTimeout))` before every read, with a comment saying it exists "so a handler that never writes fails this test rather than the whole package". `readRawControl` calls `conn.ReadMessage()` with no deadline at all.

In `captureBoardCommandFrames` no deadline is ever set on either connection, so a regression that stops one of the four frames from being emitted — `writeBoardCommandFrame` losing the `busy` case, or the runner's event channel never closing — blocks forever at `readRawControl(t, second)` / `readRawControl(t, first)`. The suite then dies on the package-level 10-minute timeout with a goroutine dump instead of a 5s failure naming the missing frame.

It is easy to miss because `captureWSControlFrames` happens to survive: its one intervening `readFrame` call sets an absolute deadline that the following `readRawControl` calls inherit. That is accidental, and it disappears if the frame order changes.

```suggestion
func readRawControl(t *testing.T, conn *websocket.Conn) json.RawMessage {
	t.Helper()

	require.NoError(t, conn.SetReadDeadline(time.Now().Add(wsReadTimeout)))
	msgType, data, err := conn.ReadMessage()
```

---

tomo-chan (commented):
Reviewed the diff for correctness. The two integration files do what they claim — I ran `go test ./internal/server -race -count=1` and the frontend contract suite locally, both green, fixtures regenerate byte-identically, and `make check-scenarios` resolves the new H16/H17/E5/E8 rows. Route precedence, the subprotocol echo, the absent-vs-rejecting distinction and the equality-over-safeParse trick all hold up.

Four findings, all about gaps in what the new gate can actually detect rather than broken code:

1. `open-url` never exercises the optional `port` field and does not declare the gap — confirmed by perturbation that a rename of `openURLResponse.Port` leaves both suites green.
2. The `discriminated union coverage` cases read only the fixture, never the schema, so they cannot catch the thing their comment says they catch.
3. `LayoutNode.Pane` is never populated by `seedRichLayout` — same class as (1), with a config-data-loss consequence.
4. `readRawControl` reads without the deadline every other read in the suite sets.

None of these block; (1) and (3) are the ones I would fix before merge, since they are holes in the gate this PR exists to build.

---

internal/server/contract_fixture_test.go
```diff
@@ -0,0 +1,623 @@
...
+		e := newAPIEnv(t)
+		// Fake panes rather than real ones: a real PTY pane's reported state
+		// depends on how far its shell has got, and this fixture is read by
+		// a schema whose `state` is an enum. api_integration_test.go drives
+		// the genuine POST /api/sessions path.
+		e.mgr.Add(newWSFakeSession("pane-editor"))
+		e.mgr.Add(newWSFakeSession("pane-build"))
+
+		rr := e.do(t, http.MethodGet, "/api/sessions", "")
+		require.Equal(t, http.StatusOK, rr.Code)
```
tomo-chan, line 190: **`sessions.json` regenerates non-deterministically, so an unrelated `make test-go` dirties the working tree roughly a quarter of the time.**

The capture writes `GET /api/sessions` straight through, and that response's order comes from `Manager.List()` (`internal/session/manager.go:85-93`), which ranges over the `m.sessions` map:

```go
list := make([]Session, 0, len(m.sessions))
for _, entry := range m.sessions {
	list = append(list, entry.session)
}
return list
```

Go randomizes map iteration, and nothing sorts before the bytes are written, so `pane-editor` and `pane-build` swap places between runs. The diff is purely positional — same keys, same values, same types — which is exactly the shape a reviewer learns to wave through.

Measured on this branch at `af08ecc`, 20 consecutive runs of `go test ./internal/server -run TestAPIContractFixtures -count=1`, recording which pane lands first:

```
build-first runs: 5 / 20
```

The consequence is specific to how D10 chose to make fixtures work. Because they are rewritten on every Go test run rather than diffed against, there is no failing test here — the run silently succeeds and leaves a modified file behind. DEVELOPMENT.md then says to "commit the fixture diff alongside the struct change; it is the contract change, in the diff a reviewer reads", so about one branch in four picks up a `sessions.json` reordering that has nothing to do with the change it ships. A contract fixture whose diff is sometimes noise is a weaker signal than one whose diff is always meaningful, and this is the one capture in the set where a reader cannot tell the two apart.

Sorting inside the capture is enough, and keeps it local to the fixture rather than changing an API behavior the dashboard does not depend on:

```go
rr := e.do(t, http.MethodGet, "/api/sessions", "")
require.Equal(t, http.StatusOK, rr.Code)

var sessions []api.SessionResponse
require.NoError(t, json.Unmarshal(rr.Body.Bytes(), &sessions))
sort.Slice(sessions, func(i, j int) bool { return sessions[i].ID < sessions[j].ID })
sorted, err := json.Marshal(sessions)
require.NoError(t, err)
return sorted, nil
```

Worth checking the other multi-element captures for the same property before merge — `directories` is the other one that returns a list, though that one is built by a `filepath` walk rather than a map, so it should already be stable.

One correction to my earlier review on this PR, since it bears on this: I reported that the fixtures regenerate byte-identically and leave `git status` clean. That was true of the run I checked, but it is not a property of the suite — this capture is a coin flip, and I happened to see the side that matched.

---

tomo-chan (commented):
One follow-up finding, plus a correction to my earlier review.

While re-checking the working tree after the previous pass I found `testdata/api-contract/sessions.json` modified with no source change behind it. The cause is Go's randomized map iteration reaching the fixture through `Manager.List()`, so that one capture is written in a different order between runs — measured at 5 of 20 runs on `af08ecc`. Because D10's design rewrites fixtures instead of diffing them, nothing fails; the run just leaves a modified file behind, and DEVELOPMENT.md's "commit the fixture diff" rule then pulls that noise into unrelated branches. Details and a fix are in the inline comment.

The correction: I previously reported that the fixtures regenerate byte-identically with a clean `git status`. That held for the run I checked, but it is not a property of the suite.

This does not change my overall read of the PR — the four earlier findings were gaps in what the new gate can detect, and this is a fifth of the same kind rather than anything broken in the code under test. I'd still fix the `open-url` optional-field gap and the root-level `pane` gap before merge; this one is cheap enough to fold in alongside them.

---

internal/server/contract_fixture_test.go
tomo-chan: Fixed in 7201702, taking the second half of your point rather than just declaring the gap.

You're right that `TestAPIContractFixtures_DocumentedGaps` could only ever catch a declared gap going undeclared — it compared a hand-written set against a hand-written set. That machinery is deleted. Optionality is stated in the Zod schemas, so the check now lives where it can be *derived*: `contract.test.ts` walks each schema against its capture and names every optional field that no capture populates anywhere in the corpus, compared against a declared list carrying a reason per entry.

It immediately found both undeclared gaps you identified — `open-url.port` and `LayoutNodeSchema.pane` (your other thread).

`port` stays declared rather than captured: `forwarded:true` with a port requires an *ssh* pane so there is somewhere to forward to, which needs a second host `make check` cannot assume. But it is now declared mechanically — a newly optional Go field that no fixture exercises fails the suite rather than passing silently, and a field that starts being covered has to be deleted from the list.

The walk is tested rather than trusted: arrays, records, `z.lazy` recursion, discriminated-union member selection, and the cycle-collapsing rule, each against a schema whose answer is known.

---

frontend/src/schemas/contract.test.ts
tomo-chan: Correct on both counts — the second case was a tautology over the first, and neither looked at the schema. Fixed in 7201702 essentially as you wrote it.

`declaredVariants()` reads the union's own `.options` and each member's `type` literal, and the two cases now assert in both directions: the emitted set is pinned, and `declared.filter(v => !emitted.includes(v))` must equal the documented non-producers (`['error', 'resize']` for control frames, `[]` for board-command frames).

Verified by perturbation: adding `z.object({ type: z.literal('ping') })` to `WSControlMessageSchema` now fails `classifies every WSControlMessage variant as emitted or accounted for`, where before it passed both cases.

---

internal/server/contract_fixture_test.go
tomo-chan: Fixed in 7201702 — and following your suggested fix literally surfaced something bigger, so recording it here rather than quietly working around it.

I first seeded a third workspace with `Layout: config.LayoutNode{Pane: &config.PaneConfig{...}}` exactly as you proposed. The capture came out as `{"pane":{...}}` and the frontend suite went **red**, not green: `config.LayoutNode`'s `direction` and `children` are both `omitempty`, while `LayoutNodeSchema` requires both. So a pane-only root layout — which `config.Validate` accepts, since `validateLayoutNode` permits an empty direction and no children — makes `WorkspacesResponseSchema.parse` throw on the *whole* response, blanking the dashboard rather than just dropping a key.

That mismatch predates this branch, and which side is wrong is a production question (loosen the schema, or normalize the layout on the Go side), so I did not decide it in a test-only PR. What I did instead:

- `pane` is now set **alongside** `children` on the seeded workspace, so the key reaches the fixture and a rename of its json tag fails. Verified: renaming it to `root_pane` now fails both the equality check and the new optional-field check, where before both suites stayed green.
- The compromise and the pane-only finding are written out in the seed's comment, so the next reader sees why it is not the shape you suggested.

Happy to open a separate issue for the pane-only shape if you want it tracked.

---

internal/server/contract_fixture_test.go
tomo-chan: Applied in 7201702, with your diagnosis kept as the comment — the accidental-inheritance detail is the part worth writing down, since it is exactly why this looked fine.

---

internal/server/contract_fixture_test.go
tomo-chan: Confirmed and fixed in 26bf924. Reproduced independently before touching anything: 4 of 20 runs flipped, matching your measurement.

Your correction stands, and mine goes with it — I reported the fixtures regenerate byte-identically, and that was two samples rather than a property.

**The fix.** The capture sorts, but not through a typed struct. Your snippet's `api.SessionResponse` does not exist — `internal/api`'s `sessionInfo` is unexported, so there is nothing to decode into from `internal/server` — and that turns out to be a good thing: decoding into a struct and re-encoding would drop any key the struct does not name, so a Go field added to the response would silently vanish from the fixture, deleting the contract the file exists to carry. It sorts `[]json.RawMessage` by the `id` field instead, leaving every element's bytes untouched.

Sorted in the capture rather than in `GetSessions`, per your point that no client depends on the order — imposing one on the API would be a fixture changing production behavior to suit itself.

**The other captures.** Checked, all stable for a stated reason: `GetSSHConnections` (`handler.go:550`) and `GetDirectories` (`handler.go:1467`) both sort in the handler, `board-messages` is seq-ordered, the two ws fixtures are frame-ordered, and `board-status` is a JSON object whose keys Go sorts on marshal.

**The guard.** `TestAPIContractFixtures_AreDeterministic` runs every capture 20 times and requires identical bytes, so the next map-ordered capture fails rather than being committed.

The repetition count is 20 because 5 did not work, and the reason is worth recording. I first sized it on the assumption that a two-element swap escapes n runs with probability `0.5^n`. Go does not hand out uniform permutations — it randomizes the starting *offset* within a bucket, so one order comes up about seven times in eight (measured 87.6% / 12.4% over 100k iterations). At 5 runs the guard caught this very bug 6 times in 12, against a predicted 19 in 20. At 20 runs it catches it 18 times in 20.

So the guard is a net for the next capture, not what makes this one correct — the sort is. Both the measurement and that caveat are in the test's comment and in `testdata/api-contract/README.md`.

---

internal/server/ws_integration_test.go
```diff
@@ -0,0 +1,517 @@
...
+func (s *wsFakeSession) Title() string        { return s.id }
+func (s *wsFakeSession) State() session.State { return s.state }
+
+func (s *wsFakeSession) Read(p []byte) (int, error) {
+	data, ok := <-s.out
+	if !ok {
+		return 0, io.EOF
+	}
+	return copy(p, data), nil
+}
```
tomo-chan, line 208: **`copy(p, data)` discards the overflow instead of retaining it, so this fake silently truncates anything larger than the pump's buffer.**

`managedSession.pump` (`internal/session/manager.go:125-127`) reads through a fixed 4096-byte buffer:

```go
buf := make([]byte, 4096)
for {
	n, err := m.session.Read(buf)
```

A real `Session` — a PTY, an SSH channel — is an `io.Reader`, so a short read is normal and the remainder arrives on the next call. This fake drops it: the `data` slice is taken off the channel whole, `copy` fills 4096 bytes of it, and the other 904 are gone with no error and no second read.

Measured against the real router on `26bf924` with a throwaway test that pushes 5000 bytes through `wsFakeSession.out` and sums what the client receives:

```
Error: Not equal:
       expected: 5000
       actual  : 4096
```

Nothing in the suite pushes more than 4096 today, so this is latent rather than a live failure. It matters because of what the file is for. `TestWSIntegration_ReplayBracketsBufferedOutput` already exercises the replay path, and the replay buffer's own bound is 256KB — the natural next test here is one that fills it and asserts what a reconnecting client gets back. Written against this fake, that test would push its payload, receive a truncated prefix, and assert successfully against the truncated value. A test that passes while measuring the wrong thing is worse than an absent one, and this is a `/ws` integration file specifically because the unit-level fakes were not trusted to represent the real transport.

The fix keeps the remainder for the next call:

```go
type wsFakeSession struct {
	// ...
	pending []byte
}

func (s *wsFakeSession) Read(p []byte) (int, error) {
	if len(s.pending) == 0 {
		data, ok := <-s.out
		if !ok {
			return 0, io.EOF
		}
		s.pending = data
	}
	n := copy(p, s.pending)
	s.pending = s.pending[n:]
	return n, nil
}
```

Worth noting `internal/ws/handler_test.go:86` copies the whole slice rather than truncating, so the two fakes for the same interface currently disagree about this.

---

internal/server/contract_fixture_test.go
```diff
@@ -0,0 +1,732 @@
...
+// DEVELOPMENT.md's path-sanitization rule applies to fixtures by name, and a
+// normalization rule that silently stopped matching would put a developer's
+// home directory into a committed file with no other test noticing.
+func TestAPIContractFixtures_ContainNoMachinePaths(t *testing.T) {
+	home, err := os.UserHomeDir()
+	require.NoError(t, err)
+
+	entries, err := os.ReadDir(contractFixtureDir)
+	require.NoError(t, err)
+
```
tomo-chan: **This test reads the committed files, not the ones this run produced — so a `-run` filter turns it into a no-op that still reports success.**

`os.ReadDir(contractFixtureDir)` reads from disk, and what is on disk is only this run's output because `TestAPIContractFixtures` (line 558) happens to be declared earlier in the same file and `go test` runs a file's tests in declaration order. Nothing states that dependency and nothing enforces it. `TestAPIContractFixtures_DirectoryHasNoStrayFiles` (line 611) has the same shape.

Demonstrated on `26bf924` by breaking a normalization rule in the generator — dropping the `map[string]string{e.home: fixtureHomePath}` literal from the `directories` capture — while leaving the committed fixtures untouched:

```
$ go test ./internal/server -run 'TestAPIContractFixtures_ContainNoMachinePaths' -count=1
ok  	panemux/internal/server	0.006s

$ go test ./internal/server -run 'TestAPIContractFixtures' -count=1
    Messages: directories.json carries a temp path from the capturing machine
FAIL
```

The first invocation is a plausible thing to type when iterating on a normalization rule — which is exactly when the rule is broken — and it passes in 6ms because it re-validated files this branch already committed. It never regenerated anything, so it could not have failed for any reason other than the repository already being dirty.

This one is worth more than the ordering fragility alone, because of the claim in the comment directly above it: "a normalization rule that silently stopped matching would put a developer's home directory into a committed file with no other test noticing." `testdata/api-contract/README.md` and D10 both present this as *the* guard for DEVELOPMENT.md's path-sanitization rule on fixtures. A guard whose subject is "the files already in git" rather than "the bytes this run just wrote" is not that guard — for the failure it describes, the committed file is the thing under suspicion, not the reference.

Making it independent of run order also makes it say what it means:

```go
func TestAPIContractFixtures_ContainNoMachinePaths(t *testing.T) {
	home, err := os.UserHomeDir()
	require.NoError(t, err)

	for name, fixture := range contractFixtures {
		t.Run(name, func(t *testing.T) {
			body, literals := fixture.capture(t)
			written := string(normalizedFixtureBytes(t, body, literals))
			assert.NotContains(t, written, home, ...)
			// ...
		})
	}
}
```

`normalizedFixtureBytes` was split out in 26bf924 for the determinism check for this same reason — "exactly what lands on disk, minus the write" — so the seam already exists. `_DirectoryHasNoStrayFiles` is genuinely about the directory rather than about content, so it can keep reading disk; it just needs to not depend on a preceding test having written it, which a comment naming the dependency would cover.

Related: nothing here runs with `-shuffle=on` today (the Makefile's `go test ./... -v -race` does not set it), so this is latent rather than flaky in CI. Turning shuffle on later would surface it as an intermittent failure whose cause is several steps removed from its symptom.

---

internal/server/contract_fixture_test.go
```diff
@@ -0,0 +1,732 @@
...
+	}
+}
+
+func normalizeFixtureString(s string, literals map[string]string) string {
+	for from, to := range literals {
+		if from == "" || from == to {
+			continue
+		}
+		s = strings.ReplaceAll(s, from, to)
+	}
```
tomo-chan, line 731: **`literals` is a map, so these substitutions are applied in Go's randomized iteration order.**

Same root cause as the `sessions` ordering fixed in 26bf924, one layer down: nothing here is sorted, so with more than one entry the result depends on which replacement runs first.

Harmless as the code stands — every capture passes at most one literal, and one replacement has no order to depend on. It stops being harmless the moment a second is added, and the two most likely candidates already overlap by prefix. `newWSEnvIn` sets both:

```go
t.Setenv("HOME", home)
```

A capture normalizing both would pass `{home: "/tmp/sample-project", home + "/.cache": "/tmp/sample-project/.cache"}`. Replacing `home` first rewrites the cache path's prefix too, leaving `/tmp/sample-project/.cache`, which the second replacement can no longer match — the same output by luck. Reverse the order and the cache path is substituted whole. The two agree here, but only because the placeholders were chosen to nest the same way the real paths do; a placeholder that did not (`/remote/home/demo` for HOME, `/tmp/sample-cache` for the cache dir) would produce two different fixtures at random from the same run.

That failure is the one this file is least able to absorb: D10 rewrites rather than diffs, so it lands as a committed file that alternates between two spellings, and the next `git status` attributes it to whichever branch ran the suite.

Sorting longest-first makes the more specific pattern win, which is the intent when one key is a prefix of another:

```go
func normalizeFixtureString(s string, literals map[string]string) string {
	froms := make([]string, 0, len(literals))
	for from := range literals {
		froms = append(froms, from)
	}
	// Longest first: XDG_CACHE_HOME is HOME + "/.cache", so replacing HOME
	// first would leave the cache path unmatchable. Sorted at all because
	// map order is randomized — see TestAPIContractFixtures_AreDeterministic.
	sort.Slice(froms, func(i, j int) bool {
		if len(froms[i]) != len(froms[j]) {
			return len(froms[i]) > len(froms[j])
		}
		return froms[i] < froms[j]
	})

	for _, from := range froms {
		to := literals[from]
		if from == "" || from == to {
			continue
		}
		s = strings.ReplaceAll(s, from, to)
	}
	// ...
}
```

Worth pointing out that `TestAPIContractFixtures_AreDeterministic` from 26bf924 would catch this one, which is the case for adding it being a net rather than a formality — it just catches it at whatever ~0.876^n rate applies to the map involved, so the sort is still what makes it correct.

---

tomo-chan (commented):
Re-reviewed at `26bf924`. The five earlier findings are all genuinely fixed — I confirmed each by perturbation rather than by reading the diff:

- `readRawControl` now sets its own deadline, and the accidental-inheritance explanation is kept as the comment.
- The derived optional-field walk really is derived: deleting `open-url.port` from the declared list makes the suite name it again, so it is no longer a set compared against itself.
- Adding an unproduced `z.literal('ping')` variant to `WSControlMessageSchema` now fails.
- Renaming `LayoutNode.Pane`'s json tag to `root_pane` now fails two checks; before it failed none.
- `sessions` is stable at 20/20 runs (was 5/20), and reverting just the sort makes the new determinism guard fire 19 of 20 times — close enough to your measured 18/20 that the 0.876^n model looks right.

Two notes on that last one. Your fix is better than the snippet I suggested: sorting `[]json.RawMessage` avoids the struct round-trip, which would have dropped any key the struct does not name and quietly deleted the contract the file carries. And sizing the guard by measurement after the uniform-permutation estimate turned out wrong is the part I'd keep — along with the comment saying the sort, not the guard, is what makes this capture correct.

Three findings from my earlier pass were never posted, so they are still open. None of them fails anything today; all three are cases where the thing meant to catch a problem cannot:

1. `wsFakeSession.Read` truncates at the pump's 4096-byte buffer instead of retaining the remainder (measured: 5000 pushed, 4096 received). Latent now, but the natural next test in that file — the 256KB replay bound through the real router — would assert against a truncated value and pass.
2. `TestAPIContractFixtures_ContainNoMachinePaths` validates the committed fixtures rather than this run's output, so `-run 'TestAPIContractFixtures_ContainNoMachinePaths'` passes in 6ms with the generator's normalization deliberately broken. That is the one D10 and the README present as the guard for the path-sanitization rule.
3. `normalizeFixtureString` applies its substitutions in map order — the same class as the `sessions` bug, one layer down, and inert only while every capture passes a single literal.

(2) is the one I'd act on before merge; `normalizedFixtureBytes`, split out in 26bf924, already gives it the seam it needs. (1) and (3) are cheap and can wait for whoever next touches those files.

---

internal/server/ws_integration_test.go
tomo-chan: Confirmed and fixed in 5bd29a3. Reproduced first with a throwaway test pushing 5000 bytes: the client got 4096 and then blocked, so the read deadline fired rather than the assertion — same defect, one step earlier than your version showed it.

Applied your fix as written. What made it worth doing now rather than leaving latent is your framing of it: the 256KB replay test is the obvious next thing to write in that file, and against a truncating fake it would have asserted successfully against a truncated prefix. That is in `Read`'s doc comment.

Also added `TestWSIntegration_TerminalRoute_StreamsOutputLargerThanThePumpBuffer`, with a payload deliberately not a multiple of 4096 so a fake that truncated at the boundary could not pass by coincidence. Reverting the `pending` logic fails it.

Good catch on `internal/ws/handler_test.go:86` — the two fakes did disagree, and this one was the wrong one.

---

internal/server/contract_fixture_test.go
tomo-chan: Confirmed and fixed in 5bd29a3 — this was the one I'd have wanted caught too, for exactly the reason you give: a guard whose subject is "the files already in git" is not a guard for a rule about what gets written into git.

Reproduced your demonstration verbatim on `26bf924`. With the `directories` capture's literal removed:

```
$ go test ./internal/server -run 'TestAPIContractFixtures_ContainNoMachinePaths' -count=1
ok      panemux/internal/server 0.071s

$ go test ./internal/server -run 'TestAPIContractFixtures' -count=1
FAIL
```

It now re-captures and normalizes through `normalizedFixtureBytes`, per your snippet. Same perturbation, run with the same `-run` filter, now fails naming `directories`.

`_DirectoryHasNoStrayFiles` keeps reading disk, since it genuinely is about the directory. Its run-order dependency is now stated in the comment, along with which half survives without one: a *stray* file from a rename is still reported when run alone, a *missing* one only means nobody generated it yet — and that weaker half is covered from the other side by the frontend's "reads every fixture the Go side captured".

On `-shuffle=on`: I ran the package under it after the fix and it passes, so the latent-vs-flaky distinction you drew is closed rather than just deferred.

---

internal/server/contract_fixture_test.go
tomo-chan: Fixed in 5bd29a3, sorted longest-first as you wrote it.

The detail I took from your comment and built the test around: the current placeholders agree in both orders only because they nest the same way the real paths do. So `TestNormalizeFixtureString_AppliesTheMoreSpecificLiteralFirst` uses deliberately **non**-nesting placeholders (`/remote/home/demo` for HOME, `/tmp/sample-cache` for the cache dir) — with nesting ones both orders agree by luck and the test would pass against the unsorted code. It fails against it, and repeats 20× so a map that happened to range favourably cannot make it flake green.

Agreed on the last point, and it is the same shape as the `sessions` conclusion: `_AreDeterministic` would catch this at whatever ~0.876^n rate the map gives, which is what makes it a net rather than a formality — but the sort is what makes it correct.

---

internal/server/contract_fixture_test.go
tomo-chan: Filed as #198, so this thread can close.

Re-verified the whole chain at `5bd29a3` before writing it up, rather than carrying over the earlier measurement:

- `ValidateLayout(LayoutNode{Pane: ...})` returns `<nil>` — Go accepts the shape.
- Seeded into the fixture, `GET /api/workspaces` emits `{"pane":{"agent_board":{},"id":"pane-root","type":"local"}}`.
- The frontend contract suite fails with `invalid_type` on the missing `direction`/`children` — and because `WorkspacesResponseSchema` covers the whole response, that is the dashboard failing to load the workspace list, not one key being dropped.

One further thing the issue records that this thread did not: nothing in `frontend/src` reads the **root** `LayoutNode.pane` at all — `SplitContainer`, `App.tsx` and `useWorkspaceAttentionMonitor` all read `child.pane` off `LayoutChild`. So the field is accepted by the schema and rendered by nothing, which is part of why "loosen the schema" looks like the weakest of the three options there.

What stays in this PR is unchanged: `pane` set alongside `children`, so the key reaches the fixture and a rename of its json tag fails two checks. Capturing the pane-only shape belongs with whichever fix #198 takes.

---

PR #199, opened 2026-09-02
fix: always serve a layout node the dashboard can parse (#198)

### LINKED ISSUE #198: fix: a pane-only root layout passes Go validation but is rejected by LayoutNodeSchema
Split out of #197 (review thread [#discussion_r3890810822](https://github.com/tomo-chan/panemux/pull/197#discussion_r3890810822)), which found it while closing gate G3(c) but deliberately did not decide it — which side is wrong is a production question, not a fixture one.

## The mismatch

`config.LayoutNode` marks **both** `direction` and `children` `omitempty`, while `LayoutNodeSchema` requires both:

```go
// internal/config/config.go
type LayoutNode struct {
	Pane      *PaneConfig   `yaml:"pane,omitempty"      json:"pane,omitempty"`
	Direction string        `yaml:"direction,omitempty" json:"direction,omitempty"`
	Children  []LayoutChild `yaml:"children,omitempty"  json:"children,omitempty"`
}
```

```ts
// frontend/src/schemas/index.ts
export const LayoutNodeSchema = z.object({
  direction: z.enum(['horizontal', 'vertical']),   // required
  children: z.array(LayoutChildSchema),            // required
  pane: PaneConfigSchema.optional(),
})
```

So a workspace whose layout is a single root pane — `layout: {pane: {...}}`, no `direction`, no `children` — serializes to `{"pane":{...}}` and `WorkspacesResponseSchema.parse` throws on the **whole** response, not just that workspace.

## Measured, at `5bd29a3`

Go validation accepts the shape (`validateLayoutNode` permits an empty direction and no children):

```
ValidateLayout(pane-only root) = <nil>
```

Seeding such a workspace into the contract fixture, `GET /api/workspaces` emits:

```json
{"pane": {"agent_board": {}, "id": "pane-root", "type": "local"}}
```

and the frontend contract suite goes red with `invalid_type` on the missing `direction`/`children`.

## Why it matters

`useLayout` parses the whole tree and PUTs it back wholesale on any edit, so a parse failure here is not a dropped key — it is the dashboard failing to load the workspace list at all. Reachable by an operator hand-writing `config.yaml` this way; panemux's own constructors (`defaultLayout`, `singleLocalPaneLayout`) always emit `direction` + `children`, which is why nothing has hit it.

Note also that nothing in `frontend/src` reads the **root** `LayoutNode.pane` — `SplitContainer`, `App.tsx` and `useWorkspaceAttentionMonitor` all read `child.pane` off `LayoutChild`. So the field is accepted by the schema but rendered by nothing.

## Options

1. **Loosen the schema** — make `direction`/`children` optional to match Go's `omitempty`. Ripples into every `layout.children.map(...)` call site, and leaves a shape the UI still cannot render.
2. **Normalize on the Go side** — have `normalizedWorkspaces` rewrite a pane-only root into `direction` + one `children` entry, so the shape never reaches the wire. Keeps the schema strict and makes the config form the UI already supports the only one served.
3. **Reject it at validation time** — make `validateLayoutNode` require a direction and children when `pane` is unset, turning a silent dashboard failure into a startup error naming the file.

(2) looks best to me: it is the only one that both keeps existing configs working and leaves the frontend with a shape it can render. (3) is the smallest change but breaks a config that loads today.

## Not covered by #197

#197 sets `pane` **alongside** `children` on a seeded workspace, so the key reaches the fixture and a rename of its json tag now fails two checks — but the pane-only shape itself is not captured, precisely because capturing it turns the frontend suite red. Whichever option is chosen here, the fixture should then capture the real shape.

tomo-chan:
`LayoutNodeSchema` requires `direction` and `children`. Both are `omitempty` on
`config.LayoutNode`, and `validateLayoutNode` permits an empty direction and no children — so config
that Go accepts and persists could serialize to JSON the dashboard cannot read. Because
`WorkspacesResponseSchema` covers the whole response, that is **the workspace list failing to load**,
not one key going missing.

## Scope is wider than the issue said

#198 named the pane-only root. Measuring the rest found three more of the same class, each confirmed
by marshalling the shape:

| config shape | serializes as | parses? |
|---|---|---|
| `{Pane: ...}` | `{"pane":{...}}` | no direction, no children |
| `{Children: [...]}` | `{"children":[...]}` | no direction |
| `{Direction: "horizontal"}` | `{"direction":"horizontal"}` | no children — the empty slice is omitted |
| `{}` | `{}` | neither |

So the fix fills in **both** keys rather than only relocating a root pane.

## The fix

`normalizeLayoutNode` defaults an empty direction to `horizontal` and substitutes an empty array for
absent children — a nil slice marshals to `null`, which an array schema rejects as surely as an
absent key — and the JSON tags for both drop `omitempty`. **YAML keeps `omitempty`**, so `config.yaml`
stays free of empty keys; only the wire always carries them.

The root pane is **relocated into the single child it means** rather than left in place, because
nothing rendered it: `SplitContainer`, `App.tsx` and `useWorkspaceAttentionMonitor` all read
`child.pane` off a `LayoutChild`, never the root's own. A single-pane workspace written by hand
therefore displayed nothing even before the schema rejected it.

An invalid direction is **deliberately left alone** — `validateLayoutNode` owns reporting that, and
correcting it here would substitute a value the operator never wrote for an error they need to see.

## Where it applies

At `WorkspacesView` and `ActiveLayout`, and — since review — on the `PUT` path too, so a layout
written over the API is stored and echoed in the same shape a load produces. On that path the order is
`NormalizeLayout` then `ExpandLayoutPaths`, matching `finishLoad`: `ExpandLayoutPaths` walks `children`
only, so expanding first would leave a relocated root pane's `~/` cwd literal. The migration persists
on the next save, so a hand-written config converges on the `direction` + `children` form instead of
being rewritten underneath the operator on read.

## Two things this changes that are worth knowing before merging

**A root pane beside children is left exactly as written, and the response still carries it.**
Relocation applies only when the node has no children: prepending the pane to children that already sum
to 100 would rescale every sibling and surface a pane that has never rendered. `LayoutNodeSchema`
therefore keeps declaring `pane` — a key the schema does not declare is stripped by `parse()`, stored
stripped, and written back on the next split, which deletes it from `config.yaml`. That is the failure
mode `PaneConfigSchema`'s comment records `agent_board` being lost to.

**A relocated root pane is validated like any other pane, and that can stop a config that used to
load.** `validateLayoutNode` never inspected `LayoutNode.Pane` and `collectPanes` never walked it, so a
root pane was previously validated by nothing — which is also why it rendered nothing. Once it is a
`LayoutChild` it goes through `validatePane`, so a root pane with no `type`, or one of `type: ssh`
naming an undefined connection, now fails startup. Confirmed against both refs with the same config:
`main` returns `err=<nil>`, this branch returns
`invalid config: pane "main" has invalid type "": must be local, ssh, tmux, or ssh_tmux`.

This is deliberate rather than defaulted away, and the reviewer confirmed the call. The same pane
written as a *child* fails identically on `main` today (measured), there is no type-defaulting anywhere
in the loader — `defaultPaneType` appears only where the loader constructs a pane for a default config
— and defaulting here would make the root position the one place that repairs config instead of
reporting it, while still not covering the undefined-`ssh`-connection case. Note it fails the *whole*
load rather than one workspace; an invalid child pane already does exactly that on `main`, so this is
the existing failure mode rather than a new class. `docs/behavior.md` and scenario row `H14c` record it.

## Review findings, all reproduced before being fixed

Round 1:

1. **`{pane, children}` was losing the operator's root pane.** Removing `pane` from `LayoutNodeSchema`
   on the grounds that "the server cannot emit it" was wrong for this shape. Fixed by restoring `pane`
   to the schema rather than relocating or clearing it — either of those loses or moves operator data.
   The net frontend diff against `main` is now a comment and a key reorder.
2. **The relocation regression above.** Documented rather than defaulted, with the reasoning and the
   measurement that drove it; the reviewer confirmed that call.
3. **`PutLayout`/`PutWorkspaceLayout` echoed the raw decoded body** — the one `LayoutNode` in any
   response that never passed through normalization, so a node `PUT` without a direction came back as
   `"direction": ""`. Both now normalize before validating, saving and echoing.

Round 2:

4. **The fix for 3 introduced an ordering bug.** `ExpandLayoutPaths` ran before `NormalizeLayout`, and
   it walks `children` only, so a root pane relocated on the `PUT` path kept a literal `~/` cwd — which
   was echoed, persisted, and would then be a relative path for whatever started that pane's session.
   Both call sites now match `finishLoad`'s order.
5. **A vitest cache file under a repo-root `node_modules/` had been committed**, from running vitest
   with the repository root as the working directory. Untracked, and the two `.gitignore` entries
   collapsed into one `node_modules/` rule — a pattern with no internal slash matches at any depth, so
   it covers the frontend path too (verified with `git check-ignore`).

All six review threads are resolved.

## Verification

`make check` passes, as do the two pull-request-only gates run locally
(`COVERAGE_BLOCKS_BASE=origin/main make coverage-blocks`, `EFFICACY_BASE=origin/main make efficacy` —
all 15 non-exempt changed tests go red with the implementation reverted). Perturbation-checked rather
than assumed:

- reverting the pane relocation fails `TestNormalizeLayoutNode_PaneOnlyRootBecomesItsSingleChild`
- restoring the `omitempty` tags fails `TestNormalizeLayoutNode_AlwaysSerializesDirectionAndChildren`
  (on the `direction with no children` case)
- restoring the nil guard fails `TestNormalizeWorkspaceLayouts_NilBecomesAnEmptySlice`
- each review finding went red before being fixed, with the exact output quoted in its thread

**Four frontend cases carry `//efficacy:exempt`.** Three are the new `LayoutNodeSchema` round-trip
cases: they pin behavior `main` already had, because this branch's earlier head removed `pane` and the
fix restores it, leaving no implementation for a revert to take away. They *do* fail against that
earlier head, which is the regression they exist to prevent. The fourth,
`survives a full layout parse`, this branch does not change at all — the gate attributes it because the
describe block added at the end of the file follows it.

## Relationship to #197

Capturing the pane-only shape in the API contract fixture belongs with #197, whose branch owns
`testdata/api-contract` — it is done there and that branch needs a rebase onto this one once both land.
This PR is deliberately free of fixture changes so it reviews as the production fix it is. Note that
findings 3 and 4 above would have surfaced in #197's fixtures as a captured `PUT` response the frontend
rejects, so closing them here removes work from that rebase.

Closes #198

https://claude.ai/code/session_01Veqb91P9norzcFVuovCQE3


internal/config/config.go
```diff
@@ -59,10 +59,64 @@ type PaneConfig struct {
...
+// frontend/src reads it: SplitContainer, App.tsx and useWorkspaceAttentionMonitor
+// all read child.pane off LayoutChild, never the root's own. A single-pane
+// workspace written by hand as `layout: {pane: ...}` therefore rendered
+// nothing even before the schema rejected it.
+//
+// An invalid direction is deliberately left alone — validateLayoutNode owns
+// reporting that, and correcting it here would substitute a value the
+// operator never wrote for an error they need to see.
+func normalizeLayoutNode(node LayoutNode) LayoutNode {
+	if node.Pane != nil && len(node.Children) == 0 {
```
tomo-chan, line 91: **Root pane + children still serializes `pane`, and dropping it from `LayoutNodeSchema` now silently deletes it from `config.yaml`.**

The guard is `node.Pane != nil && len(node.Children) == 0`, so the `{pane, children}` shape is left alone. `validateLayoutNode` never inspects `node.Pane`, so that shape is accepted config. Confirmed by marshalling the normalized node:

```
{"pane":{"id":"root",...},"direction":"horizontal","children":[{"pane":{"id":"a",...},"size":100}]}
```

So "the server cannot emit it" — the justification given for removing `pane` from `LayoutNodeSchema` — does not hold for this shape. And because `pane` is now an *undeclared* key, `WorkspacesResponseSchema.parse()` strips it rather than rejecting: `useLayout` stores the stripped tree in state and the next split/move/debounced resize `PUT`s it back verbatim, `SaveWorkspaces` persists it, and the root pane is gone from the user's config file for good.

That is precisely the failure mode the `PaneConfigSchema` comment two blocks up warns about ("a field missing from this schema is silently deleted from the user's config.yaml by an unrelated action. agent_board was lost exactly that way").

Suggest relocating the root pane unconditionally (prepend it as a child when children exist), or at minimum clearing it so it cannot be emitted — plus a test case for `{pane, children}`, which the table in `TestNormalizeLayoutNode_AlwaysSerializesDirectionAndChildren` does not currently cover.

---

internal/config/config.go
```diff
@@ -59,10 +59,64 @@ type PaneConfig struct {
...
+		node.Children = []LayoutChild{{Size: 100.0, Pane: node.Pane}}
```
tomo-chan, line 92: **Relocating the root pane turns a previously-loadable config into a startup failure.**

`finishLoad` runs `normalizeWorkspaces()` *before* `Validate()`. Until now a root `pane` was validated by nothing (`validateLayoutNode` ignores `node.Pane`) and collected by nothing (`collectPanes`/`panesInLayout` only walk `Children`), so it was inert. After this change it becomes a `LayoutChild` and goes through `validatePane`.

Reproduced against both refs with the same config:

```yaml
workspaces:
  active: main
  items:
    - id: main
      title: Main
      layout:
        pane:
          id: main
          shell: /bin/bash   # no `type:`
```

- `main`: `Load` returns `err=<nil>`
- this branch: `invalid config: pane "main" has invalid type "": must be local, ssh, tmux, or ssh_tmux` → panemux refuses to start

Same for a root pane of `type: ssh` whose `connection` isn't defined in `ssh_connections`. The affected population — hand-written single-pane workspaces — is exactly the one this PR targets, and going from "starts, renders nothing" to "won't start at all" is a worse outcome for them than the bug being fixed. Worth either defaulting the relocated pane's `type` the way the rest of the loader defaults things, or at least calling the new hard failure out in `docs/behavior.md` and the PR body.

---

internal/config/config.go
```diff
@@ -237,6 +291,10 @@ func (c *Config) normalizedWorkspaces() WorkspacesConfig {
 	if workspaces.Active == "" && len(workspaces.Items) > 0 {
 		workspaces.Active = workspaces.Items[0].ID
 	}
+	// Every JSON response carrying a LayoutNode comes from here or from
+	// ActiveLayout, so normalizing at both is what makes issue #198's
```
tomo-chan, line 303: **"Every JSON response carrying a `LayoutNode` comes from here or from `ActiveLayout`" isn't quite true.**

`PutLayout` (`internal/api/handler.go:225`) and `PutWorkspaceLayout` (`:382`) both end in `writeJSON(w, layout)` — echoing the *decoded request body*, which never passes through `normalizeLayoutNode`.

Concretely: a client `PUT`s `{"children":[{"size":100,"pane":{...}}]}` with no direction. `ValidateLayout` accepts an empty direction, and now that the JSON tag has lost `omitempty` the echo is `{"direction":"","children":[...]}`, which `LayoutNodeSchema`'s `z.enum(['horizontal','vertical'])` rejects.

Latent today only because `useLayout.saveLayout` discards the response body, but it's a hole in the invariant this branch exists to establish — and #197's contract fixtures would trip over it. One line each (`writeJSON(w, normalized)`) closes it.

---

tomo-chan (commented):
Reviewed the diff for correctness. The core fix (always emitting `direction` + `children`, relocating a pane-only root) is sound and the test coverage is good. Three findings, all anchored inline — the first is the one I'd want addressed before merge, since it is silent config data loss in the same class the `PaneConfigSchema` comment already documents.

---

internal/config/config.go
tomo-chan: Confirmed and fixed in 064ad01 — but with a third option rather than either of the two you suggested, so the reasoning is worth stating.

Reproduced first. `normalizeLayoutNode({pane: root, children: [a]})` marshals to exactly what you quoted:

```
{"pane":{"id":"root",...},"direction":"horizontal","children":[{"pane":{"id":"a",...},"size":100}]}
```

And the frontend half reproduced too, before touching anything: `LayoutNodeSchema.parse(withRootPane)` came back without `pane`, so the round-trip test went red on the branch as it stood. So both halves of your finding hold, and the PR body's "the server cannot emit it" was simply wrong.

**Why not relocate unconditionally.** The children in that shape already sum to 100, and `validateChildren` requires that (±0.1). Prepending the root pane means rescaling every sibling — silently changing proportions the operator wrote — and surfacing a pane that has never rendered. That is a bigger behavior change than the bug.

**Why not clear it.** Clearing is the same deletion you identified, just performed deliberately by us instead of silently by Zod. The operator still loses the pane from `config.yaml` on the next save.

**What I did instead: restore `pane` to `LayoutNodeSchema`.** Your finding shows the schema's job here is not only "what the dashboard renders" but round-trip fidelity — which is exactly what the `PaneConfigSchema` comment is about. A schema that omits a key the server can emit is a schema that is wrong about the server, and the `{pane, children}` shape is left exactly as the operator wrote it. The net frontend diff against `main` is now a comment and a key reorder; the branch's earlier head was the regression.

Tests added, both the ones you asked for and the frontend side:

- `TestNormalizeLayoutNode_KeepsARootPaneThatSitsBesideChildren` — pins that the pane survives *and* that siblings are not rescaled, so a later "cleanup" cannot quietly clear it
- `LayoutNodeSchema root pane round-trip` (3 cases) — accepts it, `parse()` does not strip it, and a node without a root pane still parses

One consequence worth flagging: because the frontend is now behaviorally identical to `main`, those three cases survive `make efficacy` and carry `//efficacy:exempt` with that reason. They fail against this branch's *previous* head, which is the regression they exist to prevent — main just never had it.

`docs/behavior.md` and scenario row `H14b` record the shape; PR body corrected.

---

internal/config/config.go
tomo-chan: Reproduced exactly as you describe, with your config, against both refs:

- `main`: `Load` → `err=<nil>`
- this branch: `invalid config: pane "main" has invalid type "": must be local, ssh, tmux, or ssh_tmux`

(My first run also tripped `server.port 0 is out of range` because I omitted a `server:` block; with one added the pane error is the only difference between the refs.)

I took the second of your two options — document it — rather than defaulting the type. One measurement drove that. The same pane written as a child fails identically on `main` today:

```
CHILD-PANE-NO-TYPE on main: err=invalid config: pane "main" has invalid type "": must be local, ssh, tmux, or ssh_tmux
```

There is no type-defaulting anywhere in the loader: `defaultPaneType` appears only where the loader *constructs* a pane for a default config, never where it reads one. So the root pane was not benefiting from a defaulting rule — it was the one position no validation reached, which is the same reason it rendered nothing.

Defaulting only in the relocation path would also contradict the principle this branch already states two tests up, in `TestNormalizeLayoutNode_DoesNotRepairAnInvalidDirection`: normalization does not substitute values the operator never wrote, because that turns a config error into a value they need to see and didn't write. And it would not close the case you raise second — a root pane of `type: ssh` naming an undefined connection still fails, and no default can fix that one.

On the product judgment, I read it the other way round: "starts, renders nothing" is a silent failure the operator has to debug from scratch, while the error names the pane and the exact fix. Their config was already broken; the difference is whether panemux says so.

So it is now documented rather than silent, in three places:

- `docs/behavior.md` — a new paragraph under `GET /api/layout` stating that a relocated root pane is validated like any other, naming both failure cases and why it is deliberate
- scenario row `H14c`
- `TestLoad_RelocatedRootPaneIsValidatedLikeAnyOther` (asserts the startup failure) and `TestLoad_WellFormedRootPaneRelocatesAndLoads` — the second exists so the first cannot be satisfied by normalization breaking every root pane
- PR body

If you'd rather have the defaulting despite the above, say so and I'll add it — it's a small change, and it's your call, not mine.

---

internal/config/config.go
tomo-chan: Correct, and fixed. The test went red with exactly the body you predicted:

```
{"direction":"","children":[{"pane":{"id":"one-main","type":"local","agent_board":{}},"size":100}]}
```

Both handlers now normalize, via a new exported `config.NormalizeLayout`. One deviation from your `writeJSON(w, normalized)` suggestion: I normalize once immediately after `ExpandLayoutPaths`, *before* `ValidateLayout`, and reassign `layout` — so the same node is what gets validated, saved and echoed, rather than the echo being a separately-normalized copy of something else. That means a `PUT` also *stores* the normalized shape, and a pane-only root arriving over the API is migrated the same way one loaded from disk is, instead of the two paths disagreeing about what a stored layout looks like. It also puts a relocated root pane through `validatePane` on the PUT path, matching load.

Tests:

- `TestPutLayoutRoutes_EchoANormalizedNode` — table over both routes (folded into one table because `dupl` flagged the two separate functions as duplicates)
- `TestPutLayout_RelocatesAPaneOnlyRoot` — asserts the *stored* config, not just the echo, so the normalization can't drift back to being echo-only

Scenario row `H14d`, and `docs/behavior.md` now says so under `PUT /api/layout`.

Also confirming your last point, since it changes what #197 has to do: yes, #197's contract fixtures would have tripped over this. That branch captures a `PUT` response among its fixtures, so this would have surfaced there as a fixture the frontend rejects — one more reason it's better closed here than at rebase time.

---

internal/api/handler.go
```diff
@@ -209,6 +209,11 @@ func (h *Handler) PutLayout(w http.ResponseWriter, r *http.Request) {
 	}
 
 	config.ExpandLayoutPaths(&layout)
+	// Normalize before validating, saving and echoing, so a PUT stores and
+	// returns exactly the shape a load would have produced. Without this the
+	// echo is the raw request body — the one LayoutNode in an API response
+	// that never passed through normalizeLayoutNode.
+	layout = config.NormalizeLayout(layout)
```
tomo-chan, line 220: **A root pane relocated on the PUT path keeps an unexpanded `~/` cwd — the two calls are in the opposite order from the load path.**

`ExpandLayoutPaths` walks `layout.Children` only (`config.go:575-577`); it never touches `layout.Pane`. So expanding *before* `NormalizeLayout` relocates the root pane means the pane is still at the root when expansion runs, and by the time it becomes a `LayoutChild` expansion is over. `finishLoad` does it the other way round — `normalizeWorkspaces()` (line 215) then `expandPaths()` (line 217) — which is why the config-file path is correct and this one is not.

Reproduced by driving exactly these two calls in this order:

```
PUT (root pane)  -> {"direction":"horizontal","children":[{"pane":{"id":"root","type":"local","cwd":"~/work",...},"size":100}]}
PUT (child pane) -> {"direction":"horizontal","children":[{"pane":{"id":"root","type":"local","cwd":"/root/work",...},"size":100}]}
LOAD order       -> {"direction":"horizontal","children":[{"pane":{"id":"root","type":"local","cwd":"/root/work",...},"size":100}]}
```

`ValidateLayout` passes it, so the literal `~/work` is echoed to the client, handed to `SaveLayout`, and written to `config.yaml` — where it is then a relative path for whatever starts that pane's session.

This is reachable only because of this PR: before it, a root pane on the PUT path stayed a root pane, was walked by nothing, and started no session. Promoting it into a live child is what makes the missing expansion observable, so it belongs with the relocation rather than being pre-existing.

Swapping the two lines fixes it and matches `finishLoad`:

```go
layout = config.NormalizeLayout(layout)
config.ExpandLayoutPaths(&layout)
```

Worth a test case alongside the existing PUT ones: a `{"pane":{...,"cwd":"~/x"}}` body whose response and persisted config both carry the expanded path. Same fix needed at line 375 in `PutWorkspaceLayout`.

---

internal/api/handler.go
```diff
@@ -363,6 +368,11 @@ func (h *Handler) PutWorkspaceLayout(w http.ResponseWriter, r *http.Request) {
...
```
tomo-chan, line 384: Same ordering bug as `PutLayout` above — `ExpandLayoutPaths` runs before the relocation, so a root pane's `~/` cwd reaches `SaveWorkspaceLayout` unexpanded. Both call sites need the two lines swapped.

---

node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json
```diff
@@ -0,0 +1 @@
+{"version":"4.1.11","results":[[":frontend/src/schemas/index.test.ts",{"duration":53.81904900000001,"failed":false}]]}
```
tomo-chan: **A vitest cache file under `node_modules/` was committed by accident** (added in `081b240`):

```json
{"version":"4.1.11","results":[[":frontend/src/schemas/index.test.ts",{"duration":53.8...,"failed":false}]]}
```

The reason it slipped through is that `.gitignore` line 3 ignores `frontend/node_modules/` but there is no entry for a repo-**root** `node_modules/`, which is where this path lives. So the gap will keep producing untracked-then-committed files on every branch, not just this one.

Two things:

1. `git rm --cached` the file (or drop it from the branch's history).
2. Add a root `node_modules/` entry to `.gitignore` in the same change — DEVELOPMENT.md's "Ignore generated resources" rule asks for exactly this ("Do not leave new build or release byproducts as recurring untracked files").

Unrelated to the layout fix, but it should not land with it.

---

tomo-chan (commented):
Re-reviewed at `064ad01`. The three findings from the earlier round are genuinely closed — `pane` is back in `LayoutNodeSchema` (so `{pane, children}` no longer loses the operator's pane), both PUT handlers now normalize, and the relocation regression is documented in `docs/behavior.md` with scenario row `H14c` rather than defaulted away. `go test ./internal/config/ ./internal/api/` passes.

Two things left, one of them introduced by the fix for finding 3:

1. **`internal/api/handler.go:216` and `:375` — `ExpandLayoutPaths` runs *before* `NormalizeLayout`, so a root pane relocated on the PUT path keeps its literal `~/` cwd.** `ExpandLayoutPaths` walks `Children` only, so the pane is still at the root when expansion runs. `finishLoad` orders these the other way (`normalizeWorkspaces()` then `expandPaths()`), which is why the config-file path is correct. Reproduced against both orderings; the fix is swapping the two lines. Reachable only because of this PR — before it a root pane on the PUT path was never promoted into a live pane.

2. **A vitest cache file under `node_modules/` was committed in `081b240`.** `.gitignore` ignores `frontend/node_modules/` but not a repo-root `node_modules/`, so this will recur on other branches too.

Checked and cleared: load ordering (`finishLoad` normalizes before expanding and before validating); `SaveLayout`/`SaveWorkspaces` both go through the mutating `normalizeWorkspaces`, so the migration really does persist; `LayoutChildSchema` keeps `direction`/`children` optional, so leaving nested nodes un-normalized is consistent rather than a second instance of #198; `Size: 100.0` against the sum-to-100 rule; and the claim that no frontend call site reads the root's own `pane` (`SplitContainer`, `App.tsx`, `useWorkspaceAttentionMonitor` all read `child.pane`).

---

internal/api/handler.go
tomo-chan: Confirmed and fixed in 0f1d5fa. This one is mine — introduced by my own fix for finding 3, exactly as you say.

Reproduced before touching the code, with a test that drives the real router rather than the two calls directly, so the echo and the persisted config are both observed:

```
Error: "{"direction":"horizontal","children":[{"pane":{"id":"root","type":"local","cwd":"~/work","agent_board":{}},"size":100}]}"
       should not contain "~/work"
Messages: the echo carries a literal ~/

Error: Not equal:
       expected: <temp home>/work
       actual  : "~/work"
Messages: a literal ~/ persisted here is a relative path to whatever starts the session
```

Both call sites now read exactly as you wrote it, and the comment above them says *why* the order matters rather than just stating it, so the next person to add a step here has the reason in front of them:

```go
layout = config.NormalizeLayout(layout)
config.ExpandLayoutPaths(&layout)
```

`TestPutLayoutRoutes_ExpandARelocatedRootPaneCwd` is a table over both routes (one function, since `dupl` rejects two near-identical ones — same reason the echo test is a table), and it asserts the *stored* layout as well as the echo, so an echo-only fix would not satisfy it. Scenario row `H14e`; `docs/behavior.md`'s `PUT /api/layout` section now names the ordering and `ExpandLayoutPaths`'s children-only walk as the reason for it.

I agree with your framing that it belongs here rather than being pre-existing: nothing before this branch promoted a root pane into a live child, so there was no path for the missing expansion to be observed on.

`make check`, `coverage-blocks` and `efficacy` (15/15 red) all pass on the new head.

---

internal/api/handler.go
tomo-chan: Fixed at this call site too, in 0f1d5fa — the swap was applied to both, and `TestPutLayoutRoutes_ExpandARelocatedRootPaneCwd` covers `PUT /api/workspaces/{id}/layout` as its second table row, asserting `cfg.Workspaces.Items[0].Layout` rather than only the response.

Details on the thread above.

---

node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json
tomo-chan: Both done in 0f1d5fa. The file came from my running `npx vitest` with the repository root as the working directory instead of `frontend/`, which put vitest's cache at the root — and your diagnosis of why it slipped through is right: `git ls-files | grep node_modules` returned that one path, and `.gitignore` had no root entry to catch it.

`git rm --cached` rather than a history rewrite: the file carries no path from a developer machine (its only content is a hash-named directory and a test duration), so DEVELOPMENT.md's path-sanitization rule does not require rewriting the branch, and a rewrite would invalidate your review anchors on a PR you are mid-review of.

One deviation from your suggestion, since it changes a line you did not ask me to touch: rather than *adding* a root entry alongside `frontend/node_modules/`, I collapsed the two into a single `node_modules/`. A gitignore pattern with no internal slash matches at any depth, so the root rule already covers the frontend path and keeping both would leave a line that can never match anything the other does not. Verified rather than assumed:

```
$ git check-ignore -v frontend/node_modules/vitest/foo
.gitignore:3:node_modules/	frontend/node_modules/vitest/foo
```

Say the word if you would rather keep the explicit `frontend/` line for readability and I will put it back.

---

internal/config/config.go
tomo-chan: Reviewer's call, since this one was left to me: **keep it documented, don't add the defaulting.** Your reasoning holds, and the measurement is what settles it — a child pane with no `type` already fails identically on `main`, so defaulting in the relocation path would make the root position the one place in the loader that repairs config instead of reporting it, which is the same objection `TestNormalizeLayoutNode_DoesNotRepairAnInvalidDirection` already pins. And it wouldn't close the `ssh` case anyway.

The one thing I weighed against it after filing: this fails the *whole* load, not just the broken workspace, so other workspaces go down with it. But an invalid child pane already does exactly that on `main` — it's the existing failure mode, not a new class — so consistency wins over carving out an exception.

Verified on `0f1d5fa`: `TestLoad_RelocatedRootPaneIsValidatedLikeAnyOther` and `TestLoad_WellFormedRootPaneRelocatesAndLoads` both pass, `docs/behavior.md:234-242` states the failure and why it's deliberate, and `H14e` joins `H14b`/`H14c`/`H14d` in the ledger (`make check-scenarios`: 155 names across 85 auto rows resolve).

Round 2 both confirmed by reproduction on my side as well: the ordering swap goes red when reverted (`expected /tmp/…/work, actual "~/work"` on the `workspace layout` row), and `git check-ignore -v` reports the collapsed `node_modules/` rule matching `frontend/node_modules/` as you said. `git ls-files | grep -c node_modules` → 0. Collapsing the two `.gitignore` lines is the right call — keep it.

No further findings. Resolving.

---

PR #201, opened 2026-09-03
refactor: make the session replay buffer a ring (#193)

tomo-chan:
`managedSession.publish` is the path every byte a pane produces travels,
and retaining a fixed 256KB replay window meant reallocating and copying
the whole window on every chunk. A pane open for more than a few seconds
is in that state permanently, so it paid a full window copy per 4KB of
output, forever: `make bench` measured ~598KB and two orders of magnitude
of time per chunk against the same call on a cold buffer.

Replace the slice with a ring. `replayBuffer` writes into storage it
already owns and moves two indices, so the steady-state cost is the
length of the chunk and the steady-state allocation is nothing at all.
Storage doubles from 4KB up to the limit rather than being claimed up
front, so a pane that has barely spoken is not charged a whole window —
a workspace can hold dozens of them.

Measured back to back on one container, 4KB chunk, no subscribers,
median of five: 222,527 ns/op and ~598,000 B/op before, 117 ns/op and
0 B/op after. Cold and full are now within ~2x of each other rather
than two orders of magnitude apart, which is the issue's completion
condition.

What keeps it from coming back is a test rather than those numbers:
`TestManagedSession_Publish_SteadyStateDoesNotCopyTheReplayWindow`
asserts zero allocations per steady-state publish, and
`TestReplayBuffer_SteadyStateAppendDoesNotAllocate` does the same one
layer down. An allocation count has none of the 2.9x run-to-run spread
docs/quality-gateway.md warns about for timings, so it can be asserted
in a hermetic unit test where a duration could not.

The rest of the new tests pin what the ring must not change: which
bytes are retained across wrap-around, against the naive "tail of
everything written" model the slice was, and that a snapshot handed to
a subscriber is its own copy rather than a view onto storage later
output overwrites in place.

Also records the second, inverted conclusion in quality-gateway.md.
With the window copy gone, subscriber fan-out is no longer cheap in
comparison — it is now the whole cost, ~2us per subscriber per 4KB
chunk, close to linear. That is left alone deliberately.

Claude-Session: https://claude.ai/code/session_01SDVnogNvw21GDeawq81iaE


internal/session/manager_bench_test.go
```diff
@@ -110,18 +127,20 @@ func BenchmarkSessionPublish(b *testing.B) {
 // iteration cost ~95µs of pause against a ~2µs operation, so the benchmark
 // spent 50s of wall clock on 1s of measured work — and, worse, the repeated
 // pauses reset the GC and allocator state this benchmark exists to measure,
-// reporting ~5,400 ns/op for a path that actually costs ~1,900. `entry.history
-// = nil` is a single pointer store, a constant well under 1% of the operation,
-// so timing it is far cheaper than excluding it.
+// reporting ~5,400 ns/op for a path that actually costs ~1,900. The reset is
+// two index stores — it kept the whole window's storage rather than dropping
+// it — a constant well under 1% of the operation, so timing it is far cheaper
+// than excluding it.
```
tomo-chan, line 133: Minor factual inaccuracy in the comment: the reset here never "kept the whole window's storage". Because `reset()` runs every iteration, `reserve` is only ever asked for 4096 bytes, so this benchmark's `replayBuffer.buf` stays at 4KB and never approaches the 256KB window. Measured: `-benchtime 2000x` reports `2 B/op` (= one 4096-byte allocation amortized over 2000 iterations), not the 256KB the sentence implies.

Nothing misbehaves — the benchmark still measures the intended "no eviction" path — but the comment is what a reader uses to interpret the cold-vs-full pair in `docs/quality-gateway.md`, and it overstates what the after-column's `0 B/op` is holding onto. Suggest "kept its storage" rather than "the whole window's storage".

---

tomo-chan (commented):
Reviewed the ring buffer for correctness. I verified `replayBuffer` against the naive "last `limit` bytes of everything written" model with a throwaway differential test (limits 1, 3, 100, 4095, 4096, 4097, 10000, 65536; 30 seeds x 300 random chunk sizes spanning 0..2*limit) — no mismatch, and `capacity()` never exceeded the limit in any run.

The invariant the design rests on holds: eviction can only fire when `want == limit`, which forces `reserve` to have sized storage to exactly `limit`, so `start != 0` implies `len(buf) == limit` — which is why `reserve`'s single-copy `r.buf[r.start:r.start+r.size]` can never go out of range. Each `//mutation:exempt equivalent` claim checks out individually, including the `len(chunk) >= r.limit` one (at equality the general path writes the same bytes and lands `start` at the same offset).

One low-severity comment inaccuracy noted inline. No correctness findings.

---

PR #202, opened 2026-09-03
test: freeze the a11y violation counts as a ceiling (#194)

### LINKED ISSUE #194: test: a11y の現状値を上限として凍結する（下がってよいが上がってはいけない）
#180 の項目 7（[#187](https://github.com/tomo-chan/panemux/pull/187)）が `frontend/e2e/a11y.spec.ts` を追加し、axe-core の違反を**記録**するところまでやった。項目 7 は計測のみが明示された段階だったので、閾値は決めていない。データが揃ったので決められる。

## 現状値（#187 で記録）

| ページ状態 | critical | serious | moderate |
|---|---|---|---|
| ダッシュボード | 1 (`aria-required-children`) | 1 (`color-contrast`, 2 nodes) | 1 (`region`, 7 nodes) |
| ペイン設定ダイアログ表示 | 2 (`+ select-name`) | 1 (`color-contrast`, 10 nodes) | 1 (`region`, 7 nodes) |

`.xterm` は除外している。xterm.js が所有する canvas なので、監査しても他人のマークアップについて毎回報告するだけで、こちらに対処のしようがない。

## やること

現状値を**上限**として固定する。下回るのは歓迎、上回ったら fail。

既存の UI に axe をかけると必ず backlog が出る。**最初から red のゲートは初日に迂回される**（設計原則 4）ので、`0` を目標にするのではなく現状値を天井にするのが正しい形。これは #188 が per-block カバレッジで下した判断（275/1801 に対して全体ゲートを諦めて差分スコープにした）と同じ考え方の別の解き方で、こちらは差分スコープが取れない代わりに ceiling が取れる。

決めるべきこと:

- **粒度** — 件数だけか、ルール ID ごとか。ルール ID ごとのほうが「1 つ直して 1 つ増えた」を捕まえられるが、脆くもなる
- **node 数を数えるか** — `color-contrast` は 2 nodes と 10 nodes で、同じ 1 件でも中身が違う
- **天井を下げる運用** — 直したら天井も下げないと、次の退行を許してしまう。手動更新か、自動で下げるか

`docs/quality-gateway.md` の「First measurements」に現状値があるので、天井の初期値はそこから取れる。

## 完了条件

- [x] 現状値が上限として表明されている
- [x] 違反を 1 件増やすと E2E が落ちる（摂動で確認すること）
- [x] 違反を減らしたときに天井を下げる手順が書かれている
- [x] `docs/quality-gateway.md` の G5 行と「First measurements」を更新する
- [x] `make check` / `make test-e2e` が通る

---

## 結果

**完了**（#202 `4f158bc`）。天井は `frontend/e2e/a11y-ceiling.ts` の `CEILINGS`、比較器は同ファイルの `checkAgainstCeiling`、駆動は `frontend/e2e/a11y.spec.ts`。

### 3 つの論点への答え

| 論点 | 決定 | 理由 |
|---|---|---|
| 粒度 | **ルール ID ごと** | 合計だけだと「1 つ直して 1 つ増えた」が数値に出ない。ゲートが最も捕まえるべき形がちょうどこれ |
| node 数 | **数える** | `color-contrast` は 2 nodes と 10 nodes で別の事実。ダイアログ 1 個分の低コントラストがダッシュボードに出ても「変化なし」になってしまう |
| 天井を下げる運用 | **手動編集。ただし実行時に貼り付ける行をそのまま出力する** | 本 issue のタイトルが「下がってよい」と明言している。改善でビルドが壊れると改善しなくなるし、下振れも fail にすると脆さが**両側**の flake になる |

マップに無いルールは天井 0 なので、**新種の違反は出た日に落ちる**。

天井の値は #187 の記録どおりで変更なし。別ハードウェア・別コンテナで再測定して同一値だったので、そのまま凍結した。

### 途中で見つかった偽陽性 — 決め方より重要だった

最初の版は**単体では通るのに `make test-e2e` 全体では落ちた**。node 数はレンダリング結果の性質であり、共有 E2E サーバ上で `pane-move.spec.ts` の split が先に走ると、同じページが `region` 7→10、`color-contrast` 10→11 になる。つまり**スイートの実行順で落ちるゲート**で、設計原則 4 の偽陽性そのものだった。

`core-multiplexer.yml` が「レイアウトを変更する側」として分離されているのと同じ理由（逆側）で、a11y スペックに専用の panemux プロセスとフィクスチャ（`frontend/e2e/a11y.yml`、port 4178）を与えて解決。一般則として D11 に記録した — **ページ上のものを数えるゲートは、そのページを所有する必要がある。**

### レビューで出た 2 点

- **失敗メッセージが axe のエントリ単位で重複しうる。** カウントはマップに集約しているのに失敗ループは生の `violations` を回していたため、1 ルールが 2 エントリで来ると「正しい合計値」と「同一メッセージ 2 本」になる。加えて `impact` をたまたま乗っていたエントリから読むので、深刻度が食い違う 2 本が出る可能性もあった。両ループともマップを読むよう修正。
- **比較器に単体テストが無い。** これは ratchet 全般が持つ形で、**健全なリポジトリでは `actual === allowed` しか通らず、4 つの重要な分岐が一度も走らない** — 反転はグリーンのまま出荷され、後で「検知されなかった退行」として現れる。比較器を別モジュールに切り出し、`frontend/e2e/a11y-ceiling.test.ts` に 13 ケース（ブラウザ不要）を追加した。

なお `frontend/e2e/` を 2 つのランナーで分けることは無料ではない。Playwright のデフォルト `testMatch` は `*.test.ts` も拾い、vitest ファイルを読ませると `@vitest/expect` が投げて **E2E スイート全体**が死ぬ（1 ファイルの失敗では済まない）。両 config で suffix による境界を明示してある。

### 天井を下げるとき

1. `make test-e2e` を実行する。天井を下回ったルールは `a11y[<page state>] ceiling can be lowered` として**貼り付ける行そのもの**を出力する（`set CEILINGS['pane-settings']['color-contrast'] to 9 (was 10)`、ルールが消えたなら `remove '...'`）。同じ内容が `axe-<page state>.txt` の添付にも入る。
2. `frontend/e2e/a11y-ceiling.ts` の `CEILINGS` と、`docs/quality-gateway.md` の「Accessibility」表の両方に、修正と同じコミットで反映する。

どちらの場合も実行は pass する。**手順 2 を強制するものは無い**（意図的、D11）ので、直して手順 2 を飛ばすと次の退行の余地が黙って残る。

### ドキュメント

- `docs/quality-gateway.md` — G5 行と interaction capability 行をゲートする旨に更新、**決定 D11** を追加、「First measurements」の Accessibility を天井テーブル＋引き下げ手順に書き換え
- `docs/scenarios.md` — 「性能と a11y は計測のみ」の注記を分割（性能は計測のみ、a11y は天井で保持）
- `DEVELOPMENT.md` — Quality gate 一覧に a11y ceiling の行を追加

## 関連

- #180 — 項目 7（計測のみの段階）
- #187 — a11y スキャンの追加と現状値
- #202 — 本 issue の実装（マージ済み、`4f158bc`）

tomo-chan:
Closes #194.

`a11y.spec.ts` recorded axe violations without asserting on them (#187), deliberately — item 7 of #180 was the measurement step, and a threshold guessed at before the data existed is the gate that starts red and gets routed around (design principle 4). The data exists now, so it becomes the ceiling.

## The three decisions #194 asked for

| Question | Decision | Why |
|---|---|---|
| Granularity — total or per rule ID? | **Per rule ID** | A total alone lets one violation be fixed and another introduced in the same change without the number moving — the "fixed one, added one" case, and the most likely way this UI regresses while looking maintained. |
| Count nodes? | **Yes** | `color-contrast` is one violation at 2 nodes on the dashboard and one at 10 with the dialog open. Counting rules alone treats those as the same fact, so a dialog's worth of unreadable text could appear on the dashboard and register as no change. |
| Lowering the ceiling — manual or automatic? | **Manual, with the run printing the exact replacement line** | The issue title says a count may fall. Failing a run for *improving* accessibility trains people to stop improving it, and it turns every brittleness risk into a two-sided flake instead of a one-sided one. |

A rule absent from `CEILINGS` has a ceiling of zero, so a **new kind** of violation fails the run the day it appears. Every listed count is a violation this repository still has; listing it is not approval of it.

## The false positive found on the way

The first version passed when run alone and **failed inside `make test-e2e`**. A node count is a property of whatever happens to be rendered, and on the shared E2E server `region` and `color-contrast` read 7 and 10 alone but **10 and 11** once `pane-move.spec.ts`'s splits had run first. The gate was failing on the order of the suite rather than on anything about accessibility.

Fixed the way `core-multiplexer.yml` already is, from the other side — that spec is separated because it mutates the layout, this one because it is the most sensitive to mutators. `a11y.spec.ts` gets its own panemux process and fixture (`frontend/e2e/a11y.yml`, port 4178), so the audited page is pinned rather than inherited. The general rule, recorded in D11: **a gate that counts what is on a page needs to own the page.**

## The ceiling

| Page state | Rule | Impact | Ceiling (nodes) |
|---|---|---|---|
| Dashboard | `aria-required-children` | critical | 1 |
| Dashboard | `color-contrast` | serious | 2 |
| Dashboard | `region` | moderate | 7 |
| Pane settings dialog open | `aria-required-children` | critical | 1 |
| Pane settings dialog open | `select-name` | critical | 1 |
| Pane settings dialog open | `color-contrast` | serious | 10 |
| Pane settings dialog open | `region` | moderate | 7 |

Unchanged from what #187 recorded, reproduced on different hardware in a different container.

## Review round (a8f78bb)

Two findings, both fixed:

- **The failure loop iterated axe's raw entries while the counting summed into a map.** Two entries for one rule gave one right count and two identical messages — and, less visibly, two messages that could disagree about `impact`, each read from whichever entry the loop was on. Both loops now read the map.
- **The comparator had no test.** It has the shape every ratchet has: on a healthy repository observed equals ceiling, so a green run takes only `actual === allowed` and none of the four interesting branches ever executes — an inversion in one of them ships green and surfaces later as a *missed* regression. `checkAgainstCeiling` and `CEILINGS` moved to `frontend/e2e/a11y-ceiling.ts`, unit-tested by `frontend/e2e/a11y-ceiling.test.ts` (13 cases, no browser needed).

That splits `frontend/e2e/` between two runners. Playwright's default `testMatch` claims `*.test.ts` too, and letting it load a vitest file throws in `@vitest/expect` and takes the **whole** E2E suite down, not one file — so both configs now draw the line by suffix and say why.

## Merge with main (3addb44)

#201 (ring replay buffer) and this branch both rewrote the same bullet in `docs/scenarios.md`. Both changes are kept — the performance bullet carries #201's exception about the replay buffer's allocation-count assertion, and the accessibility bullet stays separate because that half stopped being "measured, not verified".

One thing git could not see, so it is called out rather than left in the diff: #201's new note under "First measurements" says the relay and accessibility figures "have not been retaken". True when written, no longer true of the accessibility half — those counts were retaken on a second container, came back identical, and are what this PR freezes. The note now says so, and says why the container-noise caution does not apply to violation counts at all.

## Docs

- `docs/quality-gateway.md` — G5 and the interaction-capability row now say the scan gates; **decision D11** records the three choices above, the false positive that shaped them, and why the comparator is separated and unit tested; "First measurements" carries the ceiling table plus the procedure for lowering it.
- `docs/scenarios.md` — the "performance and accessibility are measured, not verified" note splits: performance still only reports, accessibility is now held at a ceiling.
- `DEVELOPMENT.md` — the quality-gate list gains the accessibility ceiling row.

No `frontend/src`, `internal/api` or `internal/config` changes, so G0 does not apply — `scenarios.md` is updated anyway because the note in it had gone stale.

## Test plan

- [x] `make check` passes (exit 0) — 33 frontend test files / 758 tests, up from 32 / 745. Re-run after the merge with main.
- [x] `make test-e2e` passes — 30 tests, including both a11y ceiling tests on the new `chromium-a11y` project. Re-run after the merge; the ceiling still holds.
- [x] `npx vitest run e2e/a11y-ceiling.test.ts` — 13 passed
- [x] **Perturbation: adding a violation fails the run.** An `img` with no `alt` and a low-contrast `span` added to `TerminalPane` failed both scans, exercising both branches — `image-alt` as a new rule at a ceiling of zero, and `region` risen from 7 nodes to 9. Reverted; the diff contains no `frontend/src` change.
- [x] **Perturbation: the new unit test goes red without its fix.** Reverting the failure loop to iterating axe's raw entries fails exactly the duplicate-entry case (1 failed / 12 passed). Done by hand because `make efficacy` skips this branch — "no implementation changed against origin/main", since everything here is test-side.
- [x] Perturbation: the *previous* shared-server version was confirmed to fail inside the full suite and pass alone, which is what motivated the dedicated fixture. The current version passes both ways.
- [x] The "ceiling can be lowered" path was observed live — during perturbation `color-contrast` dropped below its ceiling and the run printed `set CEILINGS['pane-settings']['color-contrast'] to 9 (was 10)` without failing.

Note for reviewers reproducing locally: in a container whose preinstalled Chromium revision differs from the one this Playwright version expects, point `PLAYWRIGHT_CHROMIUM_EXECUTABLE` at it — the path `playwright.config.ts` already documents. CI installs its own browser and is unaffected.

---

https://claude.ai/code/session_01YEo3uMWio51oLyuuR4GTqr


frontend/e2e/a11y.spec.ts
```diff
@@ -3,23 +3,128 @@ import { expect, test, type Page, type TestInfo } from '@playwright/test'
...
+  const ceiling = CEILINGS[label]
+  const observed = new Map<string, number>()
+  for (const v of violations) {
+    // axe reports one entry per rule, but summing is still correct and is what
+    // keeps this from depending on that.
+    observed.set(v.id, (observed.get(v.id) ?? 0) + v.nodes.length)
+  }
+
+  const failures: string[] = []
+  for (const v of violations) {
```
tomo-chan: The counting and the reporting disagree about whether axe can emit more than one entry per rule.

The `observed` loop above sums node counts per rule id precisely so this does not depend on axe's one-entry-per-rule behaviour ("summing is still correct and is what keeps this from depending on that"). But this loop iterates the raw `violations` array again, so the moment that assumption *does* break the failure message is duplicated.

Concretely: if axe returns two `color-contrast` entries (say 2 nodes and 9), `observed` correctly holds 11, and this loop pushes the identical string `pane-settings: 'color-contrast' rose to 11 node(s), ceiling is 10.` twice into `failures`. The assertion still fails — correctly — but the message repeats one violation N times, and a reader has to work out whether that means two distinct problems.

Iterating the map instead makes the reporting match the counting and drops the redundant second lookup:

```suggestion
  for (const [id, actual] of observed) {
    const allowed = ceiling[id] ?? 0
    if (actual <= allowed) continue
    const impact = violations.find((v) => v.id === id)?.impact ?? 'unknown'
```
(the `failures.push(...)` below then uses `id`, `impact` and `actual` in place of `v.id`, `v.impact ?? 'unknown'` and `actual`.)

---

frontend/e2e/a11y.spec.ts
```diff
@@ -3,23 +3,128 @@ import { expect, test, type Page, type TestInfo } from '@playwright/test'
...
+    region: 7,
+  },
+}
+
+type CeilingCheck = {
+  failures: string[]
+  lowerable: string[]
+}
+
+function checkAgainstCeiling(
```
tomo-chan: `checkAgainstCeiling` is a pure function over `(label, violations)`, and it has no test of its own — DEVELOPMENT.md's test-granularity rule ("test at the smallest unit that exercises the logic, not at the outermost entry point") points at exactly this shape.

The consequence is not hypothetical. On a healthy repository the observed counts equal the ceilings exactly, so every CI run exercises only the `actual === allowed` path and both returned arrays come back empty. None of the four interesting branches ever runs: new rule at ceiling zero, count risen above ceiling, `set CEILINGS[...] to N (was M)`, and `remove '<id>'`. An inversion introduced into any of them later — `actual >= allowed` written as `actual > allowed` in the lowerable loop, or a dropped `?? 0` making an unlisted rule compare against `undefined` and silently never fail — ships green, because the perturbation check in the test plan was a one-time manual run rather than a standing test.

It is also the cheapest thing in this PR to test: exporting the comparator (or moving it to `e2e/a11y-ceiling.ts`) makes it a vitest table over `(ceiling, observed)` pairs, covering all four branches and both message forms without needing a broken page or a Playwright run. That also gives the ceiling map a second reader, which is worth something given the doc table in `quality-gateway.md` has to be kept in sync with it by hand.

---

tomo-chan (commented):
Reviewed the diff (6 files, `a11y.spec.ts` / `a11y.yml` / `playwright.config.ts` plus docs) for correctness.

**No correctness defect found in the gate logic.** The ceiling comparison is right in both directions: an unlisted rule falls to `?? 0` and fails on first appearance, `actual <= allowed` passes at exactly the ceiling, and the lowerable loop's `actual >= allowed` correctly stays quiet on a failing rule so a rise is never also reported as lowerable. The `passes.length + violations.length > 0` guard does close the "page rendered nothing, therefore zero violations" hole, and the two `toBeVisible()` waits before each scan close the partial-render version of it.

The server separation checks out too: port 4178 is unused elsewhere in the repo, `a11y.yml` is structurally identical to the `workspace-switch.yml` / `core-multiplexer.yml` fixtures that already boot cleanly, `run-panemux-e2e.sh` takes fixture name and port positionally and copies to a per-fixture temp path so there is no collision with the other five servers, and `A11Y_SPEC` is both added to the default project's `testIgnore` and matched by the new `chromium-a11y` project — so the spec runs exactly once, against its own page. No `scenarios.md` `auto` row or `check-scenarios` reference names the two renamed test titles, so the rename does not break the ledger check.

Two comments inline, neither blocking: one real inconsistency in how failures are reported versus counted, and one about the comparator being untestable where it currently sits.

---

frontend/e2e/a11y.spec.ts
tomo-chan: Correct, and fixed in a8f78bb — both loops now read `observed` and nothing else. Taking `impact` from `violations.find(...)` as suggested, since the map only carries counts.

Worth recording that the duplicate is not the only consequence: with two entries the old loop also read `v.impact` from *whichever entry it happened to be on*, so two `color-contrast` entries with different impacts would have produced two messages disagreeing about severity for one summed count.

The case is now pinned rather than left to reasoning — `sums a rule split across several axe entries into one count and one message` in `e2e/a11y-ceiling.test.ts`. It was confirmed to fail against the previous loop (1 failed, 12 passed) before the fix was kept.

---

frontend/e2e/a11y.spec.ts
tomo-chan: Agreed, and done in a8f78bb — moved to `e2e/a11y-ceiling.ts` (the option you named) with `e2e/a11y-ceiling.test.ts` as a vitest table. 13 cases: all four branches, both message forms, `unknown` impact, a failure and a lowerable entry in the same result, reading the ceiling for the label given rather than the other one, and the duplicate-entry case from the other thread.

The framing about a green run only exercising `actual === allowed` is now decision D11's, since it is a property of ratchets generally rather than of this comparator.

Two things the move turned up that are worth flagging rather than leaving in the diff to be discovered:

- **Splitting `e2e/` between two runners is not free.** Playwright's default `testMatch` claims `*.test.ts` as well as `*.spec.ts`, and letting it load a vitest file is not a per-file failure — `@vitest/expect` throws `Cannot redefine property: Symbol($$jest-matchers-object)` and takes the **whole** E2E suite down. Both configs now draw the line by suffix and say why. Caught by running `make test-e2e`, not by review.
- **`make efficacy` skips this branch entirely** ("no implementation changed against origin/main"), so the red-check does not cover the new test. I verified the equivalent by hand instead: reverting the failure loop to its previous form fails exactly the duplicate-entry case, 1 failed / 12 passed.

`make check` passes (33 files, 758 tests — was 32/745) and `make test-e2e` is 30 passed.

---

PR #203, opened 2026-09-03
test: cover internal/api/handler.go's unexecuted error branches (#195)

tomo-chan:
`make coverage-blocks` reported 275 blocks the suite never entered, 68 of them in `internal/api/handler.go` — a quarter of the whole backlog in one file. These are error branches the 80% statement threshold cannot see: the happy path around an `if err != nil { ... }` carries its function over the threshold whether or not any test ever enters the body.

`internal/api/handler.go` goes from **68 blocks / 99 statements to 9 / 11**. Repository-wide the count reads **211–214**, down from 275; the ±3 is the `internal/ws/board_command.go` goroutine timing #195 documents, not this branch. Coverage of the gated packages goes from **86.7% to ~89.8%**.

**Nothing in the implementation changes.** This is a test-only branch, so `make efficacy` and the diff-scoped `coverage-blocks`/`mutation` gates all report "no implementation changed" and skip.

## Making production code fail where it offers no seam

- **Config writes.** `Config.filePath` is unexported and `write()` returns nil when it is empty, so a hand-built `Config` never reaches the "failed to save" branches at all. The fixture loads a real config file and then replaces its parent directory with a regular file, so every later `MkdirAll` fails with `ENOTDIR`. Dropping the directory's write permission would not work: the suite runs as root in CI containers, and root ignores the bits.
- **ssh config reads and writes.** A directory in place of the file makes `ParseHosts` fail (a missing file is deliberately not an error there, so it cannot stand in for one). A config file inside a dangling symlink reads clean and writes dirty, which is what reaching `AppendHost`'s own failure branch requires.
- **Binary lookup.** `PATH` pointed at an empty directory for the not-found arms of `findVSCode`, `findGH` and the git-not-installed classification. Note `exec.LookPath` consults `PATH` only for a bare name, so the darwin bundle probe needs a skip rather than an emptied `PATH` — otherwise the handler test would launch a real editor on a Mac.

## Behavior this pins rather than fixes

Three routes mutate in-memory state before the step that can fail and none rolls back, so the 500 they return is only half of what happened. Confirmed with temporary probes, and **tracked in #204**:

| route | state after the 500 |
|---|---|
| `DeleteWorkspace` | workspace already removed from the config; its panes still running, reachable from no tab |
| `PostWorkspace` | the workspace it reported as *not* created is still there, and selected |
| `DeleteSession` | session already unregistered **and closed**, pane already dropped from `cfg.Layout` |

In all three, the next successful write from any other route persists the change the operator was told had failed.

They are pinned as-is here deliberately: rollback semantics are a behavior change with their own scope, while this branch retires unexecuted blocks. The three tests name #204 and are expected to go red when it is fixed.

## Also fixed along the way

Eight pre-existing tests in `handler_test.go` built a `Handler` without pinning `ghBinaryPath`, so `lookupPRInfo` resolved `gh` from the developer's own `PATH` and made a real network call for a repository that does not exist. Found by attributing calls with a logging shim named `gh` placed first on `PATH`, run per test; the same shim now reports **zero** invocations of a real `gh` or `code` across `go test ./...`.

## Left in the backlog

Nine blocks that need either a marker or infrastructure this repository does not have, so they stay counted rather than silenced: the darwin-only branch of `findVSCode` (2), the success path of `listRemoteDirectories`, which needs a real SSH server (3), a second `git rev-parse` failing where the first succeeded, `git branch --show-current` failing (the code comment says detached HEAD does that; git 2.43 exits 0 there), `resolveSSHConfigHostAlias`'s empty-`Hostname` arm, which `ParseHosts` cannot produce because it fills `Hostname` from `Name`, and `filepath.Abs` failing, which needs `os.Getwd` to fail process-wide.

No `//coverage:exempt` markers were added: they would make `handler.go` an implementation change, pulling this test-only branch into the efficacy gate and requiring an exempt marker on every new test.

## Test plan

- [x] `make check` — exit 0
- [x] `go test ./internal/api/ -race`
- [x] Two tautology fixes verified by deleting the guard each covers and watching the test go red
- [x] `go test ./...` under a logging `gh`/`code` shim on `PATH` — zero external invocations

Closes part of #195. Follow-up: #204.

Claude-Session: https://claude.ai/code/session_01FpqKsB4vF9pSsqWWQm9a8Z

### DIFF
--- a/internal/api/handler_error_paths_test.go
+++ b/internal/api/handler_error_paths_test.go
@@ -1,11 +1,9 @@
 package api
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"io/fs"
-	"log"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -129,23 +127,6 @@ func writeFakeBinary(t *testing.T, dir, name, body string) string {
 	require.NoError(t, os.WriteFile(path, []byte(body), 0600))
 	require.NoError(t, os.Chmod(path, 0755))
 	return path
-}
-
-// captureLog redirects the standard logger for the duration of the test and
-// returns the buffer it writes into.
-func captureLog(t *testing.T) *bytes.Buffer {
-	t.Helper()
-
-	var buf bytes.Buffer
-	originalWriter := log.Writer()
-	originalFlags := log.Flags()
-	log.SetOutput(&buf)
-	log.SetFlags(0)
-	t.Cleanup(func() {
-		log.SetOutput(originalWriter)
-		log.SetFlags(originalFlags)
-	})
-	return &buf
 }
 
 func doRequest(t *testing.T, h *Handler, method, path, body string) *httptest.ResponseRecorder {
@@ -281,10 +262,25 @@ func TestConfigWriteFails_MutatingRoutesReturn500(t *testing.T) {
 	}
 }
 
-// DeleteWorkspace persists before it tears the workspace's sessions down, so a
-// failed write must leave the manager untouched — otherwise the config on disk
-// and the running sessions disagree after the 500.
-func TestDeleteWorkspace_ConfigWriteFails_LeavesSessionsRunning(t *testing.T) {
+// ── What a failed mutation actually leaves behind ────────────────────────────
+//
+// The 500s asserted above are only half of what these routes do. Each mutates
+// in-memory state before the step that can fail, and none of them rolls back
+// when it does, so the operator is told the operation failed while part of it
+// has already happened — and the next successful write from any other route
+// persists that part to config.yaml.
+//
+// These three tests pin the state as it is today rather than the state it
+// arguably should be. That is deliberate and not an endorsement: this branch
+// exists to retire unexecuted blocks, and rollback semantics are a behavior
+// change with a scope of their own. Pinning puts the asymmetry in the diff a
+// reviewer reads, and makes any future fix show up as a red test here instead
+// of a silent change nobody notices.
+
+// DeleteWorkspace calls RemoveWorkspace before SaveWorkspaces, so a failed
+// write leaves the workspace gone from the in-memory config while its panes go
+// on running — unreachable from any tab, since no workspace lists them.
+func TestDeleteWorkspace_ConfigWriteFails_DropsTheWorkspaceButNotItsSessions(t *testing.T) {
 	h := newUnwritableWorkspaceHandler(t)
 	h.manager.Add(newMockSession("two-main"))
 
@@ -292,7 +288,52 @@ func TestDeleteWorkspace_ConfigWriteFails_LeavesSessionsRunning(t *testing.T) {
 
 	require.Equal(t, http.StatusInternalServerError, rec.Code)
 	_, ok := h.manager.Get("two-main")
-	assert.True(t, ok, "the workspace's session must survive a failed save")
+	assert.True(t, ok, "the teardown loop runs after the save, so the session survives")
+	assert.Len(
+		t,
+		h.cfg.WorkspacesView().Items,
+		1,
+		"today the workspace is already gone from the config the 500 says was not saved",
+	)
+}
+
+// PostWorkspace calls AddDefaultWorkspace before creating any session, so a
+// failed create leaves a workspace behind whose panes have no session at all.
+func TestPostWorkspace_CreateSessionFails_LeavesThePhantomWorkspaceBehind(t *testing.T) {
+	cfg, _ := loadWorkspaceTestConfigFromFile(t)
+	h := NewHandler(cfg, session.NewManager(), nil, nil)
+	h.sshConfigPath = filepath.Join(os.TempDir(), "panemux-test-ssh-config-nonexistent")
+	h.createSession = func(*config.PaneConfig, map[string]config.SSHConnection) (session.Session, error) {
+		return nil, errors.New("pty allocation refused")
+	}
+
+	rec := doRequest(t, h, http.MethodPost, "/api/workspaces", "")
+
+	require.Equal(t, http.StatusInternalServerError, rec.Code)
+	assert.Contains(t, rec.Body.String(), "failed to create session: pty allocation refused")
+	assert.Len(
+		t,
+		h.cfg.WorkspacesView().Items,
+		3,
+		"today the workspace the 500 reports as not created is still in the config",
+	)
+	assert.Equal(t, "workspace-3", h.cfg.WorkspacesView().Active)
+}
+
+// DeleteSession removes and closes the session before it saves, so a failed
+// write returns 500 for a pane that is already gone for good.
+func TestDeleteSession_ConfigWriteFails_ClosesTheSessionAnyway(t *testing.T) {
+	h := newUnwritableWorkspaceHandler(t)
+	sess := newMockSession("one-main")
+	h.manager.Add(sess)
+
+	rec := doRequest(t, h, http.MethodDelete, "/api/sessions/one-main", "")
+
+	require.Equal(t, http.StatusInternalServerError, rec.Code)
+	_, ok := h.manager.Get("one-main")
+	assert.False(t, ok, "today the session is unregistered before the save is attempted")
+	assert.True(t, sess.closed, "and closed, so the 500 cannot be retried into a working pane")
+	assert.Empty(t, panesInLayout(h.cfg.Layout), "and dropped from the layout the 500 says was not saved")
 }
 
 // ── Creating sessions ────────────────────────────────────────────────────────
@@ -322,20 +363,6 @@ func TestPostSession_CreateFails_Returns500(t *testing.T) {
 	assert.Contains(t, rec.Body.String(), "pty allocation refused")
 	_, ok := h.manager.Get("new-pane")
 	assert.False(t, ok)
-}
-
-func TestPostWorkspace_CreateSessionFails_Returns500(t *testing.T) {
-	cfg, _ := loadWorkspaceTestConfigFromFile(t)
-	h := NewHandler(cfg, session.NewManager(), nil, nil)
-	h.sshConfigPath = filepath.Join(os.TempDir(), "panemux-test-ssh-config-nonexistent")
-	h.createSession = func(*config.PaneConfig, map[string]config.SSHConnection) (session.Session, error) {
-		return nil, errors.New("pty allocation refused")
-	}
-
-	rec := doRequest(t, h, http.MethodPost, "/api/workspaces", "")
-
-	require.Equal(t, http.StatusInternalServerError, rec.Code)
-	assert.Contains(t, rec.Body.String(), "failed to create session: pty allocation refused")
 }
 
 // ── Reading and writing ~/.ssh/config ────────────────────────────────────────
@@ -527,15 +554,6 @@ func TestGetDirectories_UnknownConnection_Returns404(t *testing.T) {
 	assert.Equal(t, http.StatusNotFound, rec.Code)
 }
 
-func TestGetDetectShell_UnknownConnection_Returns404(t *testing.T) {
-	h := NewHandler(defaultTestConfig(), session.NewManager(), nil, nil)
-	h.sshConfigPath = filepath.Join(os.TempDir(), "panemux-test-ssh-config-nonexistent")
-
-	rec := doRequest(t, h, http.MethodGet, "/api/detect-shell?connection=nope", "")
-
-	assert.Equal(t, http.StatusNotFound, rec.Code)
-}
-
 // listLocalDirectories translates os.ReadDir's failures into messages the
 // directory browser shows the user. Each arm is a different errno and a
 // different message, and only the "does not exist" one had a test.
@@ -665,13 +683,21 @@ func TestListLocalDirectories_UnresolvablePath_Errors(t *testing.T) {
 
 // The remote lister's own wrapper. session.ListRemoteDirectories rejects a path
 // carrying shell metacharacters before it dials anything, which is what lets
-// this assert the wrapping without a network round trip — the anti-pattern
-// DEVELOPMENT.md names is a test that accepts any error a real dial produces.
+// this assert the wrapping without a network round trip.
+//
+// The prefix alone cannot say that, and asserting only the prefix would be the
+// first row of DEVELOPMENT.md's anti-pattern table: session.ListRemoteDirectories
+// wraps a failed remote command with the byte-identical string, so an error that
+// escaped from a real dial would satisfy it just as well — and on a machine with
+// an unencrypted default SSH key the call gets that far, into a DNS lookup and a
+// TCP connect to a host that is not this test's business. The validation message
+// is what distinguishes "rejected up front" from "failed out on the network".
 func TestListRemoteDirectories_WrapsLookupFailure(t *testing.T) {
 	resp, err := listRemoteDirectories(session.SSHConfig{Host: "remote.example.com"}, "/srv; rm -rf /", false)
 
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "listing remote directories:")
+	assert.Contains(t, err.Error(), "directory path", "must be rejected by validateRemotePath, before any dial")
 	assert.Equal(t, directoryBrowserResponse{}, resp)
 }
 
@@ -779,7 +805,16 @@ func TestGetGitInfo_ActiveWorkdirLookupFails_FallsBackAndLogs(t *testing.T) {
 func TestGetGitInfo_RepoWithoutOrigin_ReturnsBranchWithoutRepoURL(t *testing.T) {
 	dir := t.TempDir()
 	runGit(t, dir, "init", "-b", "main", ".")
-	runGit(t, dir, "-c", "user.email=dev@example.com", "-c", "user.name=Dev", "commit", "--allow-empty", "-m", "init")
+	runGit(
+		t, dir,
+		"-c", "user.email=dev@example.com",
+		"-c", "user.name=Dev",
+		// initTempGitRepo disables signing for the same reason: a developer
+		// whose global gitconfig sets commit.gpgsign true would otherwise have
+		// this commit invoke gpg, and fail or block on a pinentry prompt.
+		"-c", "commit.gpgsign=false",
+		"commit", "--allow-empty", "-m", "init",
+	)
 
 	h := NewHandler(defaultTestConfig(), session.NewManager(), nil, nil)
 	h.ghBinaryPath = writeFakeGHBinary(t, ghNoPRScript)
--- a/internal/api/handler_error_paths_test.go
+++ b/internal/api/handler_error_paths_test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/api/handler_error_paths_test.go
```diff
@@ -0,0 +1,980 @@
...
+	},
+	{
+		name:     "put workspace layout",
+		method:   http.MethodPut,
+		path:     "/api/workspaces/one/layout",
+		body:     oneWorkspaceLayout,
+		wantBody: "failed to save workspaces",
+	},
+	{
+		name:   "delete session",
```
tomo-chan, line 234: This row asserts only the 500 and its message, so it silently pins behavior that contradicts the invariant `TestDeleteWorkspace_ConfigWriteFails_LeavesSessionsRunning` declares a few lines below.

`DeleteSession` (`internal/api/handler.go:475`) calls `h.manager.Remove(id)` **first**, and `Manager.Remove` (`internal/session/manager.go:96`) deletes the entry and calls `session.Close()`. Only afterwards does it `RemovePaneFromLayout` + `SaveLayout`.

Scenario: config file is unwritable (disk full, read-only mount). `DELETE /api/sessions/one-main` closes the PTY, drops the pane from the in-memory layout, and returns 500 "failed to save layout". The operator is told the delete failed while the pane is already unrecoverably gone — and the next successful config write from any other route persists the deletion anyway.

Worth either asserting the current (surprising) behavior explicitly, or treating the asymmetry with DeleteWorkspace as the bug it looks like.

---

internal/api/handler_error_paths_test.go
```diff
@@ -0,0 +1,980 @@
...
+			assert.Equal(t, http.StatusInternalServerError, rec.Code)
+			assert.Contains(t, rec.Body.String(), tt.wantBody)
+		})
+	}
+}
+
+// DeleteWorkspace persists before it tears the workspace's sessions down, so a
+// failed write must leave the manager untouched — otherwise the config on disk
+// and the running sessions disagree after the 500.
+func TestDeleteWorkspace_ConfigWriteFails_LeavesSessionsRunning(t *testing.T) {
```
tomo-chan: This checks only the session-manager half of the invariant the comment states. `DeleteWorkspace` (`internal/api/handler.go:323`) calls `h.cfg.RemoveWorkspace(id)` **before** `SaveWorkspaces()`, so by the time the 500 is returned the workspace is already gone from the in-memory config.

After the 500 this test asserts, `GET /api/workspaces` omits workspace `two` while its pane sessions are still running and unreachable from any tab — exactly the "config on disk and the running sessions disagree" state the comment says the test exists to prevent. And on a transient failure (disk space returns), the very next `PUT /api/workspaces/active` succeeds and writes a config with `two` already deleted, persisting a deletion the operator was told had failed.

Adding the other half would fail today:

```go
	_, ok := h.manager.Get("two-main")
	assert.True(t, ok, "the workspace's session must survive a failed save")
	assert.Len(t, h.cfg.WorkspacesView().Items, 2, "the workspace itself must survive a failed save")
```

---

internal/api/handler_error_paths_test.go
```diff
@@ -0,0 +1,980 @@
...
+
+	rec := doRequest(t, h, http.MethodPost, "/api/sessions", `{"id":"new-pane","type":"local"}`)
+
+	require.Equal(t, http.StatusInternalServerError, rec.Code)
+	assert.Contains(t, rec.Body.String(), "pty allocation refused")
+	_, ok := h.manager.Get("new-pane")
+	assert.False(t, ok)
+}
+
+func TestPostWorkspace_CreateSessionFails_Returns500(t *testing.T) {
```
tomo-chan: This asserts only the 500 body, which leaves the more interesting half of the branch unchecked. `PostWorkspace` (`internal/api/handler.go:296`) calls `h.cfg.AddDefaultWorkspace()` first, then creates a session per pane and returns 500 on the first `createSession` error — with no rollback of the workspace it just added, and no cleanup of sessions already added to the manager for earlier panes.

Scenario: after the 500 this test asserts, `h.cfg` holds a workspace whose panes have no live sessions. The next successful save on any other route (e.g. `PUT /api/workspaces/tab-position`) writes that phantom workspace to `config.yaml`, and it comes back on restart as a workspace whose panes never connect.

An assertion that the config is unchanged would fail today:

```go
	assert.Contains(t, rec.Body.String(), "failed to create session: pty allocation refused")
	assert.Len(t, h.cfg.WorkspacesView().Items, 2, "a workspace whose panes failed must not be left behind")
```

---

internal/api/handler_error_paths_test.go
```diff
@@ -0,0 +1,980 @@
...
+	assert.Contains(t, buf.String(), "active workdirs lookup failed: transcript unreadable")
+}
+
+// A repository with no `origin` remote is an ordinary state — a local-only
+// checkout — not a lookup failure. `git config --get remote.origin.url` exits
+// non-zero for it, and the pane header must still show the branch.
+func TestGetGitInfo_RepoWithoutOrigin_ReturnsBranchWithoutRepoURL(t *testing.T) {
+	dir := t.TempDir()
+	runGit(t, dir, "init", "-b", "main", ".")
+	runGit(t, dir, "-c", "user.email=dev@example.com", "-c", "user.name=Dev", "commit", "--allow-empty", "-m", "init")
```
tomo-chan: This fixture builds its repo by hand rather than through `initTempGitRepo`, and drops the one setting that helper is careful about: `git config commit.gpgsign false` (`handler_test.go:1987`).

On a developer machine whose global gitconfig has `commit.gpgsign = true` with a real GPG key, this `commit --allow-empty` invokes gpg — failing with "gpg failed to sign the data" when no agent/passphrase is available, or blocking on a pinentry prompt. `runGit`'s `require.NoError` then fails the test for a reason unrelated to git-info, while every other git test in the package passes on that same machine because they go through the helper that disables signing.

```suggestion
	runGit(t, dir, "-c", "user.email=dev@example.com", "-c", "user.name=Dev", "-c", "commit.gpgsign=false", "commit", "--allow-empty", "-m", "init")
```

---

tomo-chan (commented):
Code review of the error-branch tests. The suite is well-built (the failure-injection helpers are the right shape, and it passes clean under `-shuffle=on`, `-count=2`, and `golangci-lint`), so the findings are all about what the new tests *don't* assert.

Three of the four are the same shape: a mutating route mutates `h.cfg` (or the session manager) **before** the save it then reports as failed, and the new test stops at the status code — so the partial mutation gets locked in as accepted behavior rather than surfaced. `TestDeleteWorkspace_ConfigWriteFails_LeavesSessionsRunning` states the right invariant in its comment but only checks one half of it; the "delete session" row in the table has the opposite behavior from that invariant and asserts nothing about it.

The fourth is a test-reliability nit on the hand-rolled git fixture.

---

internal/api/handler_error_paths_test.go
```diff
@@ -0,0 +1,974 @@
...
+
+// The remote lister's own wrapper. session.ListRemoteDirectories rejects a path
+// carrying shell metacharacters before it dials anything, which is what lets
+// this assert the wrapping without a network round trip — the anti-pattern
+// DEVELOPMENT.md names is a test that accepts any error a real dial produces.
+func TestListRemoteDirectories_WrapsLookupFailure(t *testing.T) {
+	resp, err := listRemoteDirectories(session.SSHConfig{Host: "remote.example.com"}, "/srv; rm -rf /", false)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "listing remote directories:")
```
tomo-chan, line 702: This assertion cannot distinguish the property the comment above it claims, because `session.ListRemoteDirectories` wraps a *different* failure with the byte-identical prefix: `internal/session/ssh.go:634` returns `fmt.Errorf("listing remote directories: %w", err)` for a failed remote command, and `handler.go:1496` uses the same string for its own wrap. Any error that escapes the callee is therefore accepted.

Verified by deleting the guard the comment names — removing the `validateRemotePath` block at `internal/session/ssh.go:611-615` and re-running:

```
--- PASS: TestListRemoteDirectories_WrapsLookupFailure (0.00s)
```

The error it passed on was `listing remote directories: connecting to remote host: no auth methods configured for SSH connection` — i.e. the guard was gone, the call fell through to `dialSSHClient`, and the test stayed green.

The second half matters more than the tautology. It only stopped at `buildAuthMethods` here because this container has no SSH key; that function falls back to reading `~/.ssh/id_ed25519`, `id_rsa`, `id_ecdsa` (`ssh.go:804-818`), so on a developer machine with an unencrypted default key it returns methods, and the call reaches `dialTransportWithRetry` — a real DNS lookup and TCP dial to `remote.example.com:22` under a 30s `dialRetryBudget` (`ssh.go:223`). That is the first row of DEVELOPMENT.md's anti-pattern table: a test that makes a real network call while accepting any error as expected.

Pinning the validation message instead keeps the no-network property asserted rather than assumed:

```suggestion
	assert.Contains(t, err.Error(), "listing remote directories:")
	assert.Contains(t, err.Error(), "directory path", "must be rejected by validateRemotePath, before any dial")
```

---

internal/api/handler_error_paths_test.go
```diff
@@ -0,0 +1,974 @@
...
+func TestGetDirectories_UnknownConnection_Returns404(t *testing.T) {
+	h := NewHandler(defaultTestConfig(), session.NewManager(), nil, nil)
+	h.sshConfigPath = filepath.Join(os.TempDir(), "panemux-test-ssh-config-nonexistent")
+
+	rec := doRequest(t, h, http.MethodGet, "/api/directories?connection=nope&path=/home", "")
+
+	assert.Equal(t, http.StatusNotFound, rec.Code)
+}
+
+func TestGetDetectShell_UnknownConnection_Returns404(t *testing.T) {
```
tomo-chan: This is `TestGetDetectShell_SSH_ConnectionNotFound` (`handler_test.go:1961`) again: same `NewHandler(defaultTestConfig(), session.NewManager(), nil, nil)`, same nonexistent `sshConfigPath`, same `GET /api/detect-shell?connection=…` for a name no connection has, same lone 404 assertion. Only the query value (`nope` vs `nonexistent`) and the router-construction style differ.

Measured rather than eyeballed — coverage profile for each test alone, comparing the set of blocks each enters:

```
old covered blocks: 10  new: 10
--- blocks only the NEW test covers ---
--- (end) ---
```

Zero. For a branch whose stated purpose is retiring unexecuted blocks, this row buys none and adds a second place to update when the route changes. Worth deleting, or — if the intent was to make the new file's route list read as complete — a comment saying it deliberately restates existing coverage.

---

internal/api/handler_error_paths_test.go
```diff
@@ -0,0 +1,974 @@
...
+
+	path := filepath.Join(dir, name)
+	require.NoError(t, os.WriteFile(path, []byte(body), 0600))
+	require.NoError(t, os.Chmod(path, 0755))
+	return path
+}
+
+// captureLog redirects the standard logger for the duration of the test and
+// returns the buffer it writes into.
+func captureLog(t *testing.T) *bytes.Buffer {
```
tomo-chan: Good helper to have — but `handler_test.go` is the same package and still carries three verbatim copies of exactly this body (`handler_test.go:2096`, `:2637`, `:2902`, each the same `log.Writer()`/`log.Flags()` save, `SetOutput`/`SetFlags`, `t.Cleanup` restore). So the branch takes the copy count from three to four rather than down to one.

The previous commit on this branch already did this move in the other direction — `ghNoPRScript` was lifted out of this file into `handler_test.go` so the nine tests needing it share one definition. Same argument applies here; three call sites replaced with `buf := captureLog(t)` would leave the helper with a single home.

---

tomo-chan (commented):
Re-review of `fc2ae2e`.

`fc2ae2e` is a real fix, but for a different problem than the four threads on this PR: it pins `ghBinaryPath` in eight pre-existing `handler_test.go` tests and lifts `ghNoPRScript` up beside `writeFakeGHBinary`. The only change it makes to `handler_error_paths_test.go` is deleting those six constant lines. **All four earlier threads are still open, and all four still reproduce on this head** — I ran the assertions each one proposes:

| Thread | Result on `fc2ae2e` |
|---|---|
| `delete session` row (:253) | `manager.Get("one-main")` is false after the 500 — the PTY is closed while the operator is told the delete failed |
| `TestDeleteWorkspace_…LeavesSessionsRunning` (:287) | `WorkspacesView().Items` has 1, not 2 — `RemoveWorkspace` already mutated `h.cfg` |
| `TestPostWorkspace_CreateSessionFails_Returns500` (:327) | `Items` has 3, not 2 — the phantom workspace is left behind |
| `commit.gpgsign` (:782) | still absent; `initTempGitRepo` (`handler_test.go:1994`) still sets it |

Three further findings below, none of them blockers, all verified by experiment rather than reading:

- **`TestListRemoteDirectories_WrapsLookupFailure` passes with `validateRemotePath` deleted**, and on a machine with a default SSH key that deletion turns it into a real dial. This is the one I would not merge as-is; it is the anti-pattern DEVELOPMENT.md's table names, in a test whose own comment says it avoids exactly that.
- **`TestGetDetectShell_UnknownConnection_Returns404` covers zero blocks** its `handler_test.go` twin does not — measured by diffing the two per-test coverage profiles.
- **`captureLog` is a fourth copy**, not a deduplication: three identical inline copies remain in `handler_test.go`.

Everything else about the branch holds up. The failure-injection fixtures are the interesting part and they are well chosen — a regular file in place of a directory to defeat root, a dangling symlink for read-clean/write-dirty, `validateRemotePath` rejecting before any dial. The package passes `-count=1`, `-count=2` and `-shuffle=on`, and `golangci-lint run ./internal/api/...` reports 0 issues.

---

internal/api/handler_error_paths_test.go
tomo-chan: Confirmed with a temporary probe: after the 500, `h.manager.Get("one-main")` is already false, the mock session's `closed` is true, and `panesInLayout(h.cfg.Layout)` is empty. So all three of the states you describe are real, not just the ordering.

Took the first of your two options. The table row still pins the 500, and a new `TestDeleteSession_ConfigWriteFails_ClosesTheSessionAnyway` pins the rest:

```go
	_, ok := h.manager.Get("one-main")
	assert.False(t, ok, "today the session is unregistered before the save is attempted")
	assert.True(t, sess.closed, "and closed, so the 500 cannot be retried into a working pane")
	assert.Empty(t, panesInLayout(h.cfg.Layout), "and dropped from the layout the 500 says was not saved")
```

Not the second option, and the reason is scope rather than disagreement — the asymmetry does look like a bug, and it is the same one in the two threads below. Rollback semantics are a behavior change with a `docs/scenarios.md` row and the efficacy gate attached, while this branch is test-only for #195. The section comment says that outright, so it reads as a decision rather than an endorsement, and a future fix lands as a red test here instead of a silent change.

Happy to do it in this PR instead if you would rather not split it.

---

internal/api/handler_error_paths_test.go
tomo-chan: Right on both counts, and the comment was the worse half of it — it claimed the test prevented exactly the disagreement it was half-agreeing to. Probe confirms `WorkspacesView().Items` is length 1 after the 500.

Renamed to `TestDeleteWorkspace_ConfigWriteFails_DropsTheWorkspaceButNotItsSessions`, and it now asserts both halves — pinning 1, not the 2 your snippet asks for, since that is what the route does today:

```go
	_, ok := h.manager.Get("two-main")
	assert.True(t, ok, "the teardown loop runs after the save, so the session survives")
	assert.Len(
		t,
		h.cfg.WorkspacesView().Items,
		1,
		"today the workspace is already gone from the config the 500 says was not saved",
	)
```

Why pinned rather than fixed: see the reply on the DeleteSession thread — same defect class, and changing it is a behavior change outside this branch's scope.

---

internal/api/handler_error_paths_test.go
tomo-chan: Confirmed — probe shows 3 items after the 500, and `Active` is `workspace-3`, so the phantom is not only present but selected.

Folded this test into the stronger one rather than leaving two tests on the same call: `TestPostWorkspace_CreateSessionFails_LeavesThePhantomWorkspaceBehind` keeps the body assertion and adds the state.

```go
	assert.Contains(t, rec.Body.String(), "failed to create session: pty allocation refused")
	assert.Len(
		t,
		h.cfg.WorkspacesView().Items,
		3,
		"today the workspace the 500 reports as not created is still in the config",
	)
	assert.Equal(t, "workspace-3", h.cfg.WorkspacesView().Active)
```

Same pinned-not-fixed reasoning as the other two threads.

---

internal/api/handler_error_paths_test.go
tomo-chan: Applied, with the reason written down so the next hand-built repo does not drop it again:

```go
	runGit(
		t, dir,
		"-c", "user.email=dev@example.com",
		"-c", "user.name=Dev",
		// initTempGitRepo disables signing for the same reason: a developer
		// whose global gitconfig sets commit.gpgsign true would otherwise have
		// this commit invoke gpg, and fail or block on a pinentry prompt.
		"-c", "commit.gpgsign=false",
		"commit", "--allow-empty", "-m", "init",
	)
```

---

internal/api/handler_error_paths_test.go
tomo-chan: Applied your suggestion. The second half is the part I had missed — I checked that `validateRemotePath` runs before `dialSSHClient` and stopped there, without checking what happens if it ever does not, which is exactly the tautology.

The comment now explains why the prefix alone is not enough, rather than restating the property the assertion was supposed to carry:

```go
// The prefix alone cannot say that, and asserting only the prefix would be the
// first row of DEVELOPMENT.md's anti-pattern table: session.ListRemoteDirectories
// wraps a failed remote command with the byte-identical string, so an error that
// escaped from a real dial would satisfy it just as well — and on a machine with
// an unencrypted default SSH key the call gets that far, into a DNS lookup and a
// TCP connect to a host that is not this test's business. The validation message
// is what distinguishes "rejected up front" from "failed out on the network".
```

---

internal/api/handler_error_paths_test.go
tomo-chan: Removed. Zero new blocks is the whole argument — for a branch whose only job is retiring unexecuted blocks, a row that retires none is pure duplication.

Re-measured after deleting it: `internal/api/handler.go` still reports 9 unexecuted blocks, unchanged. (The repository-wide total reads 211 or 214 across runs; that is the `internal/ws/board_command.go` goroutine timing #195 documents, not this deletion — the per-file breakdown shows the swing there.)

---

internal/api/handler_error_paths_test.go
tomo-chan: Fair — I added a fourth copy while calling it a helper. Moved it to `handler_test.go` beside `writeFakeGHBinary`, same direction as `ghNoPRScript`, and replaced all three existing copies with `buf := captureLog(t)`. One definition now, four call sites.

---

internal/api/handler_error_paths_test.go
tomo-chan: Filed as #204, with the reproduction and the observed values for all three routes. The section comment now names it, so these tests read as "expected to go red when #204 is fixed" rather than as intended behavior.

Resolving here; the fix itself lives in that issue.

---

internal/api/handler_error_paths_test.go
tomo-chan: Tracked in #204 (case 1 there). Resolving.

---

internal/api/handler_error_paths_test.go
tomo-chan: Tracked in #204 (case 2 there). Resolving.

---

PR #208, opened 2026-09-05
test: cover internal/boardmcp's unentered error branches (#195)

tomo-chan:
The fourth slice of the per-block backlog, after #203, #205 and #206.

`internal/boardmcp` goes from **15 blocks to 2**, and the repository-wide count from **171 to 155**. Coverage of the gated packages goes from **92.1% to 92.6%**.

## Why these branches matter

This package is the whole surface the command center's LLM subprocess can reach — three tools over a stdio JSON-RPC transport. Its error handling is what stands between a malformed model-generated call and the subprocess losing its connection to panemux entirely. Most of what was unentered is exactly that: the arms that turn bad input into an *answer* rather than a dropped stream.

**The transport.** A blank line between requests is skipped rather than answered with an unsolicited parse error — any writer that terminates its last line and flushes produces one, and a response with no id to correlate is worse than silence. A broken stdout ends the loop with the reason instead of spinning through the remaining requests writing into a pipe nobody reads. A read failure is distinguished from a clean end of input, and the request read before it is still answered.

**Dispatch.** `notifications/initialized` sent as a *request* (with an id) reaches `dispatch`'s own arm, which the existing notification test never did — that one returns in `handleLine`, before dispatch. The answer is a null result, not method-not-found: the method is known, it just has no result.

**Malformed tool calls.** The three shapes a model can get wrong, each answered as invalid-params rather than crashing the handler.

**Backend failures.** `board_messages` and `board_broadcast` failures surface as a tool result with `isError` rather than a JSON-RPC error, since the call itself was well-formed — the model should see something it can react to, not a protocol fault. The assertions pin the exact wrapped text, so the tool-result-vs-protocol-error decision is what is being tested, not just that something came back.

**The REST client.** An unbuildable request, a server that is not listening (the ordinary case — the subprocess outlives a shutdown by however long its query takes), and a truncated response body. Plus the trailing-slash trim, which keeps a double slash out of the request path.

## Review

Ran the repo's `diff-reviewer` before opening this. It reported no findings, having verified the property earlier rounds on this issue found missing: **each new test was checked against a perturbation of the code it covers.** Thirteen perturbations, thirteen failures, each in exactly the intended test and no other — including dropping the `board_x: %w` wrap prefixes, deleting the `notifications/initialized` dispatch arm, and removing `strings.TrimSuffix` from the constructor.

It also flagged one comment that misstated *why* a URL fails to build (invalid syntax, not a non-loopback host). Corrected, and the comment now says what the test does not check.

## Left in the backlog

Two blocks: `client.go`'s `encoding broadcast request` and `server.go`'s `encoding mcp response`, both `json.Marshal` on fixed value sets that cannot fail.

No `//coverage:exempt` marker was added, for the reason #203 records: a marker would make these implementation changes, pulling a test-only branch into the efficacy gate.

## Test plan

- [x] `make check` — exit 0
- [x] `make lint-go` — 0 issues (one `fieldalignment` finding fixed)
- [x] `go test ./internal/boardmcp/ -race`
- [x] `make coverage-blocks` — `internal/boardmcp` 15 → 2, repository-wide 171 → 155

Closes part of #195.

---

### DIFF
--- a/internal/boardmcp/boardmcp_branches_test.go
+++ b/internal/boardmcp/boardmcp_branches_test.go
@@ -1,0 +1,284 @@
+package boardmcp
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Branches `make coverage-blocks` reported as never entered. Issue #195.
+//
+// This package is the whole surface the command center's LLM subprocess can
+// reach — three tools over a stdio JSON-RPC transport — so its error handling
+// is what stands between a malformed model-generated call and the subprocess
+// losing its connection to panemux entirely. Most of what was unentered is
+// exactly that: the arms that turn bad input into an answer rather than a
+// dropped stream.
+
+// ── The stdio transport ──────────────────────────────────────────────────────
+
+// A blank line between requests is skipped rather than answered with a parse
+// error. Any writer that terminates its last line and then flushes produces
+// one, and answering it would put an unsolicited error response on a stream
+// the client correlates by id.
+func TestServeSkipsBlankLinesBetweenRequests(t *testing.T) {
+	responses := serveLines(t, &fakeBoardAPIClient{},
+		`{"jsonrpc":"2.0","id":1,"method":"initialize"}`,
+		"",
+		"   ",
+		`{"jsonrpc":"2.0","id":2,"method":"tools/list"}`,
+	)
+
+	require.Len(t, responses, 2, "the blank lines must produce no response of their own")
+	assert.EqualValues(t, 1, responses[0]["id"])
+	assert.EqualValues(t, 2, responses[1]["id"])
+}
+
+// failingWriter fails every write, standing in for the subprocess's stdout
+// pipe closing because `claude` exited while a response was in flight.
+type failingWriter struct{ err error }
+
+func (f failingWriter) Write([]byte) (int, error) { return 0, f.err }
+
+// A broken stdout ends the loop with the reason, rather than spinning through
+// the remaining requests writing into a pipe nobody is reading.
+func TestServeReportsAWriteFailure(t *testing.T) {
+	in := strings.NewReader(`{"jsonrpc":"2.0","id":1,"method":"initialize"}` + "\n")
+
+	err := NewServer(&fakeBoardAPIClient{}).Serve(
+		context.Background(), in, failingWriter{err: errors.New("broken pipe")},
+	)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "writing mcp response")
+	assert.Contains(t, err.Error(), "broken pipe")
+}
+
+// failingReader fails after handing back whatever it was primed with, so the
+// scanner reports a read error rather than a clean end of input.
+type failingReader struct {
+	err       error
+	remaining string
+}
+
+func (f *failingReader) Read(p []byte) (int, error) {
+	if f.remaining == "" {
+		return 0, f.err
+	}
+	n := copy(p, f.remaining)
+	f.remaining = f.remaining[n:]
+	return n, nil
+}
+
+// A read failure is distinguished from end-of-input: Serve returns nil when
+// the client simply closed, and an error when the stream broke.
+func TestServeReportsAReadFailure(t *testing.T) {
+	r := &failingReader{
+		remaining: `{"jsonrpc":"2.0","id":1,"method":"initialize"}` + "\n",
+		err:       errors.New("connection reset"),
+	}
+	var out strings.Builder
+
+	err := NewServer(&fakeBoardAPIClient{}).Serve(context.Background(), r, &out)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "reading mcp request")
+	assert.Contains(t, err.Error(), "connection reset")
+	assert.Contains(t, out.String(), `"id":1`, "the request read before the failure was still answered")
+}
+
+// ── Dispatch ─────────────────────────────────────────────────────────────────
+
+// `notifications/initialized` normally arrives with no id and is answered with
+// nothing at all. A client that sends it as a request anyway gets a response,
+// not a "method not found": the method exists, it just has no result.
+func TestInitializedSentAsARequestIsAnsweredWithANullResult(t *testing.T) {
+	responses := serveLines(t, &fakeBoardAPIClient{},
+		`{"jsonrpc":"2.0","id":7,"method":"notifications/initialized"}`,
+	)
+
+	require.Len(t, responses, 1)
+	assert.EqualValues(t, 7, responses[0]["id"])
+	assert.Nil(t, responses[0]["result"])
+	assert.Nil(t, responses[0]["error"], "the method is known, so this is not method-not-found")
+}
+
+// ── Malformed tool calls ─────────────────────────────────────────────────────
+
+// The three shapes a model can get wrong, each answered as a JSON-RPC error
+// rather than crashing the handler or returning a tool result that looks like
+// success.
+func TestMalformedToolCallsReturnInvalidParams(t *testing.T) {
+	const invalidParams = -32602
+
+	tests := []struct {
+		name string
+		line string
+		want string
+	}{
+		{
+			name: "tools/call params are not an object",
+			line: `{"jsonrpc":"2.0","id":1,"method":"tools/call","params":"not-an-object"}`,
+			want: "invalid params:",
+		},
+		{
+			name: "board_messages arguments are not an object",
+			line: `{"jsonrpc":"2.0","id":2,"method":"tools/call",` +
+				`"params":{"name":"board_messages","arguments":"not-an-object"}}`,
+			want: "invalid arguments:",
+		},
+		{
+			name: "board_broadcast arguments are not an object",
+			line: `{"jsonrpc":"2.0","id":3,"method":"tools/call",` +
+				`"params":{"name":"board_broadcast","arguments":42}}`,
+			want: "invalid arguments:",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			responses := serveLines(t, &fakeBoardAPIClient{}, tt.line)
+
+			require.Len(t, responses, 1)
+			rpcErr, ok := responses[0]["error"].(map[string]any)
+			require.True(t, ok, "expected a JSON-RPC error, got %v", responses[0])
+			assert.EqualValues(t, invalidParams, rpcErr["code"])
+			assert.Contains(t, rpcErr["message"], tt.want)
+			assert.Nil(t, responses[0]["result"])
+		})
+	}
+}
+
+// A failure from panemux's own REST API is reported as a tool result with
+// isError set, naming the tool. It is deliberately not a JSON-RPC error: the
+// call itself was well-formed, so the model should see the failure as
+// something it can react to rather than a protocol fault.
+func TestClientFailuresSurfaceAsToolErrorsNamingTheTool(t *testing.T) {
+	tests := []struct {
+		name   string
+		client *fakeBoardAPIClient
+		line   string
+		want   string
+	}{
+		{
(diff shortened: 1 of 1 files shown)


internal/boardmcp/boardmcp_branches_test.go
```diff
@@ -0,0 +1,284 @@
...
+// nothing at all. A client that sends it as a request anyway gets a response,
+// not a "method not found": the method exists, it just has no result.
+func TestInitializedSentAsARequestIsAnsweredWithANullResult(t *testing.T) {
+	responses := serveLines(t, &fakeBoardAPIClient{},
+		`{"jsonrpc":"2.0","id":7,"method":"notifications/initialized"}`,
+	)
+
+	require.Len(t, responses, 1)
+	assert.EqualValues(t, 7, responses[0]["id"])
+	assert.Nil(t, responses[0]["result"])
```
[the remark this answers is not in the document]
tomo-chan: Correct on both halves, and confirmed against the real wire output rather than reasoned from the struct tag — a throwaway test printing what `Serve` writes gives:

```
{"id":7,"jsonrpc":"2.0"}
```

So the response carries neither key, the old `assert.Nil(responses[0]["result"])` could not distinguish an absent key from a null one, and the comment above it described a `"result": null` that is never sent.

Fixed in `93654e3` by asserting key presence, which is what actually crosses the wire:

```go
	assert.EqualValues(t, 7, responses[0]["id"])
	assert.Equal(t, "2.0", responses[0]["jsonrpc"])
	assert.NotContains(t, responses[0], "error", "the method is known, so this is not method-not-found")
	assert.NotContains(t, responses[0], "result", "and it has no result to report")
```

Renamed to `TestInitializedSentAsARequestIsAnsweredWithoutAResultOrAnError`, and re-verified against the perturbation it exists for: deleting the `notifications/initialized` arm from `dispatch`, so the method falls through to method-not-found, makes it red.

One thing your comment surfaces that I have left alone deliberately: JSON-RPC 2.0 asks a response to carry exactly one of `result`/`error`, and this shape carries neither. That is an implementation question — `omitempty` on `jsonrpcResponse.Result` — and this branch changes no implementation, which is what keeps it out of the efficacy gate. The test comment now records the observation instead of quietly endorsing the shape.

---

internal/boardmcp/boardmcp_branches_test.go
```diff
@@ -0,0 +1,293 @@
...
+	assert.Contains(t, err.Error(), "building request")
+	assert.Nil(t, raw)
+}
+
+// panemux not listening is the ordinary failure here: the command center
+// subprocess outlives a server shutdown by however long its query takes.
+func TestHTTPBoardAPIClientUnreachableServerIsReported(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
+	baseURL := server.URL
+	server.Close() // nothing is listening on that port any more
```
tomo-chan: The "nothing is listening on that port any more" precondition isn't guaranteed. `httptest.NewServer` binds `127.0.0.1:0`, and closing a listener that never accepted a connection returns its port to the ephemeral pool immediately (no TIME_WAIT on the listening socket). `go test ./...` runs package binaries concurrently and other packages here also start httptest servers, so if one of them binds that exact port between `server.Close()` and `client.Status(...)`, the request reaches a live foreign server, gets a 200, and `require.Error` fails — a rare, non-reproducible CI red.

A dial that can never succeed makes the test deterministic:

```suggestion
	client := NewHTTPBoardAPIClient("http://127.0.0.1:0", "token")
```

(with the `httptest` server/`baseURL` lines above dropped) — connecting to port 0 always fails immediately, and the assertion being made is only about the wrapped `calling panemux board api` error.

---

tomo-chan (commented):
Reviewed the diff (test-only: `internal/boardmcp/boardmcp_branches_test.go`). Tests pass, race-clean, `gofmt -s`/`go vet` clean, and they do close the branches they claim — the package goes to 98.1% statements with only two genuinely unreachable `json.Marshal` error blocks left. One low-severity flake risk noted inline.

---

internal/boardmcp/boardmcp_branches_test.go
tomo-chan: Right, and worth stating plainly: the comment I wrote there (`// nothing is listening on that port any more`) asserted a property the code does not establish. A closed listening socket has no TIME_WAIT, so the port is back in the ephemeral pool before the next line runs, and this repository does start httptest servers in several packages that `go test ./...` runs concurrently.

Applied the suggestion in `3dc2453`, after checking both halves rather than taking them on the wire:

- Port 0 fails deterministically, and the failure lands in the arm the test exists for — 5/5 runs: `calling panemux board api: Get "http://127.0.0.1:0/api/board/status": dial tcp 127.0.0.1:0: connect: connection refused`. The kernel refuses before any connection is attempted, so there is no port for another package to have taken.
- Re-ran the perturbation the test was written against: dropping the `calling panemux board api: %w` wrap from `client.go`'s `do()` still makes exactly this test red. The fix keeps the block covered and keeps the assertion meaningful.

The comment now says why port 0 rather than a closed server, so the next reader does not "simplify" it back.

`make check` exit 0, `make lint-go` 0 issues, `go test ./internal/boardmcp/ -race` clean, and `internal/boardmcp` stays at 2 unentered blocks — the two `json.Marshal` ones.

---

internal/boardmcp/boardmcp_branches_test.go
```diff
@@ -0,0 +1,297 @@
...
+func TestServeReportsAWriteFailure(t *testing.T) {
+	in := strings.NewReader(`{"jsonrpc":"2.0","id":1,"method":"initialize"}` + "\n")
+
+	err := NewServer(&fakeBoardAPIClient{}).Serve(
+		context.Background(), in, failingWriter{err: errors.New("broken pipe")},
+	)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "writing mcp response")
+	assert.Contains(t, err.Error(), "broken pipe")
```
tomo-chan: The comment above this test says the write failure "ends the loop … rather than spinning through the remaining requests", but with a single request line there are no remaining requests, so the test cannot observe that property. Input exhaustion and an aborted loop are indistinguishable here: both leave `Serve` returning the same wrapped error.

Verified by perturbation against this branch's head — rewriting `Serve` to record the write error and keep going:

```go
	if _, err := w.Write(append(data, '\n')); err != nil {
		writeErr = fmt.Errorf("writing mcp response: %w", err)
		continue
	}
}
if writeErr != nil {
	return writeErr
}
```

`go test -run TestServeReportsAWriteFailure` still passes, despite that being exactly the regression the comment names.

A second line the loop must never reach makes it observable. `fakeBoardAPIClient.gotSince` is already the recorder for it — it stays zero only if `board_messages` was never dispatched:

```suggestion
	in := strings.NewReader(strings.Join([]string{
		`{"jsonrpc":"2.0","id":1,"method":"initialize"}`,
		`{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"board_messages","arguments":{"since":42}}}`,
	}, "\n") + "\n")
	client := &fakeBoardAPIClient{}

	err := NewServer(client).Serve(
		context.Background(), in, failingWriter{err: errors.New("broken pipe")},
	)

	require.Error(t, err)
	assert.Contains(t, err.Error(), "writing mcp response")
	assert.Contains(t, err.Error(), "broken pipe")
	assert.Zero(t, client.gotSince, "the loop must end at the write failure, not keep dispatching")
```

Checked both directions before suggesting it: unperturbed the package is green (`-race -count=2`), and under the record-and-continue perturbation above this test alone goes red with `Should be zero, but was 42`. `gofmt -s` clean, and the covered block is unchanged, so the coverage-blocks count stays at 2.

---

tomo-chan (commented):
Re-reviewed at `3dc2453`, after the two fixes. Both check out, verified by perturbation rather than by reading:

- **Port 0 (`3dc2453`)** — deterministic now. 10/10 runs fail immediately with `connect: connection refused`, and there is no port for a concurrently-running package to have taken.
- **Key-presence assertion (`93654e3`)** — genuinely stronger than the `assert.Nil` it replaced. Dropping `omitempty` from `jsonrpcResponse.Result` (so `"result":null` is emitted) makes it red, and so does deleting the `notifications/initialized` arm from `dispatch`. The old assertion caught neither.

Package green under `-race -count=3`, `gofmt -s` and `go vet` clean, all 13 checks green.

One finding left, inline: `TestServeReportsAWriteFailure` cannot observe the property its own comment claims, and I confirmed a `Serve` that keeps dispatching after a write error still passes it.

On the JSON-RPC 2.0 shape (a response carrying neither `result` nor `error`): agreed with leaving it, and the comment recording it rather than endorsing it is the right call for a test-only branch. Worth an issue so it does not live only in a test comment — the client this serves is an LLM subprocess's MCP layer, and a stricter one could reject the response.

The two remaining `json.Marshal` blocks in `client.go`/`server.go` are outside this diff; marking them would pull implementation files into the efficacy gate, so #203's reasoning holds here too.

---

internal/boardmcp/boardmcp_branches_test.go
tomo-chan: Confirmed before applying, with your exact perturbation: `Serve` rewritten to record the write error and `continue`, then return it after the loop — `go test -run TestServeReportsAWriteFailure` prints `ok`. So the test asserted only the error's text, and the "ends the loop" half of its own comment was unverified.

Applied in `993c966`, and checked both directions:

- Unperturbed: `go test ./internal/boardmcp/ -race -count=2` green.
- Under record-and-continue: `--- FAIL: TestServeReportsAWriteFailure` / `Should be zero, but was 42`, and no other test in the package fails — so the new assertion is what catches it, not a side effect.

Also extended the comment to say why the second line is there, since a reader trimming it back to one line would silently restore the hole:

```go
// The second request line is what makes that observable. With a single line,
// input exhaustion and an aborted loop are indistinguishable — both leave
// Serve returning the same wrapped error — so the assertion the comment
// describes would hold for a Serve that recorded the write error and kept
// dispatching. `gotSince` is the recorder: it stays zero only if
// board_messages was never dispatched.
```

`make check` exit 0, `make lint-go` 0 issues, `gofmt -s` clean, and `internal/boardmcp` stays at 2 unentered blocks — the covered block is unchanged, as you noted.

On the JSON-RPC 2.0 shape: agreed it should not live only in a test comment. Filing it as its own issue and will link it here.

---

tomo-chan: Filed as #210.

Capturing the real wire output for it turned up a second instance of the same root cause, on the ordinary error path rather than an unusual client's: `ID` is `omitempty` too, so a parse-error response is written as `{"error":{...,"code":-32700},"jsonrpc":"2.0"}` with no `id` key. JSON-RPC 2.0 §5 requires `"id": null` when the id cannot be determined. Both shapes are in the issue, captured from the real server rather than read off the struct tags.

---

tomo-chan: Verified `993c966` independently, not taken on the commit message:

- Unperturbed: `go test ./internal/boardmcp/... -race -count=3` green, `gofmt -s` and `go vet` clean, package still at 98.1% statements.
- Record-and-continue perturbation: `--- FAIL: TestServeReportsAWriteFailure` / `Should be zero, but was 42`, and it is the only test that fails.
- Also re-checked the assertion the test already had: dropping the `writing mcp response: %w` wrap still makes exactly this test red, so the new line adds a property without displacing the old one.

The added comment is the part that will keep this from regressing — the hole was only visible if you knew what a single input line could not distinguish.

On #210, your second finding checks out on the real wire too. Same three-line probe through `Serve`:

```
{"error":{"message":"parse error: invalid character 'n' looking for beginning of object key string","code":-32700},"jsonrpc":"2.0"}
{"id":7,"jsonrpc":"2.0"}
{"error":{"message":"method not found: nope","code":-32601},"id":8,"jsonrpc":"2.0"}
```

The parse-error line has no `id` key at all, where §5 requires `"id": null` — and it is the more consequential of the two, since a client with no id to correlate cannot match that error to the request that caused it, which is exactly the case where it most needs to. Note `method not found` gets it right, so the difference is only whether `req.ID` survived unmarshalling, not a decision anywhere in the response path.

Nothing outstanding from my side. `ci` was still running when I looked; everything else is green.

---

PR #211, opened 2026-09-05
test: cover internal/commandcenter's store error branches (#195)

tomo-chan:
The fifth slice of the per-block backlog, after #203, #205, #206 and #208.

| file | blocks |
|---|---|
| `internal/commandcenter/history_store.go` | 11 → **3** |
| `internal/commandcenter/session_store.go` | 3 → **1** |
| `internal/commandcenter/atomic_write.go` | 6 → **4** |
| repository-wide | 155 → **146** |

## Why these branches matter

The command center's two persisted files — the `--resume` session id and the captured conversation history — are written on a path nobody watches. A query finishes, the runner saves, and a failure there surfaces only in a log line. That is the shape of failure #195 keeps finding: not an exception anyone sees, but a decision made silently and wrongly.

The one that matters most is `LoadSessionFile`'s non-ENOENT arm. A *missing* session file means "no conversation yet" and starts a fresh one. An *unreadable* file reported as missing would therefore abandon the operator's running conversation rather than fail the query — same observable, opposite meaning. The test pins that the two stay distinguishable.

`atomicWriteFile`'s rename arm is the other one worth calling out: it is the step that makes the write atomic, and the one that can still fail after everything before it succeeded. The test also pins that the deferred cleanup removes the temp file on that path, so a failing save does not accumulate debris next to the real file.

## How the failures are injected

The suite runs as root in CI, so permission bits cannot make a write fail. Each fixture breaks the shape of the path instead — all seven probed directly rather than assumed:

| target | injection | observed |
|---|---|---|
| `os.MkdirAll` | parent is a regular file | `not a directory` |
| `os.OpenFile` | path is a directory | `is a directory` |
| `f.Write` | `/dev/full` | `no space left on device` |
| `os.Open` non-ENOENT | path under a regular file | `not a directory` (`IsNotExist=false`) |
| `scanner.Err()` | directory fd into `bufio.Scanner` | `is a directory` |
| `os.Rename` | onto an existing directory | `file exists` |
| `os.UserHomeDir()` | `HOME=""` | `$HOME is not defined` |
| `json.Marshal` | `json.RawMessage("not json")` | validates on marshal, so it errors |

`/dev/full` is Linux-only, so `requireDevFull` skips rather than fails where it is absent.

## Verification

Every test was checked against a perturbation of the code it covers, rather than only against the block count. 12 of 13 perturbations fail exactly the intended test and no other.

**The exception is recorded in the test rather than glossed.** `LoadHistory`'s `if line == ""` arm is a fast path: deleting it leaves the whole package green, because a blank line then reaches `json.Unmarshal`, which errors on empty input, and the malformed-line arm skips it identically. That branch is *equivalent* in the mutation sense, not merely unreached, so no test can protect it. The test still pins the behavior — a blank line never becomes a zero-value entry — and its comment says exactly that much and no more.

## Review

Ran the repo's `diff-reviewer` before opening this. It re-ran the claims above rather than taking them on trust (25 perturbations of its own, plus direct probes of every fixture as uid 0) and confirmed them, including that the one survivor is genuinely equivalent. It raised two findings, both real and both fixed:

1. **An assertion that could not fail.** `assert.Empty(t, data)` in the unencodable-entry test was checking that a failed batch writes nothing, but the fixture passed only the bad entry — so the buffer was empty when the error fired, and an implementation that flushed on the failure path would still have written zero bytes. Confirmed: inserting `f.Write(buf.Bytes())` before that return left the package green. Fixed by putting a good entry before the bad one; the same mutation now fails exactly this test.
2. **A comment that was factually wrong.** It justified keeping the blank-line test on the grounds that `AppendHistory` writes a blank line itself. It does not — its separating newline *terminates* a prior truncated line rather than adding an empty one after it, which `TestAppendHistoryAddsSeparatingNewlineAfterUnterminatedPriorWrite` already pins. Verified by capturing the real output. The comment now gives the actual reason (an editor, a truncating filesystem, or a future writer) and notes the earlier claim was wrong.

An unverified claim in the one comment whose whole purpose is to be honest about limits was worth fixing rather than leaving.

## Left in the backlog

Eight blocks, all needing a seam — which would make this an implementation change and pull a test-only branch into the efficacy gate:

- `fileEndsWithoutTrailingNewline`'s `Stat` and `ReadAt` arms, plus its caller's — the three are coupled, and reaching any needs an already-open file whose `Stat` reports a size its `ReadAt` cannot serve.
- `json.Marshal` on `SessionState`, a struct of one `string`.
- `atomic_write.go`'s four arms after a successful `os.CreateTemp` (`Write`, `Close`, `Chmod`, and `CreateTemp` itself).

No `//coverage:exempt` marker was added, for the reason #203 records.

## Test plan

- [x] `make check` — exit 0
- [x] `go test ./internal/commandcenter/ -race -count=3`
- [x] `make coverage-blocks` — 155 → 146 (±3 per-run variance applies; the reviewer measured 158 → 143 on its own run)
- [x] 13 perturbations run; 12 kill exactly the intended test, the 13th documented in place

Closes part of #195.

---

### DIFF
--- a/internal/commandcenter/store_branches_test.go
+++ b/internal/commandcenter/store_branches_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -102,10 +103,21 @@ func TestAppendHistoryReportsAFailedWrite(t *testing.T) {
 	assert.Contains(t, err.Error(), "no space left on device")
 }
 
+// The mode check is not redundant with the stat. Existence alone does not
+// make /dev/full the ENOSPC-on-write character device: bind-mount a regular
+// file over it and the write *succeeds*, so this test fails with "An error is
+// expected but got nil" rather than skipping — and running as uid 0 is what
+// keeps that silent, since root bypasses the mode bits that would otherwise
+// turn it into an open failure. A skip is a true statement about the
+// environment; a red require.Error is a false statement about AppendHistory.
 func requireDevFull(t *testing.T) {
 	t.Helper()
-	if _, err := os.Stat("/dev/full"); err != nil {
+	info, err := os.Stat("/dev/full")
+	if err != nil {
 		t.Skipf("/dev/full is not available on this platform: %v", err)
+	}
+	if info.Mode()&os.ModeCharDevice == 0 {
+		t.Skipf("/dev/full is not a character device (%v)", info.Mode())
 	}
 }
 
@@ -161,15 +173,59 @@ func TestLoadHistorySkipsBlankLines(t *testing.T) {
 // A read that fails partway is reported, unlike a line that fails to parse —
 // LoadHistory tolerates the latter deliberately (see its doc comment) and
 // must not extend that leniency to losing the rest of the file silently.
-func TestLoadHistoryReportsAReadFailure(t *testing.T) {
-	path := directoryAt(t, filepath.Join(t.TempDir(), "history-is-a-directory"))
+//
+// The fixture is a good entry followed by a line past the 16 MB max token
+// size LoadHistory sets on its scanner, which is what makes the nil check
+// below mean anything: the good entry is parsed and appended before the read
+// fails, so returning the accumulated entries alongside the error — handing
+// callers a silently truncated history — is distinguishable from returning
+// nil. An `os.Open`'d directory reaches the same arm far more cheaply, but
+// its Scan fails on the first call, leaving the slice nil no matter what
+// LoadHistory does with it. That was the earlier fixture here, and the nil
+// check read as pinning a property it could not observe.
+//
+// The cost is ~150ms and a 16 MB file under t.TempDir(). It buys a real
+// assertion, and the failure itself is not hypothetical: a history line is a
+// raw stream-json line from `claude`, which is why that buffer limit is set
+// at all.
+func TestLoadHistoryReportsAReadFailureWithoutReturningPartialEntries(t *testing.T) {
+	path := filepath.Join(t.TempDir(), historyFileName)
+	writeHistoryWithAnOverlongLine(t, path)
 
 	entries, err := LoadHistory(path)
 
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "reading command center history file")
-	assert.Nil(t, entries)
+	assert.Nil(t, entries, "a failed read must not hand back the entries it managed to parse first")
 }
+
+// writeHistoryWithAnOverlongLine writes one well-formed entry followed by a
+// line longer than LoadHistory's scanner will accept, so the read fails after
+// the first entry has already been parsed.
+func writeHistoryWithAnOverlongLine(t *testing.T, path string) {
+	t.Helper()
+	f, err := os.Create(path)
+	require.NoError(t, err)
+	defer f.Close() //nolint:errcheck
+
+	_, err = f.WriteString(`{"at":"1970-01-01T00:00:00Z","raw":{"n":1}}` + "\n")
+	require.NoError(t, err)
+
+	const oneMiB = 1 << 20
+	chunk := strings.Repeat("x", oneMiB)
+	for written := 0; written <= maxHistoryLineBytes; written += oneMiB {
+		_, err = f.WriteString(chunk)
+		require.NoError(t, err)
+	}
+	_, err = f.WriteString("\n")
+	require.NoError(t, err)
+}
+
+// maxHistoryLineBytes mirrors the max token size LoadHistory passes to
+// scanner.Buffer. It is duplicated rather than exported: the test needs to
+// exceed that limit, and a shared constant would make the production value
+// changeable without this fixture noticing that it no longer does.
+const maxHistoryLineBytes = 16 * 1024 * 1024
 
 // ── LoadSessionFile ──────────────────────────────────────────────────────────
 
--- a/internal/commandcenter/store_branches_test.go
+++ b/internal/commandcenter/store_branches_test.go
@@ -1,0 +1,238 @@
+package commandcenter
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// The command center's two persisted files — the --resume session id and the
+// captured conversation history — are written on a path nobody watches: a
+// query finishes, the runner saves, and a failure there surfaces only in a
+// log line. These tests cover the arms that decide what that failure says,
+// because a wrong or swallowed one is invisible until continuity is already
+// lost.
+//
+// The suite runs as root in CI, so permission bits cannot be used to make a
+// write fail. Each fixture below breaks the shape of the path instead.
+
+// regularFileAt puts a regular file where a directory is expected, so any
+// attempt to descend through it fails with ENOTDIR.
+func regularFileAt(t *testing.T, path string) string {
+	t.Helper()
+	require.NoError(t, os.WriteFile(path, []byte("a regular file, not a directory\n"), 0600))
+	return path
+}
+
+// directoryAt puts a directory where a file is expected, so opening it for
+// writing fails with EISDIR and renaming onto it fails with EEXIST.
+func directoryAt(t *testing.T, path string) string {
+	t.Helper()
+	require.NoError(t, os.Mkdir(path, 0750))
+	return path
+}
+
+func historyEntry(raw string) HistoryEntry {
+	return HistoryEntry{At: time.Unix(0, 0).UTC(), Raw: json.RawMessage(raw)}
+}
+
+// ── AppendHistory ────────────────────────────────────────────────────────────
+
+func TestAppendHistoryReportsAnUncreatableParentDirectory(t *testing.T) {
+	notADir := regularFileAt(t, filepath.Join(t.TempDir(), "notadir"))
+	path := filepath.Join(notADir, "nested", historyFileName)
+
+	err := AppendHistory(path, []HistoryEntry{historyEntry(`{"a":1}`)})
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "creating command center history directory",
+		"the caller must be able to tell a directory failure from a write failure")
+}
+
+func TestAppendHistoryReportsAnUnopenableFile(t *testing.T) {
+	path := directoryAt(t, filepath.Join(t.TempDir(), "history-is-a-directory"))
+
+	err := AppendHistory(path, []HistoryEntry{historyEntry(`{"a":1}`)})
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "opening command center history file")
+}
+
+// json.RawMessage validates on marshal rather than passing bytes through, so
+// an entry holding something that is not JSON fails here rather than being
+// written and only failing on the next LoadHistory.
+//
+// The good entry before the bad one is what makes the batching contract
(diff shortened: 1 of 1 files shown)


internal/commandcenter/store_branches_test.go
```diff
@@ -0,0 +1,238 @@
...
+// LoadHistory tolerates the latter deliberately (see its doc comment) and
+// must not extend that leniency to losing the rest of the file silently.
+func TestLoadHistoryReportsAReadFailure(t *testing.T) {
+	path := directoryAt(t, filepath.Join(t.TempDir(), "history-is-a-directory"))
+
+	entries, err := LoadHistory(path)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "reading command center history file")
+	assert.Nil(t, entries)
```
tomo-chan: This `assert.Nil(t, entries)` cannot fail — the same class of finding the PR description records fixing in `TestAppendHistoryReportsAnUnencodableEntry`.

The directory fixture makes `scanner.Scan()` return false on its very first call, so `entries` is nil at the `scanner.Err()` check regardless of what `LoadHistory` returns there. Verified by perturbation: changing

```go
return nil, fmt.Errorf("reading command center history file: %w", err)
```

to `return entries, fmt.Errorf(...)` — which would hand callers a silently truncated history alongside the error, exactly the "a real failure must not look like a partial read" property this line reads as pinning — leaves the whole package green.

Worth noting the contrast: the identical-looking assertion at line 123 in `TestLoadHistoryReportsAnOpenFailureThatIsNotAMissingFile` *does* have teeth (returning `[]HistoryEntry{}` from that arm fails it), so the two read alike but only one is load-bearing.

The `require.Error` above still protects the branch itself, so this is not a coverage gap — just an assertion that reads stronger than it is. To give it teeth the fixture would need a file whose first line(s) parse and whose read then fails partway, e.g. a valid entry followed by a line longer than the 16MB `scanner.Buffer` max token size, so a partial-entries return is distinguishable from nil.

---

tomo-chan (commented):
Reviewed the diff (one new test file, no implementation change). I re-ran the PR's own claims rather than taking them on trust: all 22 tests in the package pass as uid 0, `gofmt -s`/`go vet` are clean, and I ran a perturbation battery against the branches this file targets — `LoadSessionFile`'s non-ENOENT arm, `atomicWriteFile`'s rename arm and its deferred temp cleanup, `atomicWriteFile`'s MkdirAll arm, `AppendHistory`'s marshal arm, and `DefaultHistoryFilePath`'s home-dir arm each kill exactly the intended test and no other. The `--` ordering of fixtures is sound and the batching fix described in the PR body (good entry before the bad one) does catch a flush-on-failure mutation, confirmed.

One finding, low severity — an assertion that reads stronger than it is. Inline.

---

internal/commandcenter/store_branches_test.go
```diff
@@ -0,0 +1,238 @@
...
+	assert.Contains(t, err.Error(), "writing command center history file")
+	assert.Contains(t, err.Error(), "no space left on device")
+}
+
+func requireDevFull(t *testing.T) {
+	t.Helper()
+	if _, err := os.Stat("/dev/full"); err != nil {
+		t.Skipf("/dev/full is not available on this platform: %v", err)
+	}
+}
```
tomo-chan, line 122: `requireDevFull` guards on existence alone, so in an environment where `/dev/full` exists but is not the ENOSPC-on-write character device, this test fails instead of skipping.

Verified rather than argued — in a mount namespace with an empty regular file bind-mounted over `/dev/full`:

```
--- FAIL: TestAppendHistoryReportsAFailedWrite (0.00s)
    store_branches_test.go:100: An error is expected but got nil.
```

Note it is not the `assert.Contains` pair failing on the wrong error: the write simply *succeeds*, so `require.Error` fails first. The suite running as uid 0 (this file's own header comment says so) is what makes that case silent — root bypasses the mode bits that would otherwise turn it into an open failure instead.

`os.Stat` can't distinguish the two, but the mode it already returns can:

```go
func requireDevFull(t *testing.T) {
	t.Helper()
	info, err := os.Stat("/dev/full")
	if err != nil {
		t.Skipf("/dev/full is not available on this platform: %v", err)
	}
	if info.Mode()&os.ModeCharDevice == 0 {
		t.Skipf("/dev/full is not a character device (%v)", info.Mode())
	}
}
```

CI is `ubuntu-latest` only, so nothing is failing today — this is about what the fixture *reports* when it is wrong. A skip is a true statement about the environment; a red `require.Error` is a false statement about `AppendHistory`, and that is the more expensive one to debug.

---

tomo-chan (commented):
Second pass on this branch. Two further findings, neither blocking; the earlier inline finding at line 171 is still open.

One is inline below, on the `/dev/full` fixture's guard.

The other has no line in this diff to sit on, because it is about lines this branch does not touch. After it lands, eight error blocks in the three store files are still unentered — taken from the package's own coverage profile, not read off the diff:

```
atomic_write.go   27  creating temp %s
atomic_write.go   33  writing temp %s
atomic_write.go   37  closing temp %s
atomic_write.go   40  setting %s mode
history_store.go  51  checking command center history file
history_store.go  81  fileEndsWithoutTrailingNewline: stat
history_store.go  88  fileEndsWithoutTrailingNewline: reading last byte
session_store.go  44  encoding command center session file
```

Two of them deserve a recorded decision rather than silence:

- **`session_store.go:44` is provably unreachable** — `json.Marshal` of `SessionState`, a struct with one `string` field, cannot fail. DEVELOPMENT.md's per-block rule asks for `//coverage:exempt <reason>` on exactly this shape, so the next reader inherits the argument instead of re-deriving it.
- **`history_store.go:51` is the one carrying real behavior.** Nothing asserts that `AppendHistory` *reports* a `fileEndsWithoutTrailingNewline` failure rather than falling through with `needsLeadingNewline == false` and concatenating onto a truncated tail — which is the corruption `TestAppendHistoryAddsSeparatingNewlineAfterUnterminatedPriorWrite` exists to prevent. I could not construct a fixture that reaches it: `/dev/full`, `/dev/null`, `/dev/zero`, `/dev/loop0`, `/dev/sda` and `/proc/self/environ` all report `Size() == 0` and take the early return, so `f.Stat()`/`f.ReadAt` never get a chance to fail. Its reachability is genuinely open — which is itself the argument for writing an answer into the file rather than leaving the next person to redo this probe.

Why this is about the ledger and not about CI: `make coverage-blocks` is diff-scoped, and this branch changes no implementation line, so none of the eight fail anything. #195's entry for `internal/commandcenter` will read as closed while they remain.

Two scope notes, so they are not mistaken for gaps: `docs/scenarios.md` is not required here (the CI rule covers `frontend/src`, `internal/api`, `internal/config`), and `make efficacy` skips a branch that changes tests but no implementation.

---

internal/commandcenter/store_branches_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+// ── Default paths ────────────────────────────────────────────────────────────
+
+// Both default paths are resolved against the home directory, and both are
+// called at startup. An empty HOME is what os.UserHomeDir reports on, and it
+// is reachable in a stripped environment such as a systemd unit or a
+// container with no passwd entry.
+func TestDefaultPathsReportAnUnresolvableHomeDirectory(t *testing.T) {
+	t.Setenv("HOME", "")
```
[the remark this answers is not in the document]
tomo-chan: Confirming the rule, with a correction to the scope — my own second-pass review reached the same finding, so rather than post a duplicate thread I'll put the verified version here.

The rule is real: DEVELOPMENT.md's testability rule names `os.UserHomeDir()` explicitly, and `internal/session/local.go:27`'s `var userHomeDirFn = os.UserHomeDir` is the seam pattern it points at.

But it is not this branch's to introduce, on two counts:

- `os.UserHomeDir()` is called directly at **14 production sites** in this repo and exactly one of them (`local.go`) has a seam. `t.Setenv("HOME", "")` is how the identical branch is already covered in `internal/config/config_branches_test.go:76` and `:87`, and in `internal/api/handler_error_paths_test.go:776`. This file follows the prevailing precedent, it does not depart from it.
- The parallelization argument does not land on this PR either: `internal/commandcenter/context_test.go:97` already calls `t.Setenv("HOME", ...)` and predates this branch, so the package was non-parallel before this file existed.

So: worth doing, worth doing repo-wide as its own change — `DefaultHistoryFilePath`, `DefaultSessionFilePath` and `DefaultContextDir` would share one seam in this package alone — and not worth blocking a test-only branch over.

---

internal/commandcenter/store_branches_test.go
tomo-chan: Correct, and reproduced before changing anything: `return entries, fmt.Errorf(...)` leaves the whole package green, while the same shape at the open-failure site (`return []HistoryEntry{}, ...`) fails its test immediately. The two really do read alike with only one load-bearing.

Fixed in `12c07c5` by taking the fixture you described — a good entry, then a line past the 16 MB max token size the scanner is given. The entry is parsed and appended before the read fails, so a partial return is distinguishable from nil. All three perturbations now kill it, and no other test:

| perturbation | result |
|---|---|
| `return entries, fmt.Errorf(...)` | `Expected nil, but got: []HistoryEntry{...{"n":1}}` |
| `return nil, err` (drop the wrap) | fails on the wrap text |
| raise the scanner limit to 64 MB | fails — so the fixture, not something incidental, is what triggers the read failure |

That last one is the check on the fixture itself: if the overlong line stopped being overlong the test would have to notice, and it does.

Cost is ~150ms and a 16 MB file under `t.TempDir()`, which I think it earns — a history line is a raw stream-json line from `claude`, so this failure is the reason that buffer limit is set at all, and the directory fixture covered no block this one does not. Renamed to `TestLoadHistoryReportsAReadFailureWithoutReturningPartialEntries`, and the comment records why it is not the cheaper directory fixture so nobody trades it back.

---

internal/commandcenter/store_branches_test.go
tomo-chan: Applied in `12c07c5`, and confirmed the fixed guard in the same mount namespace you used — a regular file bind-mounted over `/dev/full`:

```
=== RUN   TestAppendHistoryReportsAFailedWrite
    store_branches_test.go:97: /dev/full is not a character device (-rw-r--r--)
--- SKIP: TestAppendHistoryReportsAFailedWrite (0.00s)
PASS
```

The framing is the part worth keeping, so I put it in the comment above the helper rather than leaving the mode check looking like a redundant stat: a skip is a true statement about the environment, a red `require.Error` is a false statement about `AppendHistory`, and uid 0 is what makes the wrong one silent.

---

internal/commandcenter/store_branches_test.go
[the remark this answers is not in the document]
tomo-chan: Agreed on the rule and on the scoping. Filed as #212 so it does not get lost.

Counted the sites rather than estimating them, and they back the "repo-wide, not here" call: `os.UserHomeDir()` is called directly at **16 production sites across 9 files**, with `internal/session/local.go:27`'s `var userHomeDirFn = os.UserHomeDir` the only seam. **12 test files** already reach for `t.Setenv("HOME", ...)`, this one included — `internal/commandcenter/context_test.go:97` among them, which predates this branch, so the package was already non-parallel before this file existed.

Three of those sites also use `home, _ := os.UserHomeDir()` and discard the error, which is worth deciding deliberately during that pass rather than carrying forward. The issue records that too.

Not changing it here: a seam is an implementation change, and this branch is test-only precisely so `make efficacy` and the diff-scoped gates skip. Introducing one for two of the three defaults in this package — `DefaultContextDir` is the third — would also leave the repository with two different conventions instead of one.

---

PR #213, opened 2026-09-06
test: cover internal/commandcenter/runner.go's failure arms (#195)

tomo-chan:
The sixth slice of the per-block backlog, after #203, #205, #206, #208 and #211.

`internal/commandcenter/runner.go` goes from **9 blocks to 1**, and the repository-wide count from **146 to 138**.

**Nothing in the implementation changes.** One test file is added and one assertion is added to an existing test, so `make efficacy` and the diff-scoped `coverage-blocks`/`mutation` gates all report "no implementation changed" and skip.

## Why these branches matter

Every failure in a command-center turn reaches the operator as an `EventError` on the WS stream and nowhere else — there is no HTTP status to carry it and no caller to return it to. These are the arms that decide what that event says, and whether the turn stops there.

That second half is what most of them actually pin:

- **A setup failure must abort before the subprocess is launched.** Session-state load, session-id mint, work-directory creation, stdout pipe: each asserts no `claude` process was started, not just that an error was reported.
- **A failed history write must not fail the turn.** History is a record of what already happened; the operator has the answer on screen either way. The turn still reports done.
- **A failed session-id write must abort.** This is the opposite case and the one worth having a test for: the session id is what the *next* turn resumes from, so reporting done would leave the next query silently starting a new conversation with no sign anything was lost.

The work-directory test also pins that the MCP config's cleanup still runs when a later step fails — otherwise the token-bearing temp file `BuildMCPConfig` writes ([docs/security.md](docs/security.md)) would outlive the query that needed it.

## Two fixtures worth noting

**The session-id arm needs the load to succeed and only the save to fail**, so its path sits under a dangling symlink. Reading through one is `ErrNotExist`, which `LoadSessionFile` treats as "no session yet", while `MkdirAll` on one fails with `EEXIST`. A regular-file parent — the fixture the other tests use — fails the load instead and lands on a different arm.

**The stdout-pipe arm needs a `cmdRunner` whose `StdoutPipe` can fail**, which the existing `fakeCmd` cannot do: it always returns a nil error, so that arm is unreachable through it.

## Verification

Every test was checked against a perturbation of the code it covers; each kills exactly the intended test.

One process note, since this PR's own standard is "perturb and confirm red": my first pass reported the blank-line perturbation as a survivor. It was not — deleting that arm made `bytes` an unused import, and my check was matching on `^--- FAIL` only, so a build failure read as "no test failed". Re-run properly, it kills the test. Worth recording because the verification method can produce a false negative that looks exactly like a real one.

## One test written and then dropped

A `BuildMCPConfig` failure test was written for this and removed before opening the PR: `TestRunnerBuildMCPConfigFailureEmitsErrorAndNeverStartsProcess` already covers that block, so a second one would have retired nothing — the anti-pattern [#209](https://github.com/tomo-chan/panemux/issues/209) records.

What it added over the existing test moved into that test instead: an assertion on the `building mcp config` wrap prefix. The existing test asserted only that the underlying message (`disk full`) appeared, so dropping the prefix did not fail it — confirmed by perturbation, and it fails now. Without the prefix an operator cannot tell a config failure from a failure to start `claude`, which are different problems on their host.

## Left in the backlog

One block: `promptHistoryEntry`'s `json.Marshal` fallback, which the code's own comment already records as unreachable (`json.Marshal` on a `map[string]string` cannot fail; the fallback exists to keep the entry well-formed if that ever stops being true).

No `//coverage:exempt` marker was added, for the reason #203 records: a marker would make this an implementation change and pull a test-only branch into the efficacy gate.

## Test plan

- [x] `make check` — exit 0
- [x] `make lint-go` — 0 issues (one `fieldalignment` finding fixed)
- [x] `go test ./internal/commandcenter/ -race -count=3`
- [x] `make coverage-blocks` — `runner.go` 9 → 1, repository-wide 146 → 138
- [x] 9 perturbations run; each kills exactly the intended test

Closes part of #195.

---

### DIFF
--- a/internal/commandcenter/runner_branches_test.go
+++ b/internal/commandcenter/runner_branches_test.go
@@ -1,0 +1,230 @@
+package commandcenter
+
+import (
+	"context"
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Every failure in a command-center turn reaches the operator as an
+// EventError on the WS stream and nowhere else — there is no HTTP status to
+// carry it and no caller to return it to. These tests cover the arms that
+// decide what that event says, and, just as importantly, whether the turn
+// stops there or carries on.
+//
+// The distinction is the reason most of these are worth pinning: a setup
+// failure must abort before the subprocess is launched, a failed history
+// write must not fail an otherwise-good query, and a failed session-id write
+// must abort rather than report success, because reporting done would leave
+// the next turn silently starting a new conversation.
+
+// runnerFixture is the seam-by-seam Runner these tests need. newTestRunner
+// covers the happy path; this one exists to fail exactly one seam at a time.
+type runnerFixture struct {
+	started *bool
+	config  RunnerConfig
+}
+
+func newRunnerFixture(t *testing.T, stdout string) *runnerFixture {
+	t.Helper()
+	started := false
+	return &runnerFixture{
+		started: &started,
+		config: RunnerConfig{
+			ClaudeBin:   "claude",
+			SessionPath: filepath.Join(t.TempDir(), "session.json"),
+			HistoryPath: filepath.Join(t.TempDir(), "history.jsonl"),
+			BuildMCPConfig: func() (string, func(), error) {
+				return "/tmp/sample-mcp-config.json", func() {}, nil
+			},
+			NewCommand: func(_ context.Context, _, _ string, _ ...string) cmdRunner {
+				started = true
+				return &fakeCmd{stdout: io.NopCloser(strings.NewReader(stdout))}
+			},
+			Now: func() time.Time { return time.Unix(0, 0).UTC() },
+		},
+	}
+}
+
+// query runs one turn to completion and returns every event it emitted.
+func (f *runnerFixture) query(t *testing.T) []Event {
+	t.Helper()
+	events, err := NewRunner(f.config).Query(context.Background(), "how is the board")
+	require.NoError(t, err)
+	return drain(t, events)
+}
+
+// errorEvents returns just the EventError messages, so a test can assert
+// both what was reported and that nothing else was.
+func errorEvents(events []Event) []string {
+	var msgs []string
+	for _, ev := range events {
+		if ev.Type == EventError {
+			msgs = append(msgs, ev.Err)
+		}
+	}
+	return msgs
+}
+
+func hasDone(events []Event) bool {
+	for _, ev := range events {
+		if ev.Type == EventDone {
+			return true
+		}
+	}
+	return false
+}
+
+// brokenParentPath returns a path whose parent is a regular file, so both
+// reading it and creating it fail with ENOTDIR.
+func brokenParentPath(t *testing.T, name string) string {
+	t.Helper()
+	notADir := filepath.Join(t.TempDir(), "notadir")
+	require.NoError(t, os.WriteFile(notADir, []byte("not a directory\n"), 0600))
+	return filepath.Join(notADir, name)
+}
+
+// ── Setup failures, which must abort before the subprocess is launched ────────
+
+func TestQueryReportsAnUnreadableSessionFile(t *testing.T) {
+	f := newRunnerFixture(t, "")
+	f.config.SessionPath = brokenParentPath(t, "session.json")
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "loading command center session state")
+	assert.False(t, hasDone(events), "a turn that never ran must not report done")
+	assert.False(t, *f.started, "the subprocess must not be launched after a setup failure")
+}
+
+func TestQueryReportsAFailureToMintASessionID(t *testing.T) {
+	f := newRunnerFixture(t, "")
+	f.config.NewSessionID = func() (string, error) { return "", errors.New("no entropy available") }
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "no entropy available")
+	assert.False(t, hasDone(events))
+	assert.False(t, *f.started)
+}
+
+// The work directory is created after the MCP config, so this also pins that
+// the config's cleanup still runs when a later step fails — otherwise the
+// token-bearing temp file that BuildMCPConfig writes (docs/security.md) would
+// outlive the query that needed it.
+func TestQueryReportsAFailureToCreateTheWorkDirAndStillCleansUpTheMCPConfig(t *testing.T) {
+	f := newRunnerFixture(t, "")
+	cleaned := false
+	f.config.BuildMCPConfig = func() (string, func(), error) {
+		return "/tmp/sample-mcp-config.json", func() { cleaned = true }, nil
+	}
+	f.config.NewWorkDir = func() (string, func(), error) {
+		return "", nil, errors.New("creating command center work directory: no space left on device")
+	}
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "creating command center work directory")
+	assert.False(t, hasDone(events))
+	assert.False(t, *f.started)
+	assert.True(t, cleaned, "the mcp config's cleanup must run even when a later step fails")
+}
+
+// The pipe is opened after the command is constructed but before it is
+// started, so this is the last failure that still leaves nothing running.
+// It gets its own message because "creating stdout pipe" and "starting
+// claude" are different operator problems: a pipe failure is a resource
+// limit on this host, a start failure is usually a missing binary.
+func TestQueryReportsAFailureToOpenTheSubprocessStdoutPipe(t *testing.T) {
+	f := newRunnerFixture(t, "")
+	f.config.NewCommand = func(_ context.Context, _, _ string, _ ...string) cmdRunner {
+		return &noStdoutCmd{err: errors.New("too many open files")}
+	}
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "creating stdout pipe")
+	assert.Contains(t, errorEvents(events)[0], "too many open files")
+	assert.False(t, hasDone(events))
+}
+
+// noStdoutCmd fails at StdoutPipe. fakeCmd cannot: its StdoutPipe never
+// returns an error, so the arm above is unreachable through it.
+type noStdoutCmd struct{ err error }
+
+func (c *noStdoutCmd) StdoutPipe() (io.ReadCloser, error) { return nil, c.err }
+func (c *noStdoutCmd) Start() error                       { return nil }
+func (c *noStdoutCmd) Wait() error                        { return nil }
+
+// ── Persistence failures after the turn itself succeeded ─────────────────────
+
+// A history write is a record of what already happened, so losing it is
+// worth reporting but must not turn a completed turn into a failed one — the
+// operator has the answer on screen either way.
+func TestQueryReportsAFailedHistoryWriteWithoutFailingTheTurn(t *testing.T) {
+	f := newRunnerFixture(t, `{"type":"result","session_id":"s","result":"done"}`+"\n")
+	f.config.HistoryPath = brokenParentPath(t, "history.jsonl")
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "persisting command center history")
+	assert.True(t, hasDone(events), "the turn itself succeeded, so it must still report done")
+}
+
+// The session id is the opposite case: it is what the *next* turn resumes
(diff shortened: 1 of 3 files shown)


internal/commandcenter/runner_branches_test.go
```diff
@@ -0,0 +1,230 @@
...
+		return &noStdoutCmd{err: errors.New("too many open files")}
+	}
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "creating stdout pipe")
+	assert.Contains(t, errorEvents(events)[0], "too many open files")
+	assert.False(t, hasDone(events))
+}
```
tomo-chan, line 188: The comment above says "this is the last failure that still leaves nothing running", but this is the one setup test that never asserts it — the other three all check `assert.False(t, *f.started)`. `noStdoutCmd.Start()` returns `nil` and records nothing, and because this test replaces `NewCommand` wholesale, `f.started` is dead here.

Verified by perturbation: reorder `run()` so `cmd.Start()` runs *before* `cmd.StdoutPipe()` (leaving the pipe-error arm intact) and this test still passes:

```
### P1: Start() moved before StdoutPipe()
ok  	panemux/internal/commandcenter	0.006s
```

That reorder is a real defect, not a hypothetical: with a real `*exec.Cmd` the subprocess would be started and `run()` would then return on the pipe error **without ever calling `Wait()`** — a zombie `claude` per failed query — while the deferred `cleanupWorkDir()` deletes the work directory out from under the live process. This is exactly the property the three tests above exist to pin, and it is the only one of the four where it isn't pinned.

Suggest recording `Start` and asserting it, e.g. give `noStdoutCmd` a `started *bool` set in `Start()` and `assert.False(t, *started, "the subprocess must not be started when the stdout pipe cannot be opened")`. (`f.started` can also be reused by having the replacement factory set `*f.started = true`.)

---

internal/commandcenter/runner_branches_test.go
```diff
@@ -0,0 +1,230 @@
...
+		return "/tmp/sample-mcp-config.json", func() { cleaned = true }, nil
+	}
+	f.config.NewWorkDir = func() (string, func(), error) {
+		return "", nil, errors.New("creating command center work directory: no space left on device")
+	}
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "creating command center work directory")
```
tomo-chan: This assertion is a tautology with respect to the code under test. `run()` emits this arm as `errorEvent("%v", err)` — it adds **no** prefix of its own (`runner.go:241`), so the `"creating command center work directory"` text asserted on line 138 is the fixture's own literal from line 132 round-tripping through the channel. The only production code that actually produces that context is `NewWorkDir`'s wrap in `context.go:155`, and this test cannot see it.

Verified: delete the wrap in the real `NewWorkDir` —

```go
return "", func() {}, err   // was: fmt.Errorf("creating command center work directory: %w", err)
```

— and the whole package stays green (`ok panemux/internal/commandcenter 0.235s`), including this test. An operator whose `/tmp` is full would then get a bare `no space left on device` with nothing saying which step failed, and no test would notice.

This is the same gap the assertion you added to `TestRunnerBuildMCPConfigFailureEmitsErrorAndNeverStartsProcess` in this PR was written to close ("the operator has to be able to tell this apart from a failure to start `claude`"), just applied inconsistently. Either have the fixture return a bare error (`errors.New("no space left on device")`) and assert the prefix that production actually supplies, or assert only the bare message here and add the prefix assertion in `TestNewWorkDir*` where the wrap lives.

---

internal/commandcenter/runner_branches_test.go
```diff
@@ -0,0 +1,230 @@
...
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "no entropy available")
+	assert.False(t, hasDone(events))
+	assert.False(t, *f.started)
+}
+
+// The work directory is created after the MCP config, so this also pins that
+// the config's cleanup still runs when a later step fails — otherwise the
+// token-bearing temp file that BuildMCPConfig writes (docs/security.md) would
+// outlive the query that needed it.
```
tomo-chan, line 124: Since this PR is explicitly reasoning about cleanup funcs outliving the query that needed them, the mirror case is worth adding while you're here: **`cleanupWorkDir` is asserted by no test in the package.** `NewWorkDir` is referenced only by `context_test.go:103` (which tests the helper in isolation) and by line 131 here (which fails it), so nothing pins that `run()` ever invokes the cleanup it gets back.

Verified: replace `defer cleanupWorkDir()` in `run()` with `_ = cleanupWorkDir` and the package stays green (`ok panemux/internal/commandcenter 0.218s`). The failure that hides behind is that every command-center query leaks a `panemux-command-center-*` directory into `/tmp` permanently — one per palette query, never reclaimed while the host is up.

A one-line addition to this test covers it (`NewWorkDir` returning a dir plus a cleanup that flips a bool, with the failure moved to a later seam), or a separate test alongside `TestRunnerAlwaysCallsMCPConfigCleanup`, which is the exact analogue for the other cleanup.

---

tomo-chan (commented):
Reviewed the diff and ran it against the implementation. The file is in good shape overall — I re-ran the perturbations for the blank-line skip, the history-failure arm and the session-id-failure arm, and each kills exactly the intended test, so those three are genuinely load-bearing. `go vet`, `gofmt -s` and `-race -count=2` are clean, and the dangling-symlink fixture does land on the intended `SaveSessionFile` arm (`MkdirAll` → `EEXIST`, load still `ErrNotExist`) as documented.

Three findings, all confirmed by perturbation rather than inspection — two are survivors in the sense this repo's `make efficacy`/mutation gates care about, and the third is an adjacent block this PR is well placed to pick up.

---

internal/commandcenter/runner_branches_test.go
tomo-chan: Confirmed with your exact perturbation before touching anything: moving `cmd.Start()` ahead of `cmd.StdoutPipe()` leaves the whole package green. You are also right about why `f.started` was dead — this test replaces `NewCommand` wholesale, so the fixture's recorder never runs.

Fixed in `92248e3`. `noStdoutCmd` records `Start`, and the test holds a reference to the instance so it can assert it:

```go
	assert.False(t, cmd.started,
		"the pipe is opened before Start, and it must stay that way: a subprocess started and then "+
			"abandoned on the pipe error is never Wait()ed, leaving a zombie claude per failed query, "+
			"while the deferred work-dir cleanup deletes the directory out from under it")
```

The reorder now fails exactly this test and no other. Your framing of the consequence went into the message rather than the comment, so it shows up in the failure output where someone debugging the reorder will actually read it.

---

internal/commandcenter/runner_branches_test.go
tomo-chan: Right, and it is the worse of the two failure modes you could have caught me in: the assertion was not merely weak, it was reading back the fixture's own string. Deleting `NewWorkDir`'s wrap leaves the package green — reproduced.

Took the second of your two options in `92248e3`, since it puts each assertion where the code that produces it lives:

- The fixture now returns a bare `errors.New("no space left on device")` and the test asserts only that, with a comment saying why the prefix is deliberately not asserted here.
- The prefix is asserted in a new `TestNewWorkDirFailureNamesTheStepThatFailed` beside `TestNewWorkDirIsEmptyAndRemovable`. It makes `MkdirTemp` fail for real by pointing `TMPDIR` at a path under a regular file, so it exercises the wrap rather than a stub of it. Dropping the wrap now fails that test.

You are also right that this was the same gap as the `TestRunnerBuildMCPConfigFailureEmitsErrorAndNeverStartsProcess` assertion I added in this PR, applied inconsistently — the difference is that `buildMCPConfig`'s prefix really is added by `run()`, while this one is not, and I did not check which was which.

---

internal/commandcenter/runner_branches_test.go
tomo-chan: Confirmed: `_ = cleanupWorkDir` in place of the defer leaves the package green. Good catch that this PR was the right place for it — it argues about `cleanup` outliving its query in one direction and left the mirror unpinned.

Added `TestQueryCleansUpTheWorkDirWhenTheTurnFails` in `92248e3` as its own test rather than folding it into the work-dir failure test, so it reads as the analogue of `TestRunnerAlwaysCallsMCPConfigCleanup` that you pointed at. `NewWorkDir` returns a directory plus a cleanup that flips a bool, and the failure moves to the stdout-pipe seam so the cleanup is reached on a path that still ends badly:

```go
	assert.True(t, cleaned, "the work directory must be removed however the turn ends")
```

Dropping the defer now fails exactly this test.

---

internal/commandcenter/runner_branches_test.go
```diff
@@ -0,0 +1,230 @@
...
+	f := newRunnerFixture(t,
+		`{"type":"system","subtype":"init","session_id":"s"}`+"\n"+
+			"\n"+
+			"   \n"+
+			`{"type":"result","session_id":"s","result":"done"}`+"\n")
+
+	events := f.query(t)
+
+	assert.Empty(t, errorEvents(events), "a blank line is padding, not malformed output")
+	assert.True(t, hasDone(events))
```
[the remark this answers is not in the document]
tomo-chan: Correct, and reproduced before changing anything: `continue` → `break` leaves the whole package green. The turn still reaches `done` with no error while every frame after the first blank line is dropped, so the test was asserting "blank lines are accepted" when its comment claimed "blank lines are skipped".

Fixed in `92248e3` — the test now asserts the frames that follow, which is the only place skipping and stopping differ:

```go
	require.Len(t, lines, 2, "the blank lines produce no frame of their own, and neither one ends the stream")
	assert.Contains(t, lines[1], `"type":"result"`,
		"the frame after the blank lines is the answer, and it must still reach the operator")
```

Renamed to `TestStreamOutputSkipsBlankLinesAndKeepsForwardingWhatFollows`, and the comment records that absence-of-error plus presence-of-done is not sufficient here, so it does not get trimmed back. The `break` perturbation now fails exactly this test.

---

internal/commandcenter/runner_branches_test.go
```diff
@@ -0,0 +1,230 @@
...
+// operator has the answer on screen either way.
+func TestQueryReportsAFailedHistoryWriteWithoutFailingTheTurn(t *testing.T) {
+	f := newRunnerFixture(t, `{"type":"result","session_id":"s","result":"done"}`+"\n")
+	f.config.HistoryPath = brokenParentPath(t, "history.jsonl")
+
+	events := f.query(t)
+
+	require.Len(t, errorEvents(events), 1)
+	assert.Contains(t, errorEvents(events)[0], "persisting command center history")
+	assert.True(t, hasDone(events), "the turn itself succeeded, so it must still report done")
```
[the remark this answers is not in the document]
tomo-chan: Correct, and it is worse than the doc reference alone shows — verified all three parts against the code rather than reasoning from the contract:

1. **[docs/behavior.md](docs/behavior.md)** does say each of `error` and `done` is "always the last frame for that query". Only one can be.
2. **`EventDone`'s own doc comment in `runner.go`** states the same invariant directly: *"It is never sent after an EventError on the same channel."* So this path violates the runner's own documented rule, not just the external contract.
3. **The UI consequence you predicted is real.** `applyFrame` in `useBoardCommand.ts` sets `updated.error` **and** `updated.done` on the error frame, and the `done` arm only sets `done` — it never clears the message. The turn therefore renders complete and errored, for a query that succeeded.

Not changing it here: this branch is deliberately test-only, and the fix is a frame-contract decision rather than a local correction — the frame vocabulary has no non-terminal channel to report a best-effort failure on. Filed as **#214** with the three options I can see (log only / a non-terminal `warning` frame / fold warnings into `done`).

The test keeps pinning the present shape, since that is what the code does today, but its comment now records the inconsistency rather than reading as endorsement — the old *"it must still report done"* was the wrong word:

```go
	assert.True(t, hasDone(events),
		"the turn itself succeeded, so it still reports done today — see #214 for why that, "+
			"alongside the error frame above, is the shape rather than the intent")
```

---

tomo-chan:
@codex review

Requesting a re-review on `92248e3`, since the two P2 findings from the first pass and three from the owner's review all landed after the reviewed commit `485f075`.

What changed, and what was verified rather than argued — each of these five perturbations left the whole package green before the fix, and now fails exactly one test:

| perturbation | now caught by |
|---|---|
| `continue` → `break` on the blank-line skip | `TestStreamOutputSkipsBlankLinesAndKeepsForwardingWhatFollows` |
| `cmd.Start()` moved before `cmd.StdoutPipe()` | `TestQueryReportsAFailureToOpenTheSubprocessStdoutPipe` |
| drop `NewWorkDir`'s error wrap | `TestNewWorkDirFailureNamesTheStepThatFailed` |
| drop `defer cleanupWorkDir()` | `TestQueryCleansUpTheWorkDirWhenTheTurnFails` |
| drop the `creating stdout pipe` wrap | `TestQueryReportsAFailureToOpenTheSubprocessStdoutPipe` |

The `error`-then-`done` frame pair you flagged is confirmed and left in place deliberately: it violates `docs/behavior.md`, `EventDone`'s own doc comment, and renders a successful turn as failed in the dashboard — but fixing it is a frame-contract decision and this branch is test-only, so it is filed as #214 and the test now records the inconsistency instead of endorsing it.

`make check` exit 0, `make lint-go` 0 issues, `-race -count=3` clean.

---

PR #224, opened 2026-09-07
refactor: give os.UserHomeDir a single injectable seam across the repository

### LINKED ISSUE #212: refactor: give os.UserHomeDir a single injectable seam across the repository
Split out of the review on #211, where it was raised against a test-only branch and scoped out of it.

## The rule

DEVELOPMENT.md's testability rule names this case explicitly:

> **Testability rule:** if production code calls `os.UserHomeDir()`, `DefaultPath()`, or another global singleton directly, add an injectable override so tests can substitute a controlled value without touching the real filesystem or home directory.

## The actual state

The rule is followed in exactly one place. `internal/session/local.go:27` has the seam the rule points at:

```go
var userHomeDirFn = os.UserHomeDir
```

Every other production call goes straight to `os.UserHomeDir()` — 16 sites across 9 files:

| file | sites |
|---|---|
| `internal/config/config.go` | 3 |
| `internal/api/handler.go` | 2 |
| `internal/session/ssh.go` | 2 |
| `board.go` | 1 |
| `internal/board/cursor_store.go` | 1 |
| `internal/board/bootstrap_store.go` | 1 |
| `internal/config/agent_board.go` | 1 |
| `internal/commandcenter/history_store.go` | 1 |
| `internal/commandcenter/context.go` | 1 |
| `internal/commandcenter/session_store.go` | 1 |
| `internal/session/factory.go` | 1 |

And 12 test files reach for `t.Setenv("HOME", ...)` instead, which is the workaround the rule exists to remove:

`main_test.go`, `board_test.go`, `board_branches_test.go`, `internal/server/api_integration_test.go`, `internal/server/ws_integration_test.go`, `internal/config/config_branches_test.go`, `internal/api/handler_test.go`, `internal/api/handler_error_paths_test.go`, `internal/commandcenter/context_test.go`, `internal/commandcenter/store_branches_test.go`, `internal/session/factory_test.go`, `internal/session/ssh_test.go`.

## Why it is worth doing

`t.Setenv` mutates process-wide state, so any package using it can never mark those tests `t.Parallel()` — and the constraint is invisible: nothing fails, the tests simply cannot be parallelized later without a subtle, intermittent breakage. `internal/commandcenter` alone has three defaults (`DefaultHistoryFilePath`, `DefaultSessionFilePath`, `DefaultContextDir`) that would share one seam.

Three of the sites also use `home, _ := os.UserHomeDir()` — discarding the error rather than handling it. Worth deciding deliberately during this pass whether each of those should keep swallowing it.

## Suggested shape

Follow the existing pattern rather than inventing a second one: a package-level `var userHomeDirFn = os.UserHomeDir` per package that needs it, overridden in tests. A single shared helper package would work too but pulls nine packages into a dependency they do not otherwise have — worth weighing, not assumed.

Mechanical and low-risk, but it touches nine packages, so it wants to be its own change with its own review rather than riding along with unrelated work.

## Note

Not a blocker for #211 or for the other slices of #195: those follow the prevailing precedent rather than departing from it, and `internal/commandcenter` was already non-parallel before them (`context_test.go:97` predates that work). This is about removing the precedent, repo-wide, in one place.

Related: #211, #195.

tomo-chan:
DEVELOPMENT.md's testability rule requires an injectable override wherever
production code reaches a global singleton. The home directory was the case the
rule names explicitly, and it was followed in exactly one place: 16 direct
os.UserHomeDir calls across nine packages, with 12 test files reaching for
t.Setenv("HOME", ...) instead (#212).

They now all go through internal/homedir: homedir.Dir(), substituted in tests
with homedir.SetForTest / homedir.SetFailingForTest.

One package rather than one variable per caller, which is the shape the issue
suggested. The callers do not line up with the tests: internal/api's handler
tests drive tilde expansion that happens inside internal/config,
internal/server's integration tests drive config, api and session at once, and
the root package's tests drive internal/commandcenter's default paths. A
package-private variable is invisible from another package's test, so each of
those would have kept mutating $HOME — the thing the seam exists to remove.
internal/homedir depends on nothing but os, and declares its own two-method
TestingT rather than importing testing, so no binary linking it links testing.

.golangci.yml gains a forbidigo rule failing the build on os.UserHomeDir
outside internal/homedir. That is the part that keeps "one seam" true: the
testability rule had been advice long enough to accumulate 16 violations
without anyone noticing. The one deliberate exception is annotated in place —
contract_fixture_test.go's subject IS the capturing machine's own home.

The issue also asked for the four `home, _ := os.UserHomeDir()` sites to be
decided rather than inherited. All four discarded the error and then joined
against the empty string it left, and filepath.Join("", ".ssh/id_ed25519") is
".ssh/id_ed25519" — a path relative to whatever directory panemux was started
in. The decision is to keep swallowing the error (no caller has anywhere to put
one) but never to join against an empty home:

- buildAuthMethods probed for OpenSSH default keys under an empty home, so with
  no home directory it read ./.ssh/id_ed25519 and authenticated the outbound
  SSH connection with it. It now skips the probe, surfacing the existing "no
  auth methods" error.
- resolveSSHConfig expanded an ssh_config IdentityFile the same way, in both
  its ~/-prefixed and its relative arm. The path is left as written.
- internal/config's two expansions (SSH key/known_hosts files, pane cwd) are
  now one expandTilde that leaves the ~/ in place — which is what board.go's
  expandLocalAgmsgPath already documented itself as matching, and did not.
- sshconfig.DefaultPath keeps its relative fallback, and now has a test saying
  why: a config file whose absence ParseHosts already reports as an empty host
  list is a different consequence from a private key.

Those failure arms had never been executed by a test, because nothing could
reach them. They are covered now, along with the default-path arms in
internal/board and internal/commandcenter, and scenario A7 records the
use case.

Docs updated: DEVELOPMENT.md (the rule, the lint gate, the coverage list),
docs/architecture.md (the package), docs/security.md (the SSH key finding),
docs/quality-gateway.md (P3's evidence), docs/scenarios.md (A7).

Closes #212

Claude-Session: https://claude.ai/code/session_01RTnhzeMcS5jVpZAJkpq4HJ


internal/homedir/homedir_test.go
```diff
@@ -0,0 +1,103 @@
...
+// reported at the caller's line rather than inside homedir.
+func TestHelpersMarkThemselvesAsHelpers(t *testing.T) {
+	spy := &helperSpy{}
+
+	homedir.SetForTest(spy, "/workspace/user/home")
+	homedir.SetFailingForTest(spy, errors.New("no home directory"))
+
+	assert.Equal(t, 2, spy.helperCalls)
+	require.Len(t, spy.cleanups, 2)
+	for _, cleanup := range spy.cleanups {
```
tomo-chan: The spy's cleanups are run in FIFO order, which leaves `dirFn` permanently substituted for the rest of the test binary.

`cleanups[0]` was registered by `SetForTest` and restores `dirFn` to the real `os.UserHomeDir`. `cleanups[1]` was registered by `SetFailingForTest`, and the `original` it captured is `SetForTest`'s closure. Running them in index order therefore ends with `dirFn` set to `func() { return "/workspace/user/home", nil }`, not the OS value — and because `spy` is not a `*testing.T`, no real `t.Cleanup` ever puts it back.

Reproduced on this branch:

```
$ go test ./internal/homedir/ -count=1 -shuffle=on   # 6 runs
FAIL  ok  FAIL  FAIL  ok  ok
```

with

```
--- FAIL: TestDirDefaultsToTheOperatingSystemHomeDirectory
        -/root
        +/workspace/user/home
```

It passes today only because the default (source-file) ordering happens to run that test first. `go test -shuffle=on` fails roughly half the time, and any test added above it in this file fails permanently.

`testing.T` unwinds cleanups LIFO, so the spy should too:

```go
for i := len(spy.cleanups) - 1; i >= 0; i-- {
	spy.cleanups[i]()
}
```

That is also the order `TestNestedSubstitutionsRestoreTheEnclosingValue` relies on, so matching it here keeps the spy an honest stand-in for `*testing.T`.

---

.golangci.yml
```diff
@@ -28,6 +29,18 @@ linters:
   settings:
     dupl:
       threshold: 100
+    forbidigo:
+      # The home directory is a process-wide global, and DEVELOPMENT.md's
+      # testability rule requires an injectable override for those. Sixteen
+      # call sites across nine packages ignored it (issue #212), and nothing
+      # noticed, because "add a seam" was advice rather than a check.
+      forbid:
+        - pattern: ^os\.UserHomeDir$
```
tomo-chan, line 43: This rule is bypassable by an import alias, which undercuts the claim the PR body, `internal/homedir`'s package doc and DEVELOPMENT.md all make for it — that the seam "cannot quietly stop being the single entry point".

`forbidigo`'s `analyze-types` defaults to `false`, so the pattern is matched against **source text**, not against the resolved identifier. Verified live against this branch's config, on a scratch package:

| code | result |
|---|---|
| `import "os"` + `os.UserHomeDir()` | caught — `use of os.UserHomeDir forbidden because …` |
| `import osx "os"` + `osx.UserHomeDir()` | **0 issues** |
| same, with `analyze-types: true` | caught — `use of osx.UserHomeDir forbidden because …` |

One line fixes it:

```yaml
    forbidigo:
      analyze-types: true
      forbid:
        - pattern: ^os\.UserHomeDir$
```

This is worth doing rather than waving off as "nobody would alias `os`", because the gate's whole justification in the PR body is that the rule had been *advice* long enough to accumulate 16 violations without anyone noticing. A gate with a one-keyword hole is advice again for anyone who happens to write the aliased form.

(Checked at the same time: with `analyze-types: true` the `internal/homedir/` exclusion still scopes correctly, and the rest of the tree stays at 0 forbidigo issues.)

---

internal/homedir/homedir.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// This replaces t.Setenv("HOME", ""), which reached the no-home-directory arms
+// only on Unix, and only as a side effect of os.UserHomeDir rejecting an empty
+// $HOME there. Injecting the error states the intent instead of relying on it.
+func SetFailingForTest(t TestingT, err error) {
+	t.Helper()
+	setForTest(t, func() (string, error) { return "", err })
+}
+
+func setForTest(t TestingT, fn func() (string, error)) {
```
tomo-chan, line 74: `dirFn` is an unsynchronized package global, and `SetForTest`'s own doc comment sells the helper partly on a property it does not deliver.

Lines 44-48 say:

> The difference is not only style: t.Setenv mutates state the whole test binary shares, it is **refused outright in a test that has called t.Parallel** […]

`t.Setenv`'s refusal is a *feature* — it panics loudly rather than letting two parallel tests fight over one process-wide value. `setForTest` has exactly the same sharing problem (`dirFn` is one variable for the whole binary) but no such guard: under `t.Parallel` one test's `Cleanup` write races another's `Dir()` read, and the loser silently gets the wrong home directory instead of a panic. That is a strictly worse failure mode than the one the comment cites as a reason to prefer this helper.

Nothing fails today — there is no `t.Parallel` anywhere in the repo — but the doc comment (and the matching DEVELOPMENT.md text) actively invites someone to add one, since "works under `t.Parallel`" reads as the advertised advantage.

Either close the gap or stop claiming it. Cheapest honest fix is to drop the `t.Parallel` clause from both texts. If you'd rather keep the claim, `dirFn` needs a `sync.RWMutex` (or an `atomic.Pointer`) around the read in `Dir` and the writes here — noting that even then, nesting/restore is still last-writer-wins across parallel tests, so a mutex makes it race-free without making it *correct* for concurrent substitution. Given that, I'd recommend the doc fix over the mutex.

---

internal/server/api_integration_test.go
```diff
@@ -103,7 +104,7 @@ func newAPIEnv(t *testing.T) *apiEnv {
 	t.Helper()
 
 	home := t.TempDir()
-	t.Setenv("HOME", home)
+	homedir.SetForTest(t, home)
```
tomo-chan, line 114: Replacing `t.Setenv("HOME", home)` with the seam here also drops the `$HOME` override, and on macOS that reintroduces exactly the pollution the comment block above this function says the next line exists to prevent.

`os.UserCacheDir` only consults `XDG_CACHE_HOME` on non-darwin Unix. On darwin it returns `$HOME/Library/Caches` unconditionally. So with `HOME` left pointing at the developer's real home, creating a local pane in this suite installs the browser shim (`session.installLocalBrowserShim` → `userCacheDirFn` in `internal/session/browseropen.go:85`) into the real `~/Library/Caches/panemux/bin` — the three real files the comment at lines 95-100 was written to stop.

The seam is the right change for the *config/ssh-config* half; it just isn't a substitute for `HOME` where the code under test reaches `os.UserCacheDir` rather than `os.UserHomeDir`. Keeping both lines is enough:

```go
home := t.TempDir()
homedir.SetForTest(t, home)
t.Setenv("HOME", home) // os.UserCacheDir ignores XDG_CACHE_HOME on darwin
```

CI is Linux-only, so this will not go red — it only bites whoever runs the suite on a Mac.

(Worth noting `internal/session` already has a `userCacheDirFn` seam of its own; promoting it the way this PR promoted the home-directory one would remove the need for either env var here. Out of scope for this PR, but it's the shape that actually closes this.)

---

internal/server/ws_integration_test.go
```diff
@@ -77,7 +78,7 @@ func newWSEnv(t *testing.T, runner *commandcenter.Runner) *wsEnv {
 func newWSEnvIn(t *testing.T, home string, runner *commandcenter.Runner) *wsEnv {
 	t.Helper()
 
-	t.Setenv("HOME", home)
+	homedir.SetForTest(t, home)
```
tomo-chan, line 81: Same lost `$HOME` override as in `newAPIEnv` (see the comment on `internal/server/api_integration_test.go:107`), with the same darwin consequence: `os.UserCacheDir` ignores `XDG_CACHE_HOME` on macOS and resolves `$HOME/Library/Caches`, so a local pane created through this env writes the browser shim into the developer's real cache directory.

Keep `t.Setenv("HOME", home)` alongside `homedir.SetForTest(t, home)`.

---

internal/config/config_branches_test.go
```diff
@@ -40,11 +42,11 @@ func TestLoad_MissingFile_ReportsAReadFailure(t *testing.T) {
 	assert.Contains(t, err.Error(), "reading config")
 }
 
-// LoadOrDefault resolves the path from $HOME, so both of its arms are reachable
-// by pointing $HOME somewhere this test owns.
+// LoadOrDefault resolves the path from the home directory, so both of its arms
+// are reachable by pointing the userHomeDirFn seam somewhere this test owns.
```
tomo-chan: Stale identifier: `userHomeDirFn` was `internal/session`'s package-private variable, and this commit deletes it. `grep -rn userHomeDirFn .` on this branch returns exactly one hit — this comment line.

The sibling comments this PR touches in `board_test.go` / `board_branches_test.go` say "the home-directory seam"; matching them keeps the wording accurate without naming anything:

```go
// LoadOrDefault resolves the path from the home directory, so both of its arms
// are reachable by pointing the home-directory seam somewhere this test owns.
```

---

tomo-chan (commented):
Code review at `medium` effort, against `origin/main` (`598611e`), on head `7d20bca` — 49 files, +731/−154.

**The refactor itself is sound.** The production conversions are faithful, and the four `home, _ := os.UserHomeDir()` decisions are the real value here. `buildAuthMethods` no longer probes `./.ssh/id_ed25519` out of the working directory and authenticates an outbound SSH connection with whatever it finds — that is a genuine credential-confusion fix, and `TestBuildAuthMethods_UnresolvableHomeDirectory_DoesNotProbeTheWorkingDirectory` pins it properly by planting a decoy key under `t.Chdir`. `expandIdentityFile`, `expandTilde` and the `defaultKeyAuthMethods` extraction check out branch-by-branch against the deleted code.

**Findings** (6 inline, none blocking on their own; the first two are the ones I'd fix before merge):

1. `internal/homedir/homedir_test.go:92` — the spy runs cleanups FIFO, permanently substituting `dirFn` for the rest of the binary. **Reproduced:** `go test ./internal/homedir/ -shuffle=on` fails 3 of 6 runs.
2. `.golangci.yml:38` — `analyze-types` defaults to false, so the gate is source-text-only and an import alias walks past it. **Reproduced:** `import osx "os"` + `osx.UserHomeDir()` → 0 issues; with `analyze-types: true` → caught. The PR body, the package doc and DEVELOPMENT.md all claim the seam is un-bypassable.
3. `internal/homedir/homedir.go:65` — `dirFn` is an unsynchronized global while the doc cites `t.Setenv`'s `t.Parallel` refusal as a reason to prefer this helper. Latent (no `t.Parallel` in the tree), but a loud panic was traded for a silent wrong answer.
4. `internal/server/api_integration_test.go:107` and 5. `ws_integration_test.go:81` — the dropped `t.Setenv("HOME", …)` reintroduces real-cache-directory pollution on macOS, where `os.UserCacheDir` ignores `XDG_CACHE_HOME`. Linux CI will not see it.
6. `internal/config/config_branches_test.go:46` — comment names `userHomeDirFn`, which this commit deletes; it is the only remaining reference in the tree.

**Verification run** (Linux, this branch): `go build ./...`, `go vet ./...` clean; `go test ./...` green; `go test -race ./internal/... .` green; `golangci-lint run ./...` 0 issues; `make check-scenarios` resolves all 198 names across 91 `auto` rows, including the new A7 row. The forbidigo gate and the shuffle failure were both confirmed by perturbation rather than read off the config.

---

internal/config/config.go
```diff
@@ -564,10 +561,31 @@ func expandPanesCwd(children []LayoutChild) {
...
+// wrong file rather than failing where an operator can see it. Leaving the ~/
+// in place is the diagnosable answer, and the one board.go's
+// expandLocalAgmsgPath already describes itself as matching.
+func expandTilde(path string) string {
+	if !strings.HasPrefix(path, "~/") {
+		return path
+	}
+	home, err := homedir.Dir()
+	if err != nil {
+		return path
```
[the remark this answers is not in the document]
tomo-chan: Confirmed and fixed in 633094e — you were right, and the branch's own test and docs were asserting the bug.

I planted readable files at both paths to check rather than reason about it:

```
.ssh/id_ed25519:   READ SUCCESSFULLY from the working directory
~/.ssh/id_ed25519: READ SUCCESSFULLY from the working directory
```

So `TestResolveSSHConfig_UnresolvableHomeDirectory_LeavesTheIdentityFileAlone` pinned "leaves it alone" and read as if that were the safety property, and `docs/security.md` said the fix closed the working-directory read. Neither was true. Worth noting the bare-relative arm (`IdentityFile .ssh/id_ed25519`) is the more dangerous of the two, since `.ssh` is an ordinary directory name.

Rather than encode failure in the value — `expandPaths` runs on `Load` and its result is what `write()` persists, so emptying or marking the field would corrupt the operator's `config.yaml` on the next save — the guard is at the read. `requireAbsolutePath` in `internal/session/ssh.go` refuses a non-absolute path in `buildAuthMethods` (`KeyFile`) and in `resolveKnownHostsFile`. Every route that produces these paths yields an absolute one when it works, so a relative path at the read means an expansion that could not happen. That is also the absolute-path-first shape `validateShell` and `validRemotePath` already use.

I included `known_hosts` because it is the same defect with a different consequence — a file in the working directory deciding host-key verification — and fixing only the key half would have been the same overclaim you just caught.

The new test plants a readable key at each relative path first, so it fails if the guard is removed rather than merely failing to find a file. `docs/security.md`, scenario A7 and `docs/behavior.md`'s `key_file` / `known_hosts_file` rows are corrected, including an explicit note that a deliberately working-directory-relative path in `config.yaml` is now refused.

---

internal/homedir/homedir_test.go
tomo-chan: Reproduced exactly as described and fixed in 633094e. Before, 6 runs of `go test ./internal/homedir/ -count=1 -shuffle=on`:

```
FAIL  ok  FAIL  FAIL  ok  ok
--- FAIL: TestDirDefaultsToTheOperatingSystemHomeDirectory
        -/root
        +/workspace/user/home
```

After, 8/8 `ok`. The spy now unwinds LIFO, with the reason in a comment so it does not get "tidied" back to a range loop.

---

.golangci.yml
tomo-chan: Confirmed independently and fixed in 633094e. Same scratch-package check, with this branch's config:

| code | result |
|---|---|
| `import "os"` + `os.UserHomeDir()` | caught |
| `import osx "os"` + `osx.UserHomeDir()` | **0 forbidigo issues** |
| same, `analyze-types: true` | caught — `use of osx.UserHomeDir forbidden` |

`analyze-types: true` added, with the reason inline. Agreed on why it matters rather than being pedantry: the gate exists precisely because "advice" let 16 violations accumulate, and a one-keyword hole is advice again. Whole tree is still 0 forbidigo issues and `make check` is green.

---

internal/homedir/homedir.go
tomo-chan: Agreed, and I took your recommendation — doc fix, not a mutex — in 633094e. Your reasoning is the deciding part: a mutex would make it race-free without making it correct, since nesting/restore stays last-writer-wins across parallel tests, so it would buy the appearance of the guarantee rather than the guarantee.

`SetForTest`'s doc now separates what the seam buys (process environment untouched, works on Windows where `os.UserHomeDir` reads `USERPROFILE`, failures injected directly instead of conjured from an empty `$HOME`) from what it does not, and says outright that `t.Setenv` is better in one respect here because it panics rather than racing. DEVELOPMENT.md carries the same correction.

---

internal/server/api_integration_test.go
tomo-chan: Fixed in 633094e — both lines kept, as suggested. The comment block above was also stale: it still claimed "HOME points at a temp directory" while the code no longer set it, so it now says why both variables are needed rather than just listing them.

The root of it is that the directory at risk is the *cache* directory, and `os.UserCacheDir` never consults `os.UserHomeDir` at all — it reads `$HOME` itself on darwin and `XDG_CACHE_HOME` elsewhere — so the home-directory seam cannot reach it by construction. Your follow-up shape is recorded in the comment: promoting `internal/session`'s `userCacheDirFn` the way this PR promoted the home seam is what removes both env vars. Agreed it is out of scope here.

---

internal/server/ws_integration_test.go
tomo-chan: Fixed in 633094e alongside `newAPIEnv`, with a short comment pointing at the explanation there rather than repeating it.

---

internal/config/config_branches_test.go
tomo-chan: Fixed in 633094e, using your wording so it matches the sibling comments. `grep -rn userHomeDirFn` over the branch now returns nothing at all, in `.go` and `.md` alike.

---

internal/homedir/homedir_test.go
```diff
@@ -0,0 +1,103 @@
...
+		cleanup()
+	}
```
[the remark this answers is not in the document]
tomo-chan: Fixed in 633094e — the spy now unwinds LIFO like `testing.T`. Reproduced first: `go test ./internal/homedir -count=1 -shuffle=on` failed 3 of 6 runs before, passes 8 of 8 after. See the reply on the adjacent thread for the detail.

---

PR #228, opened 2026-09-10
fix: report a failed command-center history write as a warning on the done frame

### LINKED ISSUE #214: fix: a failed history write sends both an error and a done frame, and the UI shows a successful answer as failed
Found by the Codex review on #213, confirmed against the code, the documented contract and the frontend.

## What happens

When a command-center query succeeds but persisting its history fails, `finishAfterStream` emits an `EventError` and then carries on to `EventDone`:

```go
if len(p.historyEntries) > 0 {
    if err := AppendHistory(r.historyPath, p.historyEntries); err != nil {
        events <- errorEvent("persisting command center history: %v", err)
    }
}
// ... no return; execution continues to
events <- Event{Type: EventDone}
```

So the WS client receives `{"type":"error",...}` followed by `{"type":"done"}`.

## Why that is wrong three times over

**1. It breaks the documented frame contract.** [docs/behavior.md](docs/behavior.md) describes both as terminal:

> `{"type":"error","message":"..."}` — the query failed …; **always the last frame for that query**
> `{"type":"done"}` — the query finished successfully; **always the last frame for that query**

Only one of the two can be last, and here neither is.

**2. It breaks the runner's own stated invariant.** `EventDone`'s doc comment in `internal/commandcenter/runner.go` says so directly:

> EventDone marks a successful query's end. **It is never sent after an EventError on the same channel.**

**3. The operator sees a successful answer reported as failed.** `applyFrame` in `frontend/src/hooks/useBoardCommand.ts` sets both fields on the error frame, and the later `done` frame does not clear them:

```ts
case 'error':
  updated.error = frame.message
  updated.done = true
  break
case 'done':
  updated.done = true
  break
```

The turn therefore renders complete *and* errored. The query ran, the answer is on screen, and the only thing that actually failed was writing a record of it.

## The design question underneath

A failed history write genuinely is best-effort — losing the record should not fail the turn, and the current code is right to continue rather than return. What it lacks is a way to *say* that. The frame vocabulary has no non-terminal channel: `line`, `error`, `done`, `busy`, and three of the four end the turn.

Options, roughly in increasing cost:

- **Log it and emit nothing.** Smallest change, honest about the contract, but the operator never learns their history is not being written.
- **Add a non-terminal `warning` frame.** Fits the actual meaning. Costs a frame type, a Zod schema change, a contract fixture and a UI affordance.
- **Fold it into the `done` frame** (e.g. `{"type":"done","warnings":[...]}`). No new frame type, and terminal-means-terminal stays true.

I have no strong view on which; the frame contract is the part that needs deciding, not the persistence behavior.

## Note on tests

`internal/commandcenter/runner_branches_test.go`'s `TestQueryReportsAFailedHistoryWriteWithoutFailingTheTurn` currently pins the present behavior — an error event *and* a done event. That is deliberate for a test-only branch, and its comment records the inconsistency rather than endorsing it, but it will need updating with whichever option is chosen.

Related: #195 (the branch-coverage work that surfaced this), #213.

tomo-chan:
A command-center query whose history write failed emitted an EventError and
then carried on to EventDone. Both frames are documented as "always the last
frame for that query" (docs/behavior.md), EventDone's own doc comment
promised it "is never sent after an EventError on the same channel", and the
dashboard's applyFrame set `error` on the error frame without the later done
frame clearing it — so a query that succeeded, with its answer already on
screen, rendered as complete *and* errored.

Losing the record of a turn is genuinely not a failed turn, so continuing
rather than returning was right; what the frame vocabulary lacked was a way
to say it. Of the three options in the issue, this takes the third: fold the
failure into the terminal frame as `{"type":"done","warnings":[...]}`. No new
frame type, and terminal-means-terminal stays true — exactly one of `error`
and `done` ends a query.

- commandcenter.Event gains Warnings, set only on EventDone. A turn that ends
  in an EventError carries none: the error is the actionable frame. Either
  way the failure is logged, so it is never lost silently.
- The WS frame gains `warnings`, omitted when empty, so an ordinary turn's
  frame is still the bare {"type":"done"}.
- The palette renders warnings under the answer, in amber rather than the
  error red, and no longer marks the turn errored.
- The board-command contract fixture captures the warning-bearing done frame
  alongside the plain one, so the new optional field is exercised by JSON Go
  actually emitted rather than declared unexercised.

Closes #214.

Claude-Session: https://claude.ai/code/session_01RCejZLHciBscXbEkbyg4ne


internal/server/contract_fixture_test.go
```diff
@@ -515,9 +521,46 @@ func captureBoardCommandFrames(t *testing.T) []byte {
...
+		},
+	})
+
+	e := newWSEnv(t, runner)
+	conn, _ := e.dial(t, "/ws/board-command", integrationToken)
+	require.NotNil(t, conn)
+
+	require.NoError(t, conn.WriteMessage(websocket.TextMessage, []byte(`{"prompt":"which panes are working?"}`)))
+	readRawControl(t, conn) // line
+	return readRawControl(t, conn), dir
```
tomo-chan: **The capture never asserts that the frame it returns is a `done` frame carrying `warnings`.**

Both `readRawControl` calls take whatever the next text frame happens to be; the `// line` comment and the caller's `// done, carrying a warning` comment are the only things claiming what each one is. Concretely: if this Runner ever emits an `error` before reaching `done` — `BuildMCPConfig` failing, `LoadSessionFile` failing, `fixtureClaudeScript` growing a second stream-json line — then `testdata/api-contract/ws-board-command-frames.json` silently records an `{"type":"error",...}` frame in the slot documented as the warning-bearing `done`, and the Go suite stays green because it asserts nothing about the bytes. The fixture is the artifact a reviewer reads as the contract, so it would then assert a shape this path never produces.

The same gap hides the regression this capture exists to catch: if Go stops populating `warnings`, this writes a second plain `{"type":"done"}`, the Go suite passes, and the only failure is `ws-board-command-frames[].warnings` turning up in `unexercisedOptionalPaths()` over in `frontend/src/schemas/contract.test.ts` — a red test one language and one suite away from the cause.

One assertion here pins it at the source:

```suggestion
	readRawControl(t, conn) // line
	frame := readRawControl(t, conn)
	require.Contains(t, string(frame), `"type":"done"`,
		"the second frame must be the terminal done, not an error from an earlier arm")
	require.Contains(t, string(frame), `"warnings"`,
		"the whole point of this capture is the populated optional; see #214")
	return frame, dir
```

---

frontend/src/components/CommandPalette.tsx
```diff
@@ -212,6 +217,9 @@ const toolLineStyle: React.CSSProperties = {
 const lineStyle: React.CSSProperties = { color: '#ccc', whiteSpace: 'pre-wrap', wordBreak: 'break-word' }
 
 const errorLineStyle: React.CSSProperties = { color: '#f44747' }
+// Amber rather than the error red: the turn itself succeeded, and the answer
+// is right above this line. See #214.
+const warningLineStyle: React.CSSProperties = { color: '#d7ba7d' }
```
tomo-chan: **`warningLineStyle` is missing the wrapping properties its siblings carry, and the warning is the one line guaranteed to hold a long unbreakable token.**

`lineStyle` and `toolLineStyle` both set `whiteSpace: 'pre-wrap'` and `wordBreak: 'break-word'`; this one sets only a color. The string it renders is built in `finishAfterStream` as `persisting command center history: %v`, and that `%v` is always a filesystem path — e.g. `creating command center history directory: mkdir /home/<user>/.config/panemux/command-center: permission denied`. A path is a single token with no break opportunity.

The palette is a fixed `width: '560px'` column with `padding: '16px 20px'`, so the content box is ~520px, and `historyStyle` sets `overflowY: 'auto'` only — `overflow-x` stays `visible`. An operator with a long `$HOME` (a corporate `/Users/firstname.lastname/`, a container path, or `XDG_CONFIG_HOME` pointed somewhere deep) pushes that token past 520px and the warning spills out of the palette box rather than wrapping — and the spilled half is the part that says *which* path failed, which is the only actionable content in the message.

```suggestion
const warningLineStyle: React.CSSProperties = {
  color: '#d7ba7d',
  whiteSpace: 'pre-wrap',
  wordBreak: 'break-word',
}
```

(`errorLineStyle` has the same gap, but that is pre-existing and its messages are not path-shaped by construction the way these are.)

---

internal/server/contract_fixture_test.go
tomo-chan: Fixed in a09e3b1. Agreed on both halves — the capture asserted nothing about the bytes it wrote, so the fixture was documenting a shape by comment rather than by check.

One deviation from the suggestion, and the reason for it: it decodes the frame instead of substring-matching. `require.Contains(string(frame), `"warnings"`)` passes on `"warnings":[]`, which `omitempty` is the only thing keeping off the wire today — one struct-tag edit away from being reachable. `require.NotEmpty(decoded.Warnings)` fails on it.

Verified by perturbation rather than assumed: dropping `Warnings` from the `EventDone` in `finishAfterStream` fails this capture in the Go suite —

```
--- FAIL: TestAPIContractFixtures/ws-board-command-frames
    Error: Should NOT be empty, but was []
    Messages: the whole point of this capture is the populated optional; see #214
```

— which is the failure moving to the cause, one suite and one language earlier than `unexercisedOptionalPaths()`.

The earlier `error`-instead-of-`done` case is covered by the same assertion: an error frame ahead of `done` either fails `require.Equal(t, "done", decoded.Type)` or, if it was the only frame, leaves the second `readRawControl` to fail on its own read deadline.

---

frontend/src/components/CommandPalette.tsx
tomo-chan: Fixed in a09e3b1, applied as suggested.

The point about *which* line this lands on is the part worth recording, so it went into the comment above the style rather than only into the diff: the warning is the one line whose content is a path by construction (`finishAfterStream` builds it as `persisting command center history: %v`, and every `%v` on that path is an `os.MkdirAll`/`os.OpenFile` error naming a file), and `historyStyle` scrolls only vertically, so the overflowing half is the actionable half.

`CommandPalette.test.tsx` now drives the case with a realistic warning — `mkdir /workspace/user/project/.config/panemux/command-center: permission denied`, a path-shaped token rather than the previous `disk full` — and pins `whiteSpace`/`wordBreak` on the rendered element, so removing the properties goes red rather than merely looking wrong. Confirmed under `make efficacy`: that test is one of the 9 that go red with the branch's implementation reverted.

`errorLineStyle` left alone, as you noted it is pre-existing and not path-shaped.

---

internal/commandcenter/runner.go
```diff
@@ -317,9 +332,18 @@ type finishParams struct {
 func (r *Runner) finishAfterStream(ctx context.Context, cmd cmdRunner, p finishParams, events chan<- Event) {
 	waitErr := cmd.Wait()
 
+	// A lost history record does not fail the turn — the operator already
+	// has the answer — so this collects a warning for the EventDone below
+	// rather than emitting an EventError of its own. Every arm that returns
```
tomo-chan: This comment is inaccurate about one of the three arms, and the inaccuracy hides the case worth thinking about.

> Every arm that returns before that point ends the turn with an EventError instead

The `scanFailed` arm two lines below does **not** end the turn with an `EventError` — it returns having emitted nothing at all. The error it refers to was sent by `streamOutput` *before* `cmd.Wait()` and before this history write, so by the time the warning exists the client has already received its terminal frame and stopped listening for anything about this turn.

The timeout and `waitErr` arms do match the comment. The session-id-save arm at the bottom (`persisting command center session id`) is a fourth one that also discards `warnings` and isn't mentioned.

Worth tightening the wording to "the turn has already ended with an EventError, or ends with one below" so a reader doesn't go looking for an emit in the `scanFailed` branch that isn't there.

---

docs/behavior.md
```diff
@@ -659,6 +661,21 @@ client that stops reading without closing its TCP connection (a sleeping laptop,
 with no FIN) fails the write — and falls into the same drain-without-forwarding path described above —
 instead of blocking the server goroutine forever.
 
+**Exactly one of `error` and `done` ends a query, and `warnings` is how the successful one reports a
```
tomo-chan: **"Exactly one of `error` and `done` ends a query" is not true of the `busy` frame, and this paragraph is the place a client author would read the frame contract from.**

`handlePrompt` (internal/ws/board_command.go) returns immediately after writing `{"type":"busy"}` — `Query` returned `ErrBusy`, so there is no event channel and no later frame for that prompt. A `busy` prompt therefore receives *neither* `error` nor `done`.

Concrete failure: someone writing a non-browser client against this section takes the new sentence literally, treats `busy` as informational, and keeps waiting for a terminal frame that never arrives — the request hangs until the read deadline. The dashboard doesn't hit this only because `applyFrame`'s `busy` arm independently sets `done = true`, which this paragraph doesn't mention.

Suggest either scoping the sentence to a query that actually started ("a query that reaches the subprocess ends with exactly one of `error` and `done`; a prompt rejected with `busy` gets neither"), or naming `busy` as the third terminal frame.

---

internal/commandcenter/runner.go
```diff
@@ -317,9 +332,18 @@ type finishParams struct {
...
+	// before that point ends the turn with an EventError instead, and drops
+	// the warning from the wire deliberately; the log line is what keeps it
+	// from being lost entirely, on that path and on the successful one alike.
+	var warnings []string
```
tomo-chan, line 347: **The warning is collected before every failure arm and then discarded by all of them, which trades the #214 double-frame bug for a silent one on the error paths.**

Before this change a failed history write always reached the operator, on every path, as an `error` frame. Now it reaches them only when the turn also succeeds. Four arms below drop it: `scanFailed`, the timeout, `waitErr`, and the session-id save failure.

Scenario: `~/.config/panemux/command-center` is read-only (or on a full disk), and the operator's queries are timing out for an unrelated reason — a slow MCP call, a long question. Every turn produces `{"type":"error","message":"claude query timed out after 5m0s"}` and nothing else. The command center's history file has silently stopped recording every one of those turns, the palette seeds itself from `/api/board/command/history` on each open and keeps showing nothing, and the only place that says why is a `log.Printf` on the server's stdout — which, for the common `panemux --open` desktop case, is a terminal nobody is reading. The operator's evidence is "the palette forgets everything", with no path from that to the actual cause.

I can see the argument in the doc comment ("the error is the actionable frame") and it is right that a *second* terminal frame was the bug. But the frame that ends the turn is already being constructed on each of these arms, and appending the warning to it costs nothing structurally — `Event.Warnings` could ride `EventError` too, with `boardCommandFrame.Warnings` and the Zod `error` member widened to match. That keeps "exactly one terminal frame" (the property #214 asked for) while keeping the failure visible on every path (the property the issue also asked for: "not something to hide either").

At minimum, worth stating in docs/behavior.md that on an error path the *only* record is panemux's own log, so an operator debugging a missing history has somewhere to look.

---

tomo-chan: Took the fuller option, not the documentation-only minimum. You are right that this was a trade rather than a fix — the read-only-config-plus-timing-out-queries scenario is the exact operator the warning exists for, and they were the one it stopped reaching. Fixed in 9f041d8: `Warnings` now rides whichever terminal event the turn ends with.

Covering `scanFailed` needed one structural change rather than just a wider field, which is worth flagging since it touches documented behavior. `streamOutput` returned `failed bool` and emitted its own `EventError`, so that turn-ending frame was sent *before* `cmd.Wait()` and before the history write had run — there was nothing for a warning to attach to. It now returns the error and `finishAfterStream` emits it, which makes `finishAfterStream` the single owner of every terminal event a turn produces.

What that changes, stated plainly: the malformed-line error frame is now sent just *after* the cancellation instead of just before it. `run()` still calls `cancel()` the instant `streamOutput` returns non-nil, so the busy-flag release this was designed around happens at exactly the same moment — only the frame is later, by however long a killed subprocess takes to reap. behavior.md's "The client has already received the corresponding error frame by the time this happens" was true and is now false, so that sentence is rewritten rather than left to rot. `TestRunnerMalformedStreamJSONCancelsQueryContextImmediately` still passes unchanged, which is the property that actually mattered.

Ordering: `p.streamErr` is checked before `waitErr`, because a subprocess killed by that `cancel()` exits non-zero *because of the kill* — reporting the exit status would name the consequence instead of the cause.

Tests follow the arms rather than one of them. `TestQueryAttachesAFailedHistoryWriteToWhicheverTerminalEventEndsTheTurn` is a table over all five terminal events — done, non-zero exit, malformed line, timeout, failed session-id write — each with a broken history path, asserting the single terminal event carries the warning. It went red on four of the five before the fix. `TestQueryAlsoLogsAFailedHistoryWrite` keeps the log assertion separate, so the frame carrying the text is not a reason to stop writing it down.

One gap I should name rather than let you find: `contract.test.ts`'s optional-field walker keys paths by field name, so `ws-board-command-frames[].warnings` is already satisfied by the `done` capture — it does not independently prove the `error` member populates. The `internal/ws` table and the hook tests are what cover that half.

---

internal/commandcenter/runner.go
tomo-chan: Correct, and the inaccuracy did hide the case worth thinking about — it turned out to be the same one raised in the thread on line 341, so it is fixed by construction rather than by rewording.

`streamOutput` no longer emits the malformed-line `EventError`; it returns the error and `finishAfterStream` emits it. The `scanFailed` arm the comment was wrong about no longer exists — there is no longer any path out of `finishAfterStream` that returns having emitted nothing — so the sentence is now true of every arm without needing the "or has already ended" qualifier:

> Every path out of this function below ends the turn with exactly one terminal event and hands it these warnings, so there is no arm on which the failure reaches only the log.

The session-id-save arm you noted as an unmentioned fourth is covered by the same change: it takes its warnings through the shared `fail` helper, and it is one of the five rows in `TestQueryAttachesAFailedHistoryWriteToWhicheverTerminalEventEndsTheTurn`.

---

docs/behavior.md
tomo-chan: Right — the sentence was mine and it was wrong about `busy`. `handlePrompt` returns straight after writing it, so a rejected prompt gets neither `error` nor `done`, and this paragraph is exactly where a non-browser client author would read the contract from.

Took the first of your two options, scoping the claim to a query that actually started, since `busy` is terminal for a *prompt* rather than for a query — calling it a third terminal frame would blur the distinction that makes the rest of the paragraph true:

> **A query that starts ends with exactly one of `error` and `done`, and `warnings` is how either one reports a non-fatal failure.** […] (A prompt answered with `busy` never becomes a query: it was rejected before a subprocess ran, so it receives neither of the two — `busy` is itself the last frame for that prompt.)

The same scoping went into agent-board.md's one-line version, which had inherited the error. The dashboard's independent `done = true` in `applyFrame`'s `busy` arm is unchanged; it is now what the docs describe rather than what quietly rescues them.

---

PR #231, opened 2026-09-11
refactor: add seams for the atomic-write and temp-file failure paths

### LINKED ISSUE #222: refactor: add seams for the atomic-write and temp-file failure paths
Split out of #195, which reduced the never-executed block backlog to the point where what remains is no longer ordinary test-writing. This issue collects the part that needs an implementation change.

## The shape

Three files write a file the same way — create a temp file, write, close, chmod, rename — and each has four or five error arms after `os.CreateTemp` has already succeeded:

| file | arms |
|---|---|
| `internal/board/atomic_write.go` | `CreateTemp`, `Write`, `Close`, `Chmod` |
| `internal/commandcenter/atomic_write.go` | `CreateTemp`, `Write`, `Close`, `Chmod` |
| `internal/commandcenter/mcp_config.go` | `CreateTemp`, `Write`, `Close`, `Chmod` |
| `internal/commandcenter/history_store.go` | `Stat` and `ReadAt` in `fileEndsWithoutTrailingNewline`, plus its caller |

**Fifteen blocks, and no test can reach any of them**, because the file being written is one the function created itself moments earlier. `#195` covered everything around them — the `MkdirAll` arm before, the `Rename` arm after — using fixtures that break the shape of the path (a regular file where a directory is expected, a directory where a file is expected, `/dev/full`). None of those work here: the temp file always exists, is always writable, and is always owned by the process. The suite also runs as root in CI, so permission bits are not available either.

`history_store.go`'s is the same problem in a different dress: reaching `ReadAt`'s failure needs an already-open file whose `Stat` reports a size its `ReadAt` cannot serve.

## Why it is worth doing

These are the arms that fire when a disk fills or a filesystem goes read-only *partway through a write* — the exact situation the temp-file-plus-rename dance exists to survive. Right now nothing checks that any of them reports the step that failed, or that the temp file is cleaned up on the way out.

Two of them also differ from each other in a way worth pinning. `atomic_write.go` closes the temp file before returning on a `Write` failure but relies on a `defer` for removal; `mcp_config.go` calls its `cleanup()` explicitly on each arm. Whether those are equivalent under every failure ordering is not currently established by anything.

## Suggested shape

The smallest change that unlocks all of them is to take the file operations as an injectable set rather than calling `os` directly — one small interface or a struct of function values, defaulting to the real ones:

```go
type fileOps struct {
	CreateTemp func(dir, pattern string) (*os.File, error)
	Chmod      func(name string, mode os.FileMode) error
	Rename     func(oldpath, newpath string) error
}
```

`Write` and `Close` come from the returned `*os.File`, so covering those two additionally needs the temp file itself behind an interface (`io.WriteCloser` plus `Name()`), which is a slightly larger change — worth deciding whether those two arms earn it.

Note that the three `atomic_write`-shaped functions are near-duplicates already; `internal/commandcenter/atomic_write.go`'s own doc comment records that it is deliberately duplicated from `internal/board`'s rather than shared. If a seam is being added anyway, whether these three collapse into one is worth revisiting at the same time — but that is a bigger question and should not block the seam.

## Constraint worth knowing

A seam is an implementation change, so a branch adding one falls under `make efficacy`: the tests it brings must fail when the implementation diff is reverted. That is satisfiable here — without the seam the tests do not compile, let alone pass — but it means this cannot ride along in a test-only branch, which is why #195 left it alone rather than doing it in passing.

Related: #195, #212 (the `os.UserHomeDir` seam, same category).

tomo-chan:
Closes #222.

## What this is

Three files wrote a file the same way — create a temp file, write, close, chmod, rename — and each had four or five error arms after `os.CreateTemp` had already succeeded. **Fifteen blocks, and no test could reach any of them**: the file being written is one the function created itself moments earlier, so it always exists, is always writable, and is always owned by the process; and the suite runs as root in CI, so permission bits are not available either.

Those are the arms that fire when a disk fills or a filesystem goes read-only *partway through* a write — the exact situation the temp-file-plus-rename dance exists to survive.

## The seam: `internal/fileops`

```go
type Ops struct {
	CreateTemp func(dir, pattern string) (File, error)
	OpenFile   func(name string, flag int, perm os.FileMode) (File, error)
	Chmod      func(name string, mode os.FileMode) error
	Rename     func(oldpath, newpath string) error
}
```

- `AtomicWrite(path, data, mode, label)` is the whole dance — create, write, **fsync**, close, chmod, rename; `CreateTemp` / `OpenFile` / `Chmod` are the individual steps, for the two callers that need the steps rather than the dance (the MCP config file, which is never renamed, and the history file, which is opened `O_APPEND`).
- `File` is `io.WriteCloser` + `Name()` + `Sync()` + `Stat()` + `ReadAt()`, so the temp file itself is behind an interface too. The issue asked whether the `Write`/`Close` arms earn that: they do — they are the disk-fills-mid-write case, which is the whole reason the temp file exists.
- `SetOpsForTest(t, ops)` substitutes it and fills every unset field with the real one. It declares its own two-method `TestingT` rather than importing `testing`, same as `internal/homedir`, and carries the same non-`t.Parallel` caveat.
- `Spy` performs each real operation on a real file and substitutes only the reported error — an injected write failure is a **short** write, as a real ENOSPC is, so a test asserting the temp file was cleaned up is asserting about a file with real partial content in it. `spy.Files()` names them; `spy.Closes()` counts closes.

### Atomic, and durable up to a point

Two claims, not one, and the doc comment says which is which. The fsync before the rename is what keeps the rename from reaching the journal as metadata while the data blocks are still in page cache — a power loss between the two commits would otherwise leave the file present at its *final* path and zero-length, worse than the old contents. The parent directory is deliberately **not** fsynced (a crash right after the rename may leave the previous contents in place, which is safe) because that would cost a directory fsync on a path the relay takes every poll.

### Rename replaces what is at the path, rather than following it

The one thing a caller migrating off `os.WriteFile` has to check, and where review caught two regressions in this branch's own work. A symlink at the target is swapped for a regular file; a bind-mounted single file cannot be replaced at all. `config.yaml` symlinked into a dotfiles repo is an ordinary setup that `os.WriteFile` wrote through, so `internal/config` resolves its two paths with `resolveWriteTarget` first — including the **dangling** case, where the link is in place but the repo has no `config.yaml` in it yet, since `O_CREATE` through a dangling link created the target too.

That resolution is deliberately at those two callers and *not* in the seam: they are the files that followed links before, while the relay cursor, bootstrap state and session id have always been written by rename — teaching the seam to follow a link would newly let one planted at those paths redirect a write.

### Shared package, not a variable per caller

Following the decision recorded in the issue comment: a package-private `var` would not have been enough. The root package's own tests drive `internal/board`'s writes through `persistBoardCursors` / `persistBootstrapState`, and a package-private variable is invisible from there. `TestPersistHelpers_FailureMidWrite_LogAndGiveUp` in `board_branches_test.go` is exactly that case.

### Every persisted file, including the two that were not in scope

The three near-duplicate `atomicWriteFile` copies collapse into it — `internal/commandcenter`'s own doc comment recorded that it was duplicated *because `internal/board`'s copy was unexported*, a reason a shared package removes.

Review found that "one seam" was still untrue as written, so `internal/config`'s two writes came along: `config.yaml` (which holds `server.auth_token` and the whole workspace layout, and was a plain `os.WriteFile` that truncated it in place on a partial failure) and the auth token file. That also retired `internal/config`'s own `chmodConfigFile` variable, a package-private override of one of the operations `Ops` now owns.

Nothing enforces this seam the way `forbidigo` enforces `internal/homedir`, and `DEVELOPMENT.md` says so rather than implying otherwise — no rule can tell a legitimate `os.WriteFile` from one that should have been an `AtomicWrite`.

## What is covered now, none of it reachable before

- **`AtomicWrite`**: the `CreateTemp`, `Write`, `Sync`, `Close`, `Chmod` and `Rename` arms, each naming the step that failed, each leaving neither a target nor a temp file behind, and each closing the temp file exactly once.
  - That last assertion is the issue's open question about whether the `defer`-based cleanup and the explicit `Close` on the write arm are equivalent under every failure ordering. They are, and `wantCloses` is per-arm, so a double close or a leaked descriptor fails the test.
- **`BuildMCPConfig`**: its four arms, where a file left behind would have held the board bearer token in plaintext. Its explicit `cleanup()`-per-arm convention is pinned against the temp path the `Spy` recorded.
- **`AppendHistory`**: `fileEndsWithoutTrailingNewline`'s `Stat` and `ReadAt` arms *and their caller*, distinguished by their inner wrap; plus the complement (an empty file must not have its last byte read at all, which keeps the `ReadAt` case from passing for the wrong reason); plus its **close** error, which it used to discard.
- **`Config.write` and `persistAuthToken`**: a mid-write failure leaves the previous `config.yaml` byte-identical and no temp file beside it, and leaves no partial token file where the next start would read one. A symlinked target is written through rather than replaced, whether or not it exists yet.
- **The `Spy` itself**, since every other test here trusts it: with nothing injected it is the real operations end to end, each knob fails its own step and no other, a real failure travels back unchanged, and its write failure really puts bytes on disk.

Fixtures that already work were left alone: `MkdirAll` and `Rename` keep their real-filesystem fixtures (a regular file where a directory goes, a directory where a file goes), because breaking the shape of a path the *caller* supplied needs no seam.

## Review findings

Twelve threads, all resolved, plus one raised in a review summary rather than on a diff line.

| Finding | Resolution |
|---|---|
| Two `//nolint:gosec` in shipped code (×2 threads) | **Removed.** Neither was needed — lint is at 0 issues without them, so gosec was never flagging the call. They were added preemptively while moving code; `history_store.go`'s `os.OpenFile` carried none on `main`. |
| Real ops returned a typed-nil `File`, the `Spy` a true nil | **Fixed.** Both openers go through one `asFile` helper — the shape `Spy.wrap` already had. Regression test asserts `f == nil` directly, since `assert.Nil` unwraps the interface and would have held for the bad value. |
| `WriteErr` wrote nothing, so no test ever saw a partial file | **Fixed.** Now a short write, with a test reading the partial bytes back off disk. |
| No `Sync` anywhere, while the docs claimed crash-safety | **Fixed.** `Sync()` on `File`, called before the close, brokered by `Ops`, with a `SyncErr` knob and a covered arm. Docs now separate atomicity from durability and state what is still not promised. |
| `AppendHistory` discarded its close error, defeating #228's warning | **Fixed.** Named return, close error reported — and a second test that it never overwrites an earlier failure. |
| "One seam" was untrue: `config.yaml`, the token file, `chmodConfigFile` | **Fixed** by moving them, not by softening the claim. |
| `%w` of a possibly-nil error + missing `t.Helper()` in `spy_test.go` | **Fixed.** Guarded wrap (a bare return trips `wrapcheck` on `io.Writer`), and the knob moved into the shared `onATempFile` helper. |
| `wrapcheck` waiver wider than its stated reason | **Narrowed** to `AtomicWrite` by signature. Verified by perturbation: an unwrapped `fileops.Chmod` in a caller is now flagged, `AtomicWrite` beside it is not. |
| Rename replaces a symlinked `config.yaml` instead of writing through it — a regression from the fix above | **Fixed** in e907736 at the two callers that previously followed links, with the reason for that placement in the code. Reproduced first. |
| …and the same for a **dangling** link — a dotfiles setup mid-flight, which `O_CREATE` used to create through | **Fixed** in 72ee5bb, reading the link with `Readlink` rather than inferring from `EvalSymlinks`' error. Reproduced first. |
| `board_test.go:466` still named `board.atomicWriteFile`, which this branch deleted — the recorded justification for the `unwritableHome` fixture, ungreppable | **Fixed** in acd92c0. Swept the branch: `grep -rn atomicWriteFile` now returns 0, and no other reference to a symbol this branch deleted remains. |
| Two fixture rows compared a resolved path against an unresolved `t.TempDir()` — green on ubuntu CI, red on any machine whose `TMPDIR` has a symlinked ancestor (macOS: `/var` → `/private/var`) | **Fixed** in 35df550. Reproduced on Linux by pointing `TMPDIR` at a symlinked path, which failed exactly those two rows; `TMPDIR=<symlinked> go test ./...` is now green across the module, so nothing else carries the shape. |

## Test plan

- [x] `make check` — passes (exit 0)
- [x] `make efficacy` — **all 34 changed tests went red** with the implementation reverted
- [x] `COVERAGE_BLOCKS_BASE=origin/main make coverage-blocks` — nothing to report; every measured block on a changed line is executed
- [x] `MUTATION_BASE=origin/main make mutation` — no surviving mutants on changed lines
- [x] `TMPDIR=<path with a symlinked ancestor> go test ./...` — green for the whole module, which is the check CI's ubuntu-only runners cannot make
- [x] Go coverage (gated packages): 96.6%; `internal/fileops` added to `COVERAGE_PKGS` and the `coverage-go` recipe, so `TestCoverageScopeCoversEveryPackage` stays green
- [x] No `//coverage:exempt`, `//mutation:exempt` or `//efficacy:exempt` markers added, and no exemption label needed

Two notes where a gate said something worth acting on rather than waving through:

- `coverage-blocks` flagged one block in the first cut of `resolveWriteTarget` — the `Readlink` error arm behind an `Lstat` guard, genuinely unreachable there. The guard turned out to be redundant (`Readlink` reports `EINVAL` for a non-link and `ENOENT` for a missing path, which is already the answer), so it was **deleted rather than exempted**: simpler and fully covered.
- `mutation` reports two `CONDITIONALS_NEGATION` mutants on `history_store.go:58` as **not analysed** — gremlins skipped them, so it knows nothing either way. Applied both by hand instead: `cerr == nil && err == nil` and `cerr != nil && err != nil` each fail `go test ./internal/commandcenter/`. Both killed; the tool simply did not run them.

Per the issue comment's warning, the repository-wide block count is not offered as the result — the denominator moves too. The diff-scoped gate above is the claim.

## Docs

- `DEVELOPMENT.md` — the seam, when *not* to reach for it (a path whose shape can be broken is still the better fixture), the rename-vs-symlink check for a caller migrating off `os.WriteFile`, and that nothing enforces it.
- `docs/architecture.md` — new `internal/fileops` section next to `internal/homedir`, including the atomicity-vs-durability distinction and the symlink rules.
- `docs/security.md` — `BuildMCPConfig` now names the seam's operations, and the test pinning that a failed build never leaves the bearer token in the temp directory.
- `docs/agent-board.md` — the `internal/board/atomic_write.go` reference points at the shared package.
- `docs/scenarios.md` — H18b (the disk fills during a save), H18c (`config.yaml` symlinked into a dotfiles repo) and H18d (the same link before the repo has a `config.yaml` in it). No `scenarios-exempt` label needed.

https://claude.ai/code/session_01ENV1ZHpqwbf1vmPABTwq8o


internal/fileops/fileops.go
```diff
@@ -0,0 +1,160 @@
...
+// ops is the seam itself. SetOpsForTest is the only supported way to replace
+// it, and it restores the previous value when the test ends.
+var ops = realOps()
+
+func realOps() Ops {
+	return Ops{
+		CreateTemp: func(dir, pattern string) (File, error) { return os.CreateTemp(dir, pattern) },
+		OpenFile: func(name string, flag int, perm os.FileMode) (File, error) {
+			return os.OpenFile(name, flag, perm) //nolint:gosec // the caller owns the path; see each call site
+		},
```
tomo-chan: **The real ops return a typed-nil `File` on the error path; the `Spy` returns a true nil. The double and the thing it doubles disagree.**

`os.CreateTemp` / `os.OpenFile` return `(*os.File)(nil), err` on failure. Returning that pair directly through a `(File, error)` signature boxes the nil pointer into a non-nil interface. `Spy.wrap` (spy.go:85) does the opposite — it checks `err` first and returns an untyped `nil`.

Verified against this branch:

```go
f, err := realOps().CreateTemp("/nonexistent-dir-xyz", "p-*")
// err != nil, and f == nil is **false**
f.Name()  // panic: invalid memory address or nil pointer dereference
```

(`(*os.File).Name()` is `return f.name`, so it derefs nil.)

Failure scenario: a caller written in the shape this package invites — `f, err := fileops.CreateTemp(dir, p); if f != nil { defer os.Remove(f.Name()) }` — takes the branch and panics in production, while every test driving it through a `Spy` skips the branch and passes. That is precisely the class of divergence the `Spy` exists to rule out, and no test here covers the real ops' error path (`TestOpsDefaultToTheRealOperations` only exercises success; `TestSetOpsForTestFillsUnsetOperationsWithTheRealOnes` substitutes its own closure returning a proper `nil`).

Cheap to close, and it makes the seam match its double:

```suggestion
		CreateTemp: func(dir, pattern string) (File, error) {
			f, err := os.CreateTemp(dir, pattern)
			if err != nil {
				return nil, err
			}
			return f, nil
		},
		OpenFile: func(name string, flag int, perm os.FileMode) (File, error) {
			f, err := os.OpenFile(name, flag, perm) //nolint:gosec // the caller owns the path; see each call site
			if err != nil {
				return nil, err
			}
			return f, nil
		},
```


---

internal/fileops/spy.go
```diff
@@ -0,0 +1,133 @@
...
+	*os.File
+	spy *Spy
+}
+
+func (f *spyFile) Write(p []byte) (int, error) {
+	if f.spy.WriteErr != nil {
+		return 0, f.spy.WriteErr
+	}
+	return f.File.Write(p)
+}
```
tomo-chan, line 116: **`WriteErr` returns before writing anything, so the `Spy` can never produce the partially-written file this whole PR is about.**

The doc comment three lines up (spy.go:95-97) says "The real `*os.File` is always operated on first, so the filesystem ends up in the state it would have been in anyway, and only the reported error is substituted." For `Write` that is not what happens — `n=0`, nothing reaches the file. (Same for `Stat`/`ReadAt`, where it matters less, and for `CreateTemp`/`OpenFile`, where no file is created at all despite `Files()` being the mechanism tests assert cleanup with.)

The consequence is not cosmetic, because the PR description states the case being covered as "the disk fills or a filesystem goes read-only *partway through* a write". A real ENOSPC returns `n > 0, err` — a temp file containing a prefix of the payload. What the suite actually exercises is "a write that never started", which is the easy half:

- `internal/commandcenter/mcp_config_test.go`'s `write` arm is annotated "a file holding the bearer token never outlives the failure" — but under `WriteErr` no token bytes were ever written, so the arm cannot demonstrate that. It would pass identically if `BuildMCPConfig` wrote the token and then failed to clean up a file it had already truncated to empty.
- `board_branches_test.go`'s `assert.Empty(entries, "no partial or temp file may be left...")` likewise never sees a partial file.

Making the double faithful costs little and turns those assertions into the ones they claim to be — a short write, which is exactly what `io.Writer` says an ENOSPC looks like:

```suggestion
func (f *spyFile) Write(p []byte) (int, error) {
	if f.spy.WriteErr != nil {
		// A real ENOSPC is a short write, not a write that never
		// happened: n > 0 bytes land on disk and then the error is
		// reported. Reproduce that, so a test asserting the temp file
		// was cleaned up is asserting about a file with real,
		// partial content in it.
		n, _ := f.File.Write(p[:len(p)/2])
		return n, f.spy.WriteErr
	}
	return f.File.Write(p)
}
```


---

internal/fileops/fileops.go
```diff
@@ -0,0 +1,160 @@
...
+
+// ops is the seam itself. SetOpsForTest is the only supported way to replace
+// it, and it restores the previous value when the test ends.
+var ops = realOps()
+
+func realOps() Ops {
+	return Ops{
+		CreateTemp: func(dir, pattern string) (File, error) { return os.CreateTemp(dir, pattern) },
+		OpenFile: func(name string, flag int, perm os.FileMode) (File, error) {
+			return os.OpenFile(name, flag, perm) //nolint:gosec // the caller owns the path; see each call site
```
tomo-chan: **A blanket `//nolint:gosec` at the one place every file open now funnels through, with a justification no tool can follow.**

`docs/security.md`'s gosec policy says: *"Fix `gosec` findings structurally in the implementation rather than suppressing them in shipped code"* and *"Production paths should avoid suppressions and make the safety argument explicit in code structure instead."*

Two things about this particular one:

1. On `main`, the call this replaces — `os.OpenFile(path, os.O_CREATE|os.O_RDWR|os.O_APPEND, historyFileMode)` in `internal/commandcenter/history_store.go:46` — carried **no** suppression. So this is a new one, not a relocated one.
2. The reason given, "the caller owns the path; see each call site", is not checkable. G304's whole point is to flag a variable path, and the variable is now one hop away in a different package. Today there is exactly one caller (`AppendHistory`). The moment a second one lands — and the package doc invites more, since `OpenFile` is exported as a general seam — it inherits the suppression silently, with no finding for a reviewer to weigh. That is the failure mode: a future caller passing a path derived from request data gets no G304 signal at all, at the one choke point where every open in the repo will eventually pass.

Narrowing the suppression to the rule (`//nolint:gosec // G304: ...`) is the minimum; better still would be to leave `OpenFile` unsuppressed and keep the per-call-site `//nolint:gosec // G304` where the safety argument about *that* path can actually be read.

---

DEVELOPMENT.md
```diff
@@ -102,6 +102,24 @@ happens inside `internal/config`, `internal/server`'s integration tests drive co
 at once, and the root package's tests drive `internal/commandcenter`'s default paths — none of which
 a package-private variable can reach. See issue [#212](https://github.com/tomo-chan/panemux/issues/212).
 
+**Persisted files are written through one seam, and it is `internal/fileops`.** `fileops.AtomicWrite`
```
tomo-chan, line 105: **"Persisted files are written through one seam" is not true as written, and the file it most obviously excludes is the operator's `config.yaml`.**

`docs/architecture.md:168` makes the same claim in the stronger form: *"The write discipline every persisted file shares."*

Two persisted files on `main` do not go through it, and this PR does not change them:

- `internal/config/config.go:522` — `Config.Save` writes `~/.config/panemux/config.yaml` with a plain `os.WriteFile`. Non-atomic. This is the file holding `server.auth_token` and the entire workspace/pane layout, so a crash or a full disk during a save from the dashboard's settings dialog truncates it, and the next start reads a half-written YAML. That is a strictly worse outcome than the relay-cursor case `AtomicWrite` was extracted for.
- `internal/config/agent_board.go:102` — `persistAuthToken` writes `~/.config/panemux/token` the same way.

There is also a *second* seam left standing: `internal/config/config.go:29`'s `var chmodConfigFile = os.Chmod`, a package-private override of exactly one of the four operations `fileops.Ops` now owns. The `internal/homedir` precedent this section leans on ("the home directory has **one** seam") got its force from `.golangci.yml`'s `forbidigo` rule making the claim enforceable; nothing makes this one enforceable, and it is already false on the day it is written.

I'd either bring `internal/config`'s two writes onto `fileops.AtomicWrite` and delete `chmodConfigFile` (which is also what would let `Config.Save`'s own failure arms be driven from a test — the stated purpose of the seam), or soften both doc claims to name the files actually covered. The one thing not to leave is a rule in DEVELOPMENT.md that the next reader will check and find untrue.

---

internal/fileops/spy_test.go
```diff
@@ -0,0 +1,200 @@
...
+// needs the temp file itself behind an interface, not just the os functions.
+func spyFileKnobs() []spyKnob {
+	return []spyKnob{
+		{
+			name: "WriteErr",
+			spy:  func() *fileops.Spy { return &fileops.Spy{WriteErr: errInjected} },
+			exercise: func(t *testing.T, dir string) error {
+				return onATempFile(t, dir, func(f fileops.File) error {
+					_, err := f.Write([]byte("x"))
+					return fmt.Errorf("writing: %w", err)
```
tomo-chan: **Unconditional `%w` of a possibly-nil error: this arm can never report "the Spy stopped failing Write" in a legible way.**

`fmt.Errorf("writing: %w", nil)` returns a **non-nil** error whose message is `writing: %!w(<nil>)`. So if `Spy.WriteErr` ever stopped taking effect, `TestSpyFailsOnlyTheStepItWasGiven/WriteErr` would not report "expected an error, got nil" — it would report an `ErrorIs` mismatch against a `%!w(<nil>)` formatting artifact, which reads like a wrapping bug rather than the double having silently gone inert.

Given that this file's own preamble is *"a double that quietly stubbed a step it was not asked about would make every test using it assert against a fiction"*, this is the one case worth failing clearly. The sibling knobs (`StatErr`, `ReadAtErr`) already return `err` bare; this one is the odd one out and the wrap buys nothing.

```suggestion
				return onATempFile(t, dir, func(f fileops.File) error {
					_, err := f.Write([]byte("x"))
					return err
				})
```

(Minor, same hunk: this `exercise` is the only one of the eight missing the `t.Helper()` its siblings call.)

---

tomo-chan (commented):
Reviewed the full diff. The refactor itself is behaviour-preserving — I diffed `fileops.AtomicWrite` against both deleted copies line by line and they are identical, and `go build ./... && go test ./internal/fileops/... ./internal/commandcenter/... ./internal/board/...` is green on this branch. No production behaviour change found, so everything below is about the seam's fidelity and the claims made for it.

Five findings, inline. The two I'd act on before merge:

1. **`fileops.go:56-59`** — the real ops return a typed-nil `File` on the error path while `Spy.wrap` returns a true nil, so the double and the thing it doubles disagree on exactly the value a caller would nil-check. Reproduced: `f == nil` is `false` after a failed `CreateTemp`, and `f.Name()` panics.
2. **`spy.go:103`** — `WriteErr` returns before writing anything, so the `Spy` cannot produce a partially-written file. The PR's stated target ("the disk fills *partway through* a write") is therefore simulated as "a write that never started", and the `mcp_config_test.go` write arm's "a file holding the bearer token never outlives the failure" claim is not demonstrated by it.

The other three: a new blanket `//nolint:gosec` at the single choke point every future file open will pass through (`fileops.go:58`); the "every persisted file goes through one seam" claim in `DEVELOPMENT.md`/`architecture.md`, which `internal/config`'s non-atomic `config.yaml` and auth-token writes and its surviving `chmodConfigFile` var already contradict; and an unconditional `%w` of a nil error in `spy_test.go:104`.

---

internal/fileops/fileops.go
```diff
@@ -0,0 +1,160 @@
...
+		_ = tmp.Close()
+		return fmt.Errorf("writing temp %s: %w", label, err)
+	}
+	if err := tmp.Close(); err != nil {
+		return fmt.Errorf("closing temp %s: %w", label, err)
+	}
+	if err := ops.Chmod(tmpPath, mode); err != nil {
+		return fmt.Errorf("setting %s mode: %w", label, err)
+	}
+	if err := ops.Rename(tmpPath, path); err != nil {
```
tomo-chan, line 180: **No `Sync` anywhere, yet the doc comment — and now `docs/architecture.md` — assert crash-safety repo-wide.**

`grep -n Sync internal/fileops/*.go` returns nothing on this branch. The sequence is `Write` → `Close` → `Chmod` → `Rename`, with no `tmp.Sync()` before the close and no fsync of the parent directory after the rename.

That is close-then-rename, which gives **atomicity** (a reader never sees a half-written file) but not **durability across a crash**. The two claims are different, and lines 100–101 make the second one:

> so a crash or power loss mid-write can never leave a truncated or half-written file on disk

`docs/architecture.md:168-171` repeats it in the stronger, repo-wide form ("The write discipline every persisted file shares … a crash or power loss mid-write can never leave a truncated file behind").

Failure scenario, not hypothetical on the filesystems panemux runs on: the rename is journaled as metadata while the temp file's data blocks are still in page cache. On ext4 `data=writeback`, XFS, or btrfs, a power loss between the two commits leaves `board-relay-cursor.json` **present at its final path and zero-length** — which is worse than the old file, because `LoadCursorFile` now reads an empty file at the canonical path rather than the previous good one. (ext4's default `data=ordered` has a heuristic that usually saves the rename-over case; that is a mount-option-dependent courtesy, not something the doc comment can claim.)

The relevant point for *this* PR: `Ops` already brokers every other step, so this is both fixable and testable in the shape the PR is built for — add `Sync() error` to `File`, call it before the close, and `Spy` gets a `SyncErr` knob and a new covered arm for free. Without it the seam mediates four operations and omits the one that makes the discipline it is named for actually hold.

If syncing is deliberately out of scope (it does cost a real fsync per relay-cursor write), that is a defensible call — but then the doc comment here and the architecture.md sentence should say "atomic", not "can never leave a truncated or half-written file on disk", since a reader will take the stronger sentence at face value the next time they choose between this seam and a plain `os.WriteFile`.

---

internal/commandcenter/history_store.go
```diff
@@ -43,7 +44,7 @@ func AppendHistory(path string, entries []HistoryEntry) error {
 	if err := os.MkdirAll(filepath.Dir(path), 0750); err != nil {
 		return fmt.Errorf("creating command center history directory: %w", err)
 	}
-	f, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR|os.O_APPEND, historyFileMode)
+	f, err := fileops.OpenFile(path, os.O_CREATE|os.O_RDWR|os.O_APPEND, historyFileMode)
```
tomo-chan, line 47: **This PR makes the `Close` arm injectable and then leaves it the one arm still discarded — which silently defeats the warning path added in #228, the commit this branch sits on.**

Four lines below the changed line, unchanged by this PR:

```go
defer f.Close() //nolint:errcheck
```

On a buffered write the error does not necessarily surface at `f.Write` — with ext4 delayed allocation an ENOSPC/EDQUOT is commonly reported at close, not at write. So `AppendHistory` returns `nil`, `runner.go:349` collects no warning, and the done frame tells the operator the history was persisted when nothing reached the disk. `runner.go:343-346`'s own comment states the intent this breaks: *"there is no arm on which the failure reaches only the log."*

What makes it worth raising **on this PR specifically** rather than as pre-existing debt: before this change the arm was untestable, which was a real argument for leaving it. Now it isn't — `Spy.CloseErr` exists and is already exercised in `fileops_test.go:97` and `mcp_config_test.go:124`; `grep -rn CloseErr internal/commandcenter/` shows `history_store_test.go` is the one place it is missing. So the branch adds the seam, covers the neighbouring `Stat` and `ReadAt` arms it reached through it, and skips the arm with the user-visible consequence.

The shape the rest of the function already uses:

```go
	defer func() {
		if cerr := f.Close(); cerr != nil && err == nil {
			err = fmt.Errorf("closing command center history file: %w", cerr)
		}
	}()
```

(named return `err`), plus a `CloseErr` row in `history_store_test.go`'s table asserting `AppendHistory` returns non-nil — which is what makes it reach `runner.go`'s `warnings`.

---

.golangci.yml
```diff
@@ -65,6 +65,13 @@ linters:
       checks: ["all"]
     whitespace:
       multi-if: true
+    wrapcheck:
+      # internal/fileops' own errors already carry the caller's label — that is
+      # what its `label` parameter is for, so every message names the file that
+      # failed. Wrapping them again at each call site would only prepend a
+      # second name for the same file.
+      ignore-package-globs:
+        - panemux/internal/fileops
```
tomo-chan: **The stated reason holds for `AtomicWrite` only, but the glob waives the whole package — including the three functions that carry no label at all.**

The comment's argument is exact for `AtomicWrite`: its errors already read `writing temp relay cursor file: …`, so a second wrap at the call site would name the same file twice. That is what the `label` parameter is for.

`CreateTemp`, `OpenFile` and `Chmod` (fileops.go:85-97) have no `label` parameter. They are one-line pass-throughs that return the raw `*fs.PathError` from `os`:

```go
func Chmod(name string, mode os.FileMode) error {
	return ops.Chmod(name, mode)
}
```

So a future `return fileops.Chmod(p, 0600)` in a caller escapes unwrapped **and unflagged** — the one case `wrapcheck` is enabled to catch, at a package the docs invite every new persisted-file write to route through. Both current step callers do wrap today (`mcp_config.go`, `history_store.go`), so this costs nothing right now; it is the next caller that pays.

Narrowing it to the function keeps the reason and the rule aligned:

```suggestion
    wrapcheck:
      # internal/fileops.AtomicWrite's errors already carry the caller's label —
      # that is what its `label` parameter is for, so every message names the
      # file that failed. Wrapping them again at each call site would only
      # prepend a second name for the same file. The step operations
      # (CreateTemp / OpenFile / Chmod) carry no label and are NOT waived: they
      # return the raw *fs.PathError, so their callers must still wrap.
      ignore-sigs:
        - panemux/internal/fileops.AtomicWrite(
```

(If `ignore-sigs` proves awkward against this signature, `ignore-package-globs` with the comment corrected to say it waives the label-less pass-throughs too is the honest fallback — the thing to avoid is a waiver whose written rationale does not cover what it actually waives.)

---

tomo-chan (commented):
Second pass over this branch, at the same head (`1990a79`). Verification run before writing any of it: `go vet ./...` clean, `golangci-lint run` reports 0 issues on the three affected packages, and `go test -race -count=1` passes for `internal/fileops`, `internal/board`, `internal/commandcenter` and the root package. The consolidation itself is behaviour-preserving — `fileops.AtomicWrite` is line-for-line identical to both deleted copies, and the package doc comment from the removed `internal/commandcenter/atomic_write.go` is correctly preserved in `doc.go`.

Three new findings are inline above. They do not repeat the five already posted on this PR.

**One more that has no diff line to anchor to**, since the file is untouched by this branch:

- `board_test.go:466` — `// helpers go through board.atomicWriteFile, whose first statement is os.MkdirAll` … This PR deletes `board.atomicWriteFile`, and `grep -rn atomicWriteFile` over the whole tree now returns **this comment and nothing else**. It is not a cosmetic stale reference: the comment is the recorded argument for why the `unwritableHome` fixture has to plant a regular *file* where the directory goes (an absent directory is simply created by that first `MkdirAll`, so the earlier version of those tests drove the happy path while claiming to cover the failure branch). The premise is still true of `fileops.AtomicWrite` — fileops.go:109-111 is the same `os.MkdirAll` — so the fix is one identifier: `board.atomicWriteFile` → `fileops.AtomicWrite`. Left as is, the next reader who greps the name finds nothing and has no way to check the fixture's justification.

Ordering, if it helps: the `Close`-arm finding in `history_store.go` is the one with a user-visible consequence today (a done frame reporting history as persisted when it was not), and it is the cheapest to close now that `Spy.CloseErr` exists. The `Sync` question in `AtomicWrite` is the one worth an explicit decision rather than a quick patch — either the operation joins the seam, or the two doc claims come down to "atomic". The `wrapcheck` glob and the stale comment are small.

---

internal/fileops/fileops.go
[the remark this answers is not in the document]
tomo-chan: Both suppressions are gone, and neither was ever needed: with them removed `make lint-go` reports 0 issues, so gosec was not flagging either call. `os.OpenFile` in `internal/commandcenter/history_store.go` carried none on `main` either — I added the annotation preemptively when moving the call, which is exactly the "centralized waiver" problem you describe.

Nothing was restructured to achieve that, because nothing had to be. The `Spy` stays in the production package for the reason its doc comment gives (its users are in other packages — `internal/board`'s callers are driven from the root package's tests), but it no longer carries a suppression either.

Fixed in 8ce5451.

---

internal/fileops/fileops.go
tomo-chan: Confirmed and fixed. This was a real divergence, and the one a double must not have.

Rather than repeating the `if err != nil` in both openers, both now go through one helper, which is literally the shape `Spy.wrap` already had:

```go
func asFile(f *os.File, err error) (File, error) {
	if err != nil {
		return nil, err
	}
	return f, nil
}
```

`TestRealOpsReportATrueNilFileWhenTheyFail` pins it. One thing worth recording about that test, since it nearly went in tautological: `assert.Nil` is **not** enough — testify unwraps the interface and reports a typed-nil pointer as nil, so it would have held for exactly the value the test exists to reject. It asserts `created == nil` directly instead, with the reason in a comment.

Fixed in 8ce5451.

---

internal/fileops/spy.go
tomo-chan: Right, and this was the more important of the two divergences — the suite was exercising the easy half of the case the whole branch is about. Taken as suggested.

The doc comment above it is corrected too, since it was the thing asserting a property the code did not have:

> Where a real failure would still have changed the file, the real `*os.File` is operated on first and only the reported error is substituted — see `Write`, which is a short write rather than a write that never happened.

And the claim is now checked rather than written down. `TestSpyWriteErrIsAShortWriteNotAWriteThatNeverHappened` writes `"abcdef"`, asserts `n == 3` and `errInjected`, then reads the file back and asserts `"abc"` is genuinely on disk — so the partial content the `mcp_config_test.go` and `board_branches_test.go` assertions are about really exists before it is cleaned up.

Fixed in 8ce5451.

---

internal/fileops/fileops.go
tomo-chan: Removed, not narrowed — and point 1 is the thing that settles it. It was a new suppression, not a relocated one, and with it deleted `make lint-go` is at 0 issues, so gosec was never flagging the call in the first place. I had added it defensively while moving the code, which is the worst reason to add one.

Point 2 is the part I'd have gotten wrong even with a narrowed `// G304:` rule, so it is worth agreeing with explicitly: the argument "the caller owns the path" is not checkable at a choke point, and it silently extends to every future caller. There is now nothing there to extend.

Fixed in 8ce5451.

---

DEVELOPMENT.md
tomo-chan: Took the first option: the two writes moved onto the seam rather than the sentence being softened.

- `Config.write` → `fileops.AtomicWrite(c.filePath, data, configFileMode, "config")`. Its own `os.MkdirAll` went with it, since `AtomicWrite`'s message for that arm with this label is character-for-character the one that was there (`creating config directory: …`).
- `persistAuthToken` → the same, labelled `"auth token file"`.
- `chmodConfigFile` deleted; `tightenConfigFilePermissions` calls `fileops.Chmod`, and `config_test.go`'s override becomes `fileops.SetOpsForTest(t, (&fileops.Spy{ChmodErr: …}).Ops())`.

The failure arms that opened up are covered rather than just migrated: `TestWrite_FailureMidWrite_LeavesThePreviousConfigIntact` saves a real config, injects a mid-write failure, and asserts the previous `config.yaml` is byte-identical afterwards and no temp file is beside it — the regression that was previously only arguable. `TestEnsureAuthTokenReportsAFailureMidWrite` is its counterpart.

Two follow-on notes:

- `TestWrite_PathIsADirectory_ReportsAWriteFailure` now asserts `replacing config` rather than `writing config`: with a rename at the end, a directory at the target fails there, not at the write. Same for the auth token test, which also had a `NotContains` assertion that had quietly become vacuous — it asserts the step name positively now.
- `docs/scenarios.md` gains H18b, since `internal/config` changed and the disk-fills-during-save case is a use case rather than an implementation detail. No `scenarios-exempt` label needed.

On enforceability, you are right that this seam has no `forbidigo` equivalent, and I do not think one exists — nothing can tell a legitimate `os.WriteFile` from one that should have been an `AtomicWrite`. `DEVELOPMENT.md` now says so rather than leaning on the `homedir` precedent's authority:

> Unlike the home-directory seam, nothing enforces this one … It is true as written today because the writes were moved onto it, not because the build would fail otherwise, so a new persisted file is the reader's job to route correctly.

Fixed in 8ce5451.

---

internal/fileops/spy_test.go
tomo-chan: Both fixed. The `%!w(<nil>)` artifact is a fair catch — in the one file whose job is to prove the double has not gone inert, the failure message would have pointed at the wrap instead.

Not quite the suggested form, because a bare `return err` here trips `wrapcheck` (`io.Writer.Write` is a stdlib interface method, which is why this knob was the odd one out to begin with — its siblings return `fileops.File` methods, which wrapcheck does not treat the same way). Guarding the wrap gets both properties:

```go
if _, err := f.Write([]byte("x")); err != nil {
	return fmt.Errorf("writing: %w", err)
}
return nil
```

Now a Spy that stopped failing `Write` reports "expected an error, got nil", which is the legible failure you wanted.

The missing `t.Helper()` is gone too — that `exercise` moved into the shared `onATempFile` helper, which calls it, along with the `Stat`/`ReadAt` knobs and a new `Sync` one.

Fixed in 8ce5451.

---

internal/fileops/fileops.go
tomo-chan: You are right on the distinction, and I took the fix rather than the wording change — the failure you describe is the one this function exists to prevent, and leaving it while claiming otherwise would have been the worst of the two options.

`Sync() error` is on `File`, `AtomicWrite` calls `tmp.Sync()` before the close (closing the temp file on that arm, like the write arm does), and `Spy` gets a `SyncErr` knob. The arm is covered by the existing table — it is now six rows rather than five, asserting `syncing temp test file`, exactly one close, and no temp file left behind — plus a `SyncErr` row in `spy_test.go`'s knob table.

The doc comment now separates the two claims explicitly, and says what is still *not* promised rather than leaving it to be inferred:

> **Atomic, and durable up to a point — the two are not the same claim.** … The parent directory is *not* fsynced, so a crash right after the rename may leave the previous contents in place; that is safe, and buying the stronger guarantee would cost a directory fsync on a path the relay takes every poll.

`docs/architecture.md` carries the same two paragraphs, including your ext4 `data=ordered` point as a mount option rather than a guarantee.

One thing I added while I was in there, since it is the same class of "a reader will take this at face value": rename gives the target a new inode, so a path that is a bind-mounted single file cannot be replaced this way. That is now recorded in both places — it is the one behavior change `config.yaml` moving onto this function could surprise someone with.

Fixed in 8ce5451.

---

internal/commandcenter/history_store.go
tomo-chan: Agreed, including on why it belongs on this PR: the "it is untestable" argument for leaving it died the moment `Spy.CloseErr` existed, and I covered the two arms next to it while skipping the one with the user-visible consequence.

Taken as suggested — named return, close error captured only when nothing failed earlier. The `//nolint:errcheck` is gone with it.

Two tests rather than one, because "reports the close error" and "does not overwrite an earlier failure" are separate decisions and the second is the one that quietly goes wrong:

- `TestAppendHistoryReportsAFailedClose` — `CloseErr` alone, asserts non-nil and `closing command center history file`, which is what makes it reach `finishAfterStream`'s `warnings`.
- `TestAppendHistoryKeepsTheEarlierFailureWhenTheCloseAlsoFails` — `WriteErr` *and* `CloseErr`, asserts the write failure survives and the close message does not appear. Without the `&& err == nil` guard the cause would be replaced by its consequence.

One incidental fix the named return forced: `go vet`'s `shadow` then flagged `if err := os.MkdirAll(...)`, which is now `mkdirErr`.

Fixed in 8ce5451.

---

.golangci.yml
tomo-chan: Narrowed by signature, as suggested — `extra-ignore-sigs` rather than `ignore-sigs`, so golangci-lint's own defaults stay in place:

```yaml
    wrapcheck:
      extra-ignore-sigs:
        - panemux/internal/fileops.AtomicWrite(
```

with the comment now saying both halves, including that the step operations are deliberately *not* waived.

I checked it rather than assuming it held, since a signature match that silently waived nothing (or everything) would be worse than the glob. Adding an unwrapped `return fileops.Chmod(path, cursorFileMode)` to `internal/board/cursor_store.go`:

```
internal/board/cursor_store.go:56:10: error returned from external package is unwrapped:
  sig: func panemux/internal/fileops.Chmod(name string, mode os.FileMode) error (wrapcheck)
1 issues
```

— flagged, while `AtomicWrite` on the line below it passes. That is the property you were asking for, so the next caller pays nothing and gets the signal.

One consequence worth flagging, because it is a widened exclusion elsewhere in the same file: dropping the package glob surfaced `wrapcheck` findings in *test* files, where a helper handing an error to `assert.ErrorIs` must return it exactly as it got it. `wrapcheck` is now in the existing `_test\.go` exclusion rule alongside `errcheck`/`goconst`, with that reason written next to it.

Fixed in 8ce5451.

---

internal/config/config.go
```diff
@@ -519,10 +514,12 @@ func (c *Config) write() error {
...
-	if err := os.WriteFile(c.filePath, data, configFileMode); err != nil {
-		return fmt.Errorf("writing config: %w", err)
-	}
-	return nil
+	// Through the shared seam rather than os.WriteFile: this file holds
+	// server.auth_token and the whole workspace layout, and a direct write
+	// that fails partway leaves it truncated for the next start to read. It
+	// also creates the parent directory ("creating config directory: …") and
+	// sets the mode, which is why neither is done here.
+	return fileops.AtomicWrite(c.filePath, data, configFileMode, "config")
```
tomo-chan: **New in `8ce5451`: moving `config.yaml` onto rename-based writing silently breaks a symlinked config, and this is the one caller where that setup is likely.**

The doc comment on `AtomicWrite` now records the inode consequence — *"a path that is a bind-mounted single file cannot be replaced this way"* — but names the rare case and not the common one. A symlink is hit by the same mechanism: `rename(2)` replaces the **link itself**, not what it points at.

Reproduced on this branch with exactly the sequence `AtomicWrite` performs (temp beside the target, then `os.Rename` onto it):

```
still a symlink: false
dotfiles copy now contains: "original\n"
```

Failure scenario, and the reason it is worth a line rather than a shrug: `ln -s ~/dotfiles/panemux/config.yaml ~/.config/panemux/config.yaml` is an ordinary way to keep this file in a dotfiles repo, and until this commit `os.WriteFile` **followed** the link and wrote through it. Now the first save from the dashboard — adding a pane, renaming a workspace, `EnsureAuthToken` on a first run — replaces the symlink with a regular file. No error, nothing in the log. The dotfiles copy freezes at its old contents, and the user finds out when a later `git status` there is clean and their edits stop taking effect. `internal/config/agent_board.go:99`'s token file has the same shape, though it is a less likely thing to symlink.

This is not an argument against the change — `config.yaml` is exactly the file that most needed atomic replacement, and the PR makes that case well. It is an argument that the one caller whose path an operator hand-manages should resolve the link first:

```suggestion
	// Through the shared seam rather than os.WriteFile: this file holds
	// server.auth_token and the whole workspace layout, and a direct write
	// that fails partway leaves it truncated for the next start to read. It
	// also creates the parent directory ("creating config directory: …") and
	// sets the mode, which is why neither is done here.
	//
	// Resolved first, because AtomicWrite renames onto the path it is given
	// and rename replaces a symlink rather than following it. A config.yaml
	// symlinked into a dotfiles repo — which os.WriteFile used to write
	// through — would otherwise be silently replaced by a regular file on the
	// first save, detaching it from the repo with no error anywhere.
	target := c.filePath
	if resolved, resolveErr := filepath.EvalSymlinks(target); resolveErr == nil {
		target = resolved
	}
	return fileops.AtomicWrite(target, data, configFileMode, "config")
```

(`EvalSymlinks` failing is the ordinary first-run case — the file does not exist yet — so it falls back to the configured path rather than erroring. A test would be the one I ran: symlink `config.yaml` at a target in a second directory, `SaveWorkspaces()`, then assert `os.Lstat(link).Mode()&os.ModeSymlink != 0` and that the target picked up the new contents.)

---

tomo-chan (commented):
Re-checked all nine of my findings against `8ce5451`. **Eight are closed, one is not**, and the fix commit introduces one new thing worth a look (inline above).

Verification run on `8ce5451`: `go build ./...` and `go vet ./...` clean, `golangci-lint run ./...` reports **0 issues** with `gosec` still enabled, and `go test -count=1 ./...` passes for the whole module (`-race` too, on `internal/fileops`, `internal/config`, `internal/commandcenter`, `internal/board` and the root package). All four test names the new `docs/scenarios.md` H18b row cites resolve to real functions.

### Closed

| # | Finding | How |
|---|---|---|
| 1 | Typed-nil `File` from the real ops | `asFile` converts explicitly; `TestRealOpsReportATrueNilFileWhenTheyFail` pins it — and correctly uses `== nil` rather than `assert.Nil`, which would have held for exactly the value the test rejects |
| 2 | `WriteErr` wrote nothing | Now a short write (`p[:len(p)/2]`), with `TestSpyWriteErrIsAShortWriteNotAWriteThatNeverHappened` asserting the partial payload is really on disk |
| 3 | Blanket `//nolint:gosec` | Both suppressions deleted outright, not narrowed. Lint is green with `gosec` on, so the waiver was never load-bearing |
| 4 | "one seam" was false | Fixed by making it true rather than by softening it: `config.yaml` and the auth token file both moved onto `AtomicWrite`, and `chmodConfigFile` — the second seam — is gone. The added paragraph noting that *nothing enforces this one*, unlike the `forbidigo`-backed homedir seam, is the right caveat |
| 5 | Unconditional `%w` of a nil error | Guarded. The `t.Helper()` half is moot — `onATempFile` calls it, so the failure line is already attributed |
| 6 | No `Sync` | `File.Sync()`, a `SyncErr` knob, a `sync` arm in the table, and the doc comment split into what atomicity buys vs. what the fsync buys. Documenting that the parent directory is deliberately *not* fsynced, with the cost reason, is better than what I asked for |
| 7 | `AppendHistory` discarded the close error | Named return plus a deferred close that reports `cerr` only when `err == nil`; `store_branches_test.go` drives it through `Spy.CloseErr` |
| 8 | `wrapcheck` package-wide waiver | Now `extra-ignore-sigs: panemux/internal/fileops.AtomicWrite(` — and `extra-` rather than plain `ignore-sigs` is the better pick, since it adds to the defaults instead of replacing them |

One thing in the same commit that is a widening rather than a fix, flagged only so it is a decision and not a side effect: `wrapcheck` was also added to the `_test\.go` exclusion. The rationale given (a helper handing an error to `assert.ErrorIs` must return it verbatim) is sound, but it is a broader waiver than finding 5 needed.

### Still open

- **`board_test.go:466`** — `// helpers go through board.atomicWriteFile, whose first statement is os.MkdirAll`. `grep -rn atomicWriteFile` over the tree still returns this line and nothing else. The premise holds for `fileops.AtomicWrite` (same `os.MkdirAll` at fileops.go:147), so it is still a one-identifier change, and it is the recorded justification for why `unwritableHome` has to plant a regular file where the directory goes.

---

internal/config/config.go
tomo-chan: Confirmed and fixed. I reproduced it before touching anything, with the exact sequence `AtomicWrite` performs (temp beside the target, `Sync`, `Close`, `Rename`) against a `config.yaml` symlinked into a second directory:

```
still a symlink: false
dotfiles copy now contains: "original\n"
```

Identical to your run. And you are right that the doc comment named the rare case and missed the common one — I wrote "bind-mounted single file" thinking about inodes and did not follow the same mechanism to the setup people actually have.

Taken as suggested, with one change of shape: the resolution is a named helper rather than inline, because the token file needs it too.

```go
func resolveWriteTarget(path string) string {
	resolved, err := filepath.EvalSymlinks(path)
	if err != nil {
		return path
	}
	return resolved
}
```

`internal/config/agent_board.go`'s `persistAuthToken` calls it as well. "Less likely to be symlinked" is true, but it is the same regression, and it is worse on a first run: `EnsureAuthToken` is the very first save on a fresh install, so a symlinked token file would be detached before the operator ever opened the dashboard.

**Why it is at the callers and not in `AtomicWrite`** — this was the one real decision here, so it is in the helper's doc comment rather than only in this thread. These two files are the ones that *were* written with `os.WriteFile` and so did follow a link; the seam's other files (relay cursor, bootstrap state, command-center session id) have always been written by rename and have never followed one. Putting `EvalSymlinks` in the seam would fix these two and simultaneously make a link planted at any of those other paths able to redirect a write — a new exposure in exchange for a restored behavior. So: restore it exactly where it existed.

Tests are the ones you described. `TestWrite_SymlinkedConfig_WritesThroughTheLinkInsteadOfReplacingIt` asserts `os.Lstat(link).Mode()&os.ModeSymlink != 0`, that the dotfiles-side target picked up the new contents, and that no temp file is left in the directory holding the link; `TestEnsureAuthToken_SymlinkedTokenFile_WritesThroughTheLink` is its counterpart. Both were red first, for the right reasons (`Should not be zero, but was ----------`).

`EvalSymlinks` failing falls back to the configured path, as you noted — first run, file not there yet.

The rule is recorded in three places, since "rename replaces rather than follows" is what any future caller migrating off `os.WriteFile` has to check: `AtomicWrite`'s doc comment, `docs/architecture.md`, and `DEVELOPMENT.md`'s seam paragraph. `docs/scenarios.md` gains H18c for the dotfiles setup itself.

Gates after the fix: `make check` exit 0, lint 0 issues, `make efficacy` all **32** changed tests red, diff-scoped `coverage-blocks` clean, `mutation` no survivors.

Fixed in e907736.

---

internal/config/config.go
```diff
@@ -519,10 +514,40 @@ func (c *Config) write() error {
...
+// A path that cannot be resolved is returned unchanged: the ordinary case is a
+// first run, where the file does not exist yet.
+func resolveWriteTarget(path string) string {
+	resolved, err := filepath.EvalSymlinks(path)
+	if err != nil {
+		return path
 	}
-	return nil
+	return resolved
 }
```
tomo-chan, line 573: **The `err != nil` fall-through is right for a first run and wrong for a dangling link — and a dangling link is a dotfiles setup mid-flight, not an exotic state.**

`filepath.EvalSymlinks` fails when *any* component of the chain is missing, which covers two different situations that want opposite answers:

1. **No file at `path` at all** — first run. Returning `path` unchanged is correct.
2. **`path` IS a symlink, but its target does not exist yet** — you linked `~/.config/panemux/config.yaml` into a dotfiles repo that has no `config.yaml` in it yet, expecting the first save to populate it. `EvalSymlinks` fails here too, so the link is handed to `AtomicWrite` and replaced, which is exactly what this function was added to prevent.

Reproduced on this branch, running `resolveWriteTarget` as written against a link whose target is absent:

```
resolveWriteTarget returns the link unchanged: true
os.WriteFile: link still a symlink=true, target created=true
AtomicWrite:  link still a symlink=false, target created=false
```

The `os.WriteFile` row is the behavior this file had before the seam: `O_CREATE` through a dangling symlink **creates the target**. So case 2 is the same regression as the one just fixed, in the same scenario, one step earlier in the user's setup.

Separating the two cases needs `Lstat` rather than the error:

```suggestion
func resolveWriteTarget(path string) string {
	if resolved, err := filepath.EvalSymlinks(path); err == nil {
		return resolved
	}
	// EvalSymlinks fails for two different reasons and only one of them means
	// "leave it alone": no file here yet (a first run), or a link whose target
	// does not exist yet (linked into a dotfiles repo that has no config.yaml
	// in it, expecting this save to create one — which os.WriteFile did, since
	// O_CREATE through a dangling link creates the target).
	info, err := os.Lstat(path)
	if err != nil || info.Mode()&os.ModeSymlink == 0 {
		return path
	}
	dest, err := os.Readlink(path)
	if err != nil {
		return path
	}
	if !filepath.IsAbs(dest) {
		dest = filepath.Join(filepath.Dir(path), dest)
	}
	return dest
}
```

(One hop, not a chain — a dangling link to a dangling link is past the point of guessing. `AtomicWrite`'s own `MkdirAll` then creates the repo directory if it is missing, which is the same thing `os.WriteFile` could not do and is the harmless direction to differ in.)

The existing test extends by one line: build the same fixture as `TestWrite_SymlinkedConfig_WritesThroughTheLinkInsteadOfReplacingIt` but skip the `os.WriteFile(target, …)` that pre-creates it, then assert the link survived and the target now exists.

Worth saying explicitly since the doc comment above reasons carefully about it: this does **not** widen the redirect surface the comment declines to open. It applies to the same two paths for the same reason, and a link planted at one of them already redirects the write today whenever its target happens to exist — which is the strictly easier thing for an attacker to arrange.

---

tomo-chan (commented):
Re-checked `e907736`. The symlink finding is addressed, and addressed better than I proposed — putting `resolveWriteTarget` at the two callers rather than inside `AtomicWrite`, with the reason written down (the other files on the seam have always been written by rename, so teaching the seam to follow a link would newly let one planted at those paths redirect a write) is the right call and one I had not made. `docs/architecture.md`, `DEVELOPMENT.md` and the `AtomicWrite` doc comment all say the same thing consistently, and H18c in `docs/scenarios.md` names two tests that exist and pass.

Verification on `e907736`: `go build ./...` and `go vet ./...` clean, `golangci-lint run ./...` **0 issues**, `go test -race -count=1 ./internal/config/... ./internal/fileops/...` green.

One gap left inside that fix, inline above: `EvalSymlinks` returning an error covers both "no file yet" and "a link whose target does not exist yet", and only the first wants the path left alone. The second is the same dotfiles scenario one step earlier in the setup, and `os.WriteFile` used to create the target through the dangling link — reproduced on this branch.

### Still open from the previous round

- **`board_test.go:466`** — `// helpers go through board.atomicWriteFile, whose first statement is os.MkdirAll`. `grep -rn atomicWriteFile` over the tree still returns this one line. It is the recorded justification for why `unwritableHome` plants a regular file where the directory goes, and the premise still holds for `fileops.AtomicWrite` (`os.MkdirAll` at fileops.go:155), so it is a one-identifier change. Second time flagging it, so if it is being left deliberately that is fine — just say so and I will stop raising it.

---

internal/config/config.go
tomo-chan: Right on both counts, and I reproduced it before touching anything — same three lines you got:

```
resolveWriteTarget returns the link unchanged: true
os.WriteFile: link still a symlink=true, target created=true
AtomicWrite:  link still a symlink=false, target created=false
```

The middle row is the part that settles it: `O_CREATE` through a dangling link creates the target, so this was the same regression as the one I had just fixed, one step earlier in the operator's setup. My fall-through read "cannot resolve → leave it alone" as one case when it is two.

Taken, with **one simplification to the suggested shape**: `Readlink` alone, no `Lstat` before it.

```go
func resolveWriteTarget(path string) string {
	if resolved, err := filepath.EvalSymlinks(path); err == nil {
		return resolved
	}
	dest, err := os.Readlink(path)
	if err != nil {
		return path
	}
	if !filepath.IsAbs(dest) {
		dest = filepath.Join(filepath.Dir(path), dest)
	}
	return dest
}
```

`Readlink` reports `EINVAL` for a path that is not a link and `ENOENT` for one that is not there, so its error is already the "leave it alone" answer for both — the `Lstat` adds nothing the `Readlink` error does not already say. What made me look: with the `Lstat` guard in place, the diff-scoped `coverage-blocks` gate reported `config.go:566-568` (the `Readlink` error arm) as never executed, and it is genuinely unreachable behind that guard — only a link deleted between the two calls could enter it. Dropping the guard is simpler *and* fully covered, which beats an exemption.

Tests are the ones you described, plus the two cases that would otherwise have gone in unverified. `writeTargetCases()` is a table over `resolveWriteTarget` directly:

| case | must resolve to |
|---|---|
| no file here at all — a first run | the path itself |
| an ordinary file | itself |
| a link whose target exists | the target |
| a dangling link, absolute target | the target |
| a dangling link, target relative to the link's own directory | the target, joined against the link's dir |

That last row is the one carrying its weight beyond the report: without it `!filepath.IsAbs(dest)` resolving against the *working* directory would have passed everything else. `TestWrite_SymlinkedConfigWithAMissingTarget_CreatesItThroughTheLink` is the end-to-end half — link survives, target created. Both red first.

One hop only, as you said. And agreed that this widens no redirect surface, for the reason you give: a link planted at either path already redirects the write whenever its target exists, which is the easier thing to arrange — so the helper's doc comment says why the resolution is at these two callers, not that it is a boundary.

`docs/scenarios.md` gains H18d for the mid-flight dotfiles setup; `docs/architecture.md` records why `EvalSymlinks`' error is not the signal.

Gates: `make check` exit 0, lint 0 issues, `make efficacy` all **34** changed tests red, diff-scoped `coverage-blocks` clean, `mutation` no survivors.

Fixed in 72ee5bb.

---

tomo-chan:
Re [#pullrequestreview-5179627787](https://github.com/tomo-chan/panemux/pull/231#pullrequestreview-5179627787) — the `board_test.go:466` item. Not deliberate, and thank you for flagging it a second time rather than dropping it.

**Why I missed it twice**, since that is the more useful half: it is a finding in a review's *summary body*, not a thread on a diff line, and I was working from the review-thread list. Every threaded finding got picked up; this one was invisible to how I was reading the PR. Noted, and I checked the summary bodies of all your earlier reviews before writing this — that one item is the only thing in them I had not already closed.

Fixed in acd92c0: `board.atomicWriteFile` → `fileops.AtomicWrite`, one identifier, premise unchanged (`os.MkdirAll` is still `AtomicWrite`'s first statement).

Your reasoning for why it is not cosmetic is the reason I did not just delete the sentence: it is the recorded argument for why `unwritableHome` has to plant a regular *file* where the directory goes, and a reader who greps the name to check that justification would have found nothing.

I also swept the branch for the same shape rather than fixing only the line you named — a symbol this branch deleted, still referenced somewhere. `grep -rn atomicWriteFile` over the tree now returns **0**. The one other hit worth mentioning is `docs/architecture.md`'s `chmodConfigFile`, which is deliberate: it appears in a "that also retired `internal/config`'s own `chmodConfigFile` variable" note, where the name being gone is the point.

Gates on `acd92c0`: `make check` exit 0, `make lint-go` 0 issues, `make efficacy` all 34 changed tests red, diff-scoped `coverage-blocks` clean, `mutation` no survivors.

---

internal/config/config_branches_test.go
```diff
@@ -131,6 +166,145 @@ func TestWrite_NoFilePath_IsANoOp(t *testing.T) {
...
+		{
+			name: "a link whose target exists",
+			build: func(t *testing.T) (string, string) {
+				t.Helper()
+				target := filepath.Join(t.TempDir(), "config.yaml")
+				require.NoError(t, os.WriteFile(target, []byte("server:\n"), 0600))
+				link := filepath.Join(t.TempDir(), "config.yaml")
+				require.NoError(t, os.Symlink(target, link))
+				return link, target
+			},
```
tomo-chan, line 255: **Two of these five rows compare against an unresolved `t.TempDir()`, which makes them pass on the CI runner and fail on a developer's Mac.**

The two rows that take the `EvalSymlinks` success path — *"an ordinary file is already its own target"* (`return path, path`) and *"a link whose target exists"* (`return link, target`) — assert that the resolved string equals a path built from `t.TempDir()`. That holds only when no ancestor of the temp directory is itself a symlink, because `EvalSymlinks` resolves the **whole** chain, not just the last component.

On macOS `TMPDIR` sits under `/var/folders/…`, and `/var` is a symlink to `/private/var`, so every `t.TempDir()` has a symlinked ancestor. I do not have a Mac here, so I am not claiming to have run it — what I did reproduce on this branch is the mechanism, with a symlinked ancestor standing in for `/var`:

```
path     = /tmp/anc2098367721/var/T/TestX/config.yaml
resolved = /tmp/anc2098367721/private-real/T/TestX/config.yaml
equal    = false
```

The other three rows are unaffected: the first-run and dangling-link rows never reach `EvalSymlinks`' success path, and `Readlink` returns the stored target verbatim rather than resolving anything.

Every workflow in `.github/workflows/` is `runs-on: ubuntu-latest`, so CI will not catch this — a contributor on macOS gets two red rows out of `make check` and no hint from the failure that it is about their `TMPDIR` rather than about `resolveWriteTarget`. Given that this repository ships darwin builds and `main.go` has a darwin branch in `browserOpenArgv`, that is a real audience.

Normalizing the fixture once keeps every row asserting what it means to assert:

```suggestion
		{
			name: "an ordinary file is already its own target",
			build: func(t *testing.T) (string, string) {
				t.Helper()
				path := filepath.Join(resolvedTempDir(t), "config.yaml")
				require.NoError(t, os.WriteFile(path, []byte("server:\n"), 0600))
				return path, path
			},
		},
		{
			name: "a link whose target exists",
			build: func(t *testing.T) (string, string) {
				t.Helper()
				target := filepath.Join(resolvedTempDir(t), "config.yaml")
				require.NoError(t, os.WriteFile(target, []byte("server:\n"), 0600))
				link := filepath.Join(t.TempDir(), "config.yaml")
				require.NoError(t, os.Symlink(target, link))
				return link, target
			},
		},
```

with, next to the table:

```go
// resolvedTempDir is t.TempDir() with any symlinked ancestor already followed.
// On macOS TMPDIR lives under /var, which is a symlink to /private/var, so a
// row asserting that EvalSymlinks returns the path it was given would compare
// two spellings of the same file. Only the rows that reach EvalSymlinks'
// success path need it; Readlink returns the stored target verbatim.
func resolvedTempDir(t *testing.T) string {
	t.Helper()
	resolved, err := filepath.EvalSymlinks(t.TempDir())
	require.NoError(t, err)
	return resolved
}
```

(The `link` side deliberately stays an unresolved `t.TempDir()` — the link's own spelling is an input here, not something the assertion is about.)

---

tomo-chan (commented):
Re-checked `acd92c0`. **Both of my open items are closed**, and there is nothing left from the original review.

- **Dangling symlink** (`72ee5bb`) — closed, and the implementation is better than the one I suggested: dropping the `Lstat` and letting `Readlink`'s own error (EINVAL for a non-link, ENOENT for nothing there) be the "leave it alone" answer removes an arm that only a link deleted between the two calls could enter. The relative-target join is handled, the "one hop, not a chain" boundary is stated, and `TestResolveWriteTargetFollowsALinkWhoseTargetDoesNotExistYet` covers both reasons `EvalSymlinks`' error cannot separate plus both link shapes — a table where I had only asked for one case.
- **`board_test.go:466`** (`acd92c0`) — closed. `grep -rn atomicWriteFile` over the tree now returns nothing.

Verification on `acd92c0`: `go build ./...` and `go vet ./...` clean, `golangci-lint run ./...` **0 issues**, `go test -count=1 ./...` green for the whole module, and `-race` green on `internal/config`, `internal/fileops` and the root package. H18d's two test names resolve.

One new thing, inline above, and it is about the test fixture rather than the fix: two of the five table rows compare a resolved path against an unresolved `t.TempDir()`, which passes on `ubuntu-latest` and fails wherever `TMPDIR` has a symlinked ancestor — macOS being the standard case, since `/var` is a symlink to `/private/var`. Every workflow here is ubuntu-only, so CI will not surface it. I could not run it on a Mac; what I reproduced on this branch is the mechanism, with a symlinked ancestor standing in for `/var`.

---

internal/config/config_branches_test.go
tomo-chan: Correct, and taken as suggested — `resolvedTempDir` next to the table, applied to exactly those two rows, with the link side left as an unresolved `t.TempDir()` for the reason you give.

I have no Mac either, but this one does not need one: **`t.TempDir()` honours `TMPDIR`**, so pointing it at a path with a symlinked ancestor reproduces it on Linux directly, against the real test rather than a standalone program.

```
$ root=$(mktemp -d); mkdir -p "$root/private-real"; ln -s "$root/private-real" "$root/var"
$ TMPDIR="$root/var" go test ./internal/config/ -run ResolveWriteTarget

--- FAIL: TestResolveWriteTargetFollowsALinkWhoseTargetDoesNotExistYet
    --- FAIL: .../an_ordinary_file_is_already_its_own_target
    --- FAIL: .../a_link_whose_target_exists
```

Two rows, the two you named — which is also the evidence for your claim about the other three, rather than my re-deriving it: the first-run and dangling-link rows never reach `EvalSymlinks`' success path, and `Readlink` hands back the stored target verbatim. After the fix the same command is green.

Then I turned the instrument on the rest of the branch instead of only the line you flagged, since "compares a resolved path against an unresolved one" is a shape rather than an incident:

```
$ TMPDIR="$root/var" go test ./...
(every package ok)
```

So nothing else in the module carries it today. That command is in the helper's doc comment, so the next person who writes a path-resolving test has the reproduction rather than the conclusion.

Ubuntu-only CI is the part worth keeping in view — it would not have caught this one and will not catch the next, and adding a macOS job is a bigger change than this PR should make. I have not opened an issue for it; say the word if you would like one.

Gates on 35df550: `make check` exit 0, lint 0 issues, `make efficacy` all 34 changed tests red, diff-scoped `coverage-blocks` clean, `mutation` no survivors.

Fixed in 35df550.

---

tomo-chan (commented):
Re-checked `35df550`. **Closed, and the commit message's own claim holds — I verified it rather than taking it.**

The comment on `resolvedTempDir` says the failure is reproducible on Linux by pointing `TMPDIR` at a path with a symlinked ancestor, and that exactly the two rows fail. Run against `acd92c0` with `TMPDIR` set to a symlink:

```
--- PASS: …/no_file_here_at_all_—_a_first_run
--- FAIL: …/an_ordinary_file_is_already_its_own_target
--- FAIL: …/a_link_whose_target_exists
--- PASS: …/a_dangling_link,_absolute_target
--- PASS: …/a_dangling_link,_target_relative_to_the_link's_own_directory
```

Two failures, and only those two — so the claim is accurate down to which rows, and it also means this is now a *checkable* claim rather than an assertion about a platform nobody here can run. Same command on `35df550` passes all five, and the whole `internal/config` package is green under that `TMPDIR` too. Keeping the `link` side unresolved, with the reason inline, is the right half to leave alone.

Verification on `35df550`: `go build ./...` and `go vet ./...` clean, `golangci-lint run ./...` **0 issues**, `go test -count=1 ./...` green for the whole module, `-race` green on `internal/config` and `internal/fileops`.

**Nothing of mine is open on this PR.** All nine findings from the two review passes are closed, as are the two raised against the fix commits (the symlink replacement and its dangling-link gap) and this fixture one. For the record, the only item I raised and did not pursue was the `wrapcheck` waiver on `_test\.go` added in `8ce5451` — flagged as a widening rather than a defect, with a rationale I agreed with, and I am not asking for a change there.

---

Issue #227, opened 2026-09-10
test: the home-directory seam does not make substitution safe under t.Parallel, which was #212's stated reason for it


tomo-chan:
Raised in the review on #224. The seam landed; this is the half of #212's justification it does not deliver, recorded so it is a decision rather than a surprise.

> **Resolved as option 1 in #234** — see the [Resolution](#resolution-decided-in-234) section at the end. The text below is the original filing and is left as written.

## What #212 said

> `t.Setenv` mutates process-wide state, so any package using it can never mark those tests `t.Parallel()` — and the constraint is invisible: nothing fails, the tests simply cannot be parallelized later without a subtle, intermittent breakage.

That is accurate about `t.Setenv`. It is **not** fixed by what replaced it.

## What actually shipped

`internal/homedir`'s seam is one unsynchronized package variable:

```go
var dirFn = os.UserHomeDir
```

`SetForTest` writes it and registers a `t.Cleanup` that writes it back; `Dir()` reads it. Two tests running in parallel therefore race on it, and the failure mode is **worse than what it replaced**:

| | two parallel tests substituting it |
|---|---|
| `t.Setenv` | panics — `testing` refuses it outright |
| `homedir.SetForTest` | races; the loser silently reads the other test's home directory |

`t.Setenv`'s refusal is a feature. The seam has the same sharing problem with no guard.

Nothing is broken today: `grep -rn "t.Parallel()" --include="*_test.go" .` returns **0** across the repository. The risk is that the seam *reads* like the thing that unblocks parallelism, so the first person to try it gets an intermittent, order-dependent failure rather than a panic pointing at the cause.

`internal/session`'s `userCacheDirFn` (#226) has the identical shape, so whatever is decided here should apply to both.

## Interim mitigation, already in place

#224 stopped the texts from claiming it. `SetForTest`'s doc comment and DEVELOPMENT.md's testability rule now separate what the seam buys (process environment untouched, works on Windows where `os.UserHomeDir` reads `USERPROFILE`, failures injected directly instead of conjured from an empty `$HOME`) from what it does not, and say outright that `t.Setenv` is better in this one respect. **This issue is not about the docs being wrong — they are correct now.** It is about whether the underlying limitation is worth removing.

## Options, with the objection to each

1. **Leave it, and keep the warning.** Costs nothing. The repo has no parallel tests and no measured need for them; `make test-go` is already seconds. The warning is one `grep` away from anyone about to add `t.Parallel`.
[...]

Issue #226, opened 2026-09-10
refactor: give os.UserCacheDir the same single seam os.UserHomeDir now has


tomo-chan:
Raised in the review on #224 and scoped out of it as the follow-up that actually closes the gap that PR could only paper over.

## The state

`internal/session/browseropen.go:85` has the seam:

```go
var userCacheDirFn = os.UserCacheDir
```

That is the *per-package variable* shape #224 replaced for the home directory, and it has the same limitation: it is package-private, so only `internal/session`'s own tests can substitute it. `internal/session/browseropen_test.go` does, twice, with the hand-rolled save/restore #224 removed everywhere else.

`internal/server`'s two integration-env helpers need it and cannot reach it, so they still mutate the process environment:

| file | line |
|---|---|
| `internal/server/api_integration_test.go` | `t.Setenv("HOME", home)`, `t.Setenv("XDG_CACHE_HOME", ...)` |
| `internal/server/ws_integration_test.go` | the same two |

These are the **only** `t.Setenv("HOME", ...)` calls left in the repository after #224. Every other surviving `t.Setenv` sets `PATH` or `TMPDIR`, which is a different variable and out of scope here.

## Why both variables, and why the seam does not reach them

`os.UserCacheDir` never consults `os.UserHomeDir`. It reads the environment itself, and differently per platform:

- **non-darwin Unix**: `XDG_CACHE_HOME`, else `$HOME/.cache`
- **darwin**: `$HOME/Library/Caches`, and `XDG_CACHE_HOME` is ignored entirely

So `homedir.SetForTest` cannot influence it by construction, and neither variable covers it alone. #224's first revision dropped the `$HOME` line, which on macOS would have made a local pane install the browser shim (`session.installLocalBrowserShim`) into the developer's **real** `~/Library/Caches/panemux/bin` — the three real files the comment above `newAPIEnv` was written to prevent. CI is Linux-only, so it would not have gone red; it would only have bitten whoever ran the suite on a Mac. Both lines were restored, with the reason recorded inline.

## Suggested shape

Follow #224: one package, `internal/cachedir`, with `Dir()` plus `SetForTest` / `SetFailingForTest` taking the same two-method `TestingT` so nothing links `testing` into the binary. `internal/session`'s `userCacheDirFn` and its two hand-rolled overrides collapse into it, and both `internal/server` helpers drop both env vars.

[...]

Issue #218, opened 2026-09-06
fix: the bootstrap watcher re-logs a detection failure on every tick


tomo-chan:
Found while covering `bootstrap.go`'s branches for #195, confirmed by measurement.

## What happens

When `DetectInteractiveAgentType` fails for a pane, `checkPane` logs and drops the pane from `pending`:

```go
	agmsgType, detected, err := detector.DetectInteractiveAgentType()
	if err != nil {
		log.Printf("Warning: agent board bootstrap: detecting agent type for pane %q: %v", paneID, err)
		delete(b.pending, paneID)
		return
	}
```

Dropping from `pending` suppresses nothing. The next `pollOnce` finds the same live session, calls `checkPane` again, and logs again.

Measured with a fake session whose detector always errors — three ticks, three identical lines:

```
Warning: agent board bootstrap: detecting agent type for pane "pane-a": boom
Warning: agent board bootstrap: detecting agent type for pane "pane-a": boom
Warning: agent board bootstrap: detecting agent type for pane "pane-a": boom
```

## Why it matters

Detection failures are not transient by nature. A pane whose tmux server went away, or whose SSH connection is down, keeps failing for as long as it stays in that state — which can be the whole life of the process. At the watcher's poll interval that is a warning line per tick, indefinitely, for a condition the operator has already been told about.

The failure mode is not the log volume so much as what it does to the log's usefulness: a board with one broken pane drowns out everything else panemux writes.

## The shape of the fix already exists here

`bootstrap.go` has `warnOnce`, keyed per pane, and uses it for the presence warning:

```go
func (b *bootstrapWatcher) warnOnce(paneID, message string) {
	if b.presenceWarned[paneID] {
		return
	}
	b.presenceWarned[paneID] = true
	log.Printf("Warning: %s", message)
}
```

So the question is not what to build but whether this warning should share that treatment, and if so whether it wants its own map — a pane that starts detecting successfully after a transient failure should probably become eligible to warn again, which `presenceWarned` does not currently reset either.

Worth deciding deliberately rather than copying: "warn once ever" and "warn once per failing streak" are different, and the second is what a poll loop usually wants.

## Note on tests

[...]

PR #234, opened 2026-09-12
fix: four backlog items — drain ordering, warning streaks, JSON-RPC shape, cache-dir seam

### LINKED ISSUE #230: fix: streamOutput's drain runs before run()'s cancel(), so a malformed-line error frame can be delayed by the whole query timeout
Found while reviewing #228 (`9f041d8`), and deliberately kept out of that PR: the underlying ordering is older than that change, and the part the change does move is small enough not to justify widening the PR.

## The ordering

`streamOutput` (`internal/commandcenter/runner.go`) drains the rest of stdout in a `defer`:

```go
defer func() { _, _ = io.Copy(io.Discard, stdout) }()
```

A deferred call runs before the caller resumes, so the drain completes **before** `streamOutput` returns — and therefore before `run()`'s

```go
if streamErr != nil {
    cancel()
}
```

ever executes. The comment on that `cancel()` says it exists so that "a subprocess that doesn't exit on its own after a malformed line would [not] hold the busy flag (and this goroutine) for up to the full query timeout". But in exactly that case the drain in front of it blocks on `read(2)` against a pipe the subprocess is still holding open, and nothing releases it until the query-timeout context kills the process and the pipe reaches EOF. The `cancel()` cannot shorten a wait that happens before it.

So the busy flag and the goroutine are held for up to `QueryTimeout` (default 5 minutes) on that path, which is the outcome the `cancel()` was written to prevent.

## What #228 changed about it

Before `9f041d8`, `streamOutput` emitted the malformed-line `EventError` itself, from inside the function body — i.e. *before* the deferred drain. The client got its error frame immediately, and only the server-side goroutine was stuck.

`9f041d8` moved terminal-event emission into `finishAfterStream` (correctly — that is what lets warnings ride every terminal event, per #214's follow-up). The consequence is that the client-visible error frame is now behind the drain as well:

| implementation | `EventError` reaches the client |
|---|---|
| `a09e3b1` (before) | immediately |
| `9f041d8` (current) | after the pipe closes |

Measured with a fake `cmdRunner` whose stdout yields one malformed line and then blocks for 400 ms before EOF, standing in for a subprocess that stays alive with stdout open:

```
a09e3b1:  EventError ... arrived after 0s
9f041d8:  EventError ... arrived after 400ms
```

In production that 400 ms is however long it takes the pipe to close, which for a subprocess that does not exit on its own is the full query timeout. The palette shows a spinner for that whole time instead of the error.

`9f041d8`'s commit message describes the delay as "however long a killed subprocess takes to reap", which is the `cmd.Wait()` part only — it does not account for the drain in front of the `cancel()`.

## Reproduction

A throwaway test in `internal/commandcenter` reproduces it without touching the real CLI:

```go
// blockingStdout yields a malformed line, then blocks on read until release
// is closed — standing in for a subprocess that stays alive with its stdout
// pipe open after emitting garbage.
type blockingStdout struct {
	head    *strings.Reader
	release <-chan struct{}
	once    sync.Once
}

func (b *blockingStdout) Read(p []byte) (int, error) {
	if b.head.Len() > 0 {
		return b.head.Read(p)
	}
	b.once.Do(func() { <-b.release })
	return 0, io.EOF
}
func (b *blockingStdout) Close() error { return nil }
```

driven through `NewRunner(cfg).Query(...)` with `NewCommand` returning a `fakeCmd` wrapping it, timing when the `EventError` arrives. `release` is closed after 400 ms.

## Suggested fix

Move the drain so it cannot precede the cancellation — either drain in a separate goroutine, or have `run()` cancel before the drain rather than after `streamOutput` returns. Whatever shape it takes, the property worth pinning in a test is the one the existing comment already claims and cannot currently deliver: **after a malformed line, neither the busy flag nor the client's error frame waits on a subprocess that has stopped writing but not exited.**

`TestRunnerMalformedStreamJSONCancelsQueryContextImmediately` passes today because the fake stdout in that test reaches EOF immediately, so the drain returns at once — the test never reaches the case the comment is about.

## Scope note

Two separate things, worth keeping apart when fixing:

- **Pre-existing**: the busy flag / goroutine being held for up to the query timeout, because the drain sits in front of the `cancel()`. Present before #228.
- **Introduced by `9f041d8`**: the client-visible error frame being subject to the same wait. Not a correctness bug — the frame does arrive, and exactly one terminal frame per query still holds — but it is a real latency regression on that path.

tomo-chan:
Closes #230, #218, #210, #226, #227. (#222 was already closed by #231, which is in `main`.)

## #230 — the drain ran before the cancel

`streamOutput` drained the rest of stdout in a `defer`, which runs before the caller resumes and therefore before `run()`'s `cancel()`. Against a subprocess that stops writing but does not exit, that drain blocks on `read(2)` until the query timeout kills the process — so the busy flag, the goroutine, and (since terminal events moved into `finishAfterStream`) the client's error frame all waited out the full `QueryTimeout`, which is exactly what the `cancel()` existed to prevent.

The drain now lives in `run()`, after the cancellation, so it can only ever wait on a subprocess already being killed. The drain itself stays — a subprocess blocked writing into a pipe no one reads can never exit — it just no longer sits in front of the thing that unblocks it.

`TestRunnerMalformedStreamJSONCancelsQueryContextImmediately` could not catch this: its fake stdout reaches EOF at once. The new test's stdout blocks until the query context is canceled, mirroring a real pipe's lifetime, and fails at a 2s bound against the old ordering with a 1-minute timeout configured.

## #218 — one broken pane drowned out the log

A pane whose `DetectInteractiveAgentType()` failed was logged on **every** poll tick; dropping it from `pending` suppresses nothing, since the next tick finds the same live session. Detection failures are not transient by nature, so that was a warning line per tick for the life of the process.

Decided deliberately rather than copied, as the issue asked: **once per failing streak**, not once per tick and not once ever. `warnOnce` keys suppression per pane and per kind, and `clearWarning` releases it on the success side of each condition, so a failure that returns after a recovery is news again, and one kind never silences another. This also fixes `presenceWarned` never resetting.

Suppression is keyed against the pane's **session identity** too, the way `bootstrapped` and `givenUp` already are (review finding). A restarted pane is a new `Session` object, and without that the one case producing silence rather than one line would be the strongest recovery signal there is — the pane having been torn down and recreated. A pane that disappears from the manager drops its entry outright, which also keeps the map from growing for panes that no longer exist.

## #210 — responses carrying neither result nor error

`omitempty` on `Result` and `ID` produced two shapes §5 does not allow: `notifications/initialized` sent as a *request* answered with neither member, and a parse error with no `id` at all where the spec requires `null`.

Struct tags cannot express "exactly one of result/error" — `omitempty` drops a nil result, and removing it emits `"result": null` beside an error. `MarshalJSON` picks between two wire types instead.

**And `null` is not the right result either** (review finding): MCP's own schema narrows a successful response's `result` to an object, so a JSON `null` fails it exactly as the absent key did — §5 alone would have moved that response from one invalid shape to another. A nil `Result` now serializes as `{}`. I checked the other `handleLine`/`dispatch` paths in the same pass as the issue asked: every one of them carries an id and exactly one member once this is in place.

## #226 / #227 — the cache-directory seam, and the decision on parallelism

`internal/cachedir` mirrors `internal/homedir`, and `internal/server`'s two integration helpers drop the last `t.Setenv("HOME", ...)` calls in the repository.

**Two packages, not one with two functions**, decided in this PR as the issue asked. It is a second global with its own platform behavior (`os.UserCacheDir` never consults `os.UserHomeDir`, and reads `XDG_CACHE_HOME` on Unix but `$HOME/Library/Caches` on darwin), each seam's doc comment is about the variable it owns, and — the decisive one — a merged package could not keep its `forbidigo` waiver narrow.

That last argument only holds with the waivers *on the calls*: a `linters.exclusions.rules` path entry suppresses every `forbidigo` diagnostic in the directory, so the first revision of this PR let an `os.UserHomeDir` call inside `internal/cachedir` through (review finding). The four intentional calls now carry a line-level `//nolint:forbidigo` instead, confirmed by perturbation — adding such a call to `internal/cachedir` now fails lint, where it passed before.

**#227 resolves as option 1**, recorded in the three places that must continue to agree (`internal/homedir`'s doc comment, DEVELOPMENT.md's testability rule, docs/architecture.md): keep the warning, do **not** add a mutex. A lock removes the data race without making restore correct across parallel tests — the appearance of a guarantee — and per-test injection is the only real fix, at the cost of exported signature changes across six packages for a suite with zero `t.Parallel` calls.

## Notes for review

- Five `//efficacy:exempt` markers, each with its reason. Three are unmodified tests the red-check attributes a neighbouring helper, doc comment or `//nolint` annotation to; two guard the obvious wrong fix for the issue they sit under rather than the behavior before it.
- **`gofmt` moves a `//directive`-shaped comment to the end of its doc-comment block**, so "the whole reason on and after the marker line" is not reachable in a gofmt-clean file. Each block reads: prose paragraph, blank `//`, then a self-contained one-line marker — in order, with the marker last where gofmt keeps it.
- `TestCheckPaneLogsADetectionFailureAndDropsThePane` and `TestRunnerStreamOutputDrainsRemainingOutputAfterScannerError` pinned the behavior being changed, and were rewritten rather than deleted — the first is now the streak assertion its own comment said would need rewriting, the second now pins both halves (streamOutput leaves the rest unread; `run()` drains it).
- Docs updated in the same change: behavior.md (the cancel-then-drain ordering, replacing a paragraph that claimed nothing waited on the reordering), agent-board.md (the warning discipline and the wire shape), architecture.md (`internal/cachedir`, and the `t.Parallel` caveat on both seams), scenarios.md (C9, E10, E11), DEVELOPMENT.md (the coverage-gate package list, which had also been missing `internal/fileops` since #231).
- The `mutation` check failed on the first push for a reason outside this diff: gremlins gathers coverage by running the whole suite, and `internal/api`'s `TestRestartSession_ClosesPortForwards` hit its own fixture race — `freeLoopbackPort` closes a listener to learn a free port, and under a parallel `go test ./...` something else can take that port before `Ensure` binds it, which surfaces as the handler's `409`. Green on the re-run; this branch touches neither `internal/api` nor `internal/portforward`.

## Test plan

- [x] `make check`
- [x] `EFFICACY_BASE=origin/main make efficacy` — 19 changed tests, all red on revert
- [x] `COVERAGE_BLOCKS_BASE=origin/main make coverage-blocks` — nothing to report
- [x] `MUTATION_BASE=origin/main make mutation` — no surviving mutants on changed lines
- [x] `make check-scenarios` — 233 names across 98 auto rows resolve
- [x] Each fix's test confirmed failing before its implementation
- [x] The forbidigo waiver's narrowing confirmed by perturbation, not by reading

https://claude.ai/code/session_01EYZ6SqeC5yix8Srkx8dWkd


internal/boardmcp/server.go
```diff
@@ -43,14 +43,63 @@ type jsonrpcRequest struct {
...
+func (r jsonrpcResponse) MarshalJSON() ([]byte, error) {
+	id := r.ID
+	if len(id) == 0 {
+		id = json.RawMessage("null")
+	}
+	if r.Error != nil {
+		//nolint:wrapcheck // an encoding detail of this type's own marshaler
+		return json.Marshal(errorResponse{JSONRPC: r.JSONRPC, Error: r.Error, ID: id})
+	}
+	// A nil Result serializes as null, which is what a known method with
```
tomo-chan: `"result": null` satisfies JSON-RPC 2.0 §5 but is still invalid MCP, so the stated motivation ("the client here is an LLM subprocess's own MCP layer, whose strictness panemux does not control") isn't actually served by this half of the fix.

MCP's own schema defines a successful response's `result` as an object (`Result = { _meta?: ..., [key: string]: unknown }`, and `JSONRPCResponse.result: Result`). A JSON `null` fails that schema exactly as the previously-absent key did. So for `{"jsonrpc":"2.0","id":7,"method":"notifications/initialized"}` the wire shape goes from one invalid form to another.

The id half of #210 (null rather than absent on a parse error) is right, and the "exactly one of result/error" half is right — it's specifically the nil-`Result` serialization that wants an empty object rather than null:

```suggestion
	// A known method with nothing to report (notifications/initialized sent
	// as a request) answers with an *empty object*, not null: §5 requires a
	// result member, and MCP narrows that member to an object, so null would
	// be rejected by a strict MCP client exactly as an absent key was.
	result := r.Result
	if result == nil {
		result = struct{}{}
	}
	//nolint:wrapcheck // an encoding detail of this type's own marshaler
	return json.Marshal(resultResponse{JSONRPC: r.JSONRPC, Result: result, ID: id})
```

`TestInitializedSentAsARequestIsAnsweredWithANullResult` and its scenarios.md row (E11) would need the matching update.

---

bootstrap.go
```diff
@@ -355,22 +388,42 @@ func (b *bootstrapWatcher) agmsgPresent(ctx context.Context, paneID, host string
...
+	b.clearWarning(paneID, warnKindProbe)
 	return present, true
 }
 
-func (b *bootstrapWatcher) warnOnce(paneID, message string) {
-	if b.presenceWarned[paneID] {
+// warnOnce logs message unless kind is already being suppressed for paneID,
+// i.e. unless this pane is already inside a failing streak of that kind.
+// clearWarning ends the streak.
+func (b *bootstrapWatcher) warnOnce(paneID, kind, message string) {
```
tomo-chan, line 431: The suppression is keyed by pane ID alone and nothing ever clears it when the pane's underlying `Session` is replaced, so a restarted pane's fresh failure is swallowed entirely — not logged once, logged zero times.

Concretely: `pane-a` fails `DetectInteractiveAgentType` because its tmux server is gone, one warning is logged, `warned["pane-a"]["detect"]` is set. The operator restarts the pane. `manager.Get` now hands back a *different* `Session` object — `bootstrapped` and `givenUp` both handle that correctly, since they compare session identity — but `warned["pane-a"]` still carries the stale entry, and `checkPane`'s disappeared-pane arm (`sess, ok := b.manager.Get(paneID); if !ok`) deletes only from `pending`. The new session's detection failure is therefore never reported.

That is the same class of thing the PR body argues against for "once ever": a pane being torn down and recreated is a stronger recovery signal than a detection call that happened to succeed once, yet it's the one case where the operator hears nothing. Clearing `b.warned[paneID]` on the `!ok` arm (and/or when the observed session identity changes) would close it, and would also stop `warned` growing unboundedly for panes that no longer exist.

---

DEVELOPMENT.md
```diff
@@ -102,6 +102,30 @@ happens inside `internal/config`, `internal/server`'s integration tests drive co
 at once, and the root package's tests drive `internal/commandcenter`'s default paths — none of which
 a package-private variable can reach. See issue [#212](https://github.com/tomo-chan/panemux/issues/212).
 
+**The cache directory has its own seam, and it is `internal/cachedir`.** Same shape — `cachedir.Dir()`,
```
tomo-chan, line 105: This PR adds `./internal/cachedir/...` to both `COVERAGE_PKGS` and `coverage-go`'s test list in the Makefile, but line 187 below — the enumeration readers are pointed at for "which packages the 80% gate covers" — still ends at `internal/boardmcp`, `internal/homedir`, and the root package.

So a developer reading line 187 concludes `internal/cachedir` is ungated, then hits a `make coverage-go` failure the doc said couldn't happen. (`internal/fileops` is missing from the same list too, inherited from #231 — worth sweeping in while you're here, since the two omissions look identical to a reader.)

---

internal/commandcenter/runner_test.go
```diff
@@ -446,6 +446,12 @@ func (f *ctxAwareFakeCmd) Wait() error {
 	return f.ctx.Err() //nolint:wrapcheck // test fake mirrors exec.Cmd's own unwrapped ctx.Err() propagation
 }
 
+// fake stdout reaches EOF at once, so it never reaches the drain ordering the branch changes — that
+// is TestRunnerMalformedStreamJSONDoesNotWaitOnASubprocessThatStopsWritingWithoutExiting's job. The
+// red-check sees it as changed only because the ctxBlockingStdout helper added below it falls inside
+// this test's line range.
+//
+//efficacy:exempt unmodified by this branch: the cancellation it pins predates it, and this test's
```
tomo-chan: All four new `//efficacy:exempt` blocks have the reason sentence split with its tail placed *above* the marker, so the comment reads out of order and the marker line stops mid-clause.

Here: lines 449-453 end with `…falls inside this test's line range.`, then the marker line reads `unmodified by this branch: the cancellation it pins predates it, and this test's` — and stops. Same inversion at:

- `runner_test.go:765-768` — marker ends `…The red-check sees`, continued by `it as changed because…` three lines earlier
- `internal/boardmcp/boardmcp_branches_test.go:135-139` — marker ends `…dropping `omitempty` from Result, which`, continued above with `would emit `"result": null` beside every error…`
- `bootstrap_branches_test.go:159-163` — same shape

The tooling is satisfied either way (the marker carries a non-empty reason), but DEVELOPMENT.md's argument for preferring the marker over the label is precisely that "it sits in the diff a reviewer reads" — and split this way it can't be read in order. Putting the whole reason on and after the marker line fixes it with no behavior change.

---

tomo-chan (commented):
Careful single-pass review of the four fixes. The #230 drain reorder is correct — I checked that `run()` has no early return between `streamOutput` and the new `io.Copy`, so the drain still covers every post-`Start()` path and still precedes `cmd.Wait()`. `go vet ./...` and the `boardmcp` / `commandcenter` / `cachedir` / root suites pass on the branch.

Four findings, inline. The #210 one is the substantive one: the id half of the fix is right, but `"result": null` is not valid MCP, so the strictness argument the PR makes for it doesn't hold.

---

.golangci.yml
```diff
@@ -86,11 +97,17 @@ linters:
...
-      # supposed to name os.UserHomeDir.
+      # internal/homedir and internal/cachedir are the seams themselves: each
+      # is the one place that is supposed to name its own os function. They
+      # stay separate packages partly for this reason — one merged package
+      # would have to be excused from both rules, so a stray os.UserHomeDir
+      # inside the cache-directory seam would no longer be caught.
       - path: internal/homedir/
         linters:
           - forbidigo
+      - path: internal/cachedir/
```
tomo-chan: **The forbidigo exclusion excuses the whole linter, not one pattern** — so the justification this branch repeats in four places is false.

`linters.exclusions.rules` entries are per-*linter*, not per-pattern. Verified empirically on this branch: dropping a file into `internal/cachedir` containing `os.UserHomeDir()` and running `golangci-lint run ./internal/cachedir/` reports **no forbidigo issue at all** (only an unrelated `wrapcheck` one). The hole is symmetric — an `os.UserCacheDir()` call inside `internal/homedir` is equally unflagged by the rule at line 105.

So the claim in the comment above ("a stray `os.UserHomeDir` inside the cache-directory seam would no longer be caught" — implying it *is* caught today), and the same claim in `DEVELOPMENT.md:115`, `docs/architecture.md:186` and `internal/cachedir/cachedir.go:16`, does not hold. Concretely: someone later adds a darwin fallback inside `internal/cachedir` that reaches for the home directory directly, `make lint` stays green, and the one-seam invariant silently breaks — exactly the failure the separate-packages argument is documented as preventing.

Scoping the rule by text makes the claim true. Confirmed: with the line below added, the stray `os.UserHomeDir` in `internal/cachedir` *is* reported, and the package's own legitimate `os.UserCacheDir` calls (including `cachedir_test.go`'s) stay excluded.

```suggestion
      - path: internal/cachedir/
        linters:
          - forbidigo
        text: os\.UserCacheDir
```

The `internal/homedir/` rule at line 105 wants the matching `text: os\.UserHomeDir`.

---

internal/commandcenter/runner.go
```diff
@@ -286,17 +286,30 @@ func (r *Runner) run(ctx context.Context, prompt string, events chan<- Event) {
...
+	// blocked writing into a pipe no one reads, and every read completes
+	// before Wait() closes the pipe under it.
+	//
+	// This sits here rather than in a defer inside streamOutput because a
+	// deferred drain runs before the caller resumes, i.e. before the
+	// cancel() above: against a subprocess that stops writing but does not
+	// exit, it blocks on read(2) until the query timeout kills the process,
+	// so the cancellation meant to cut that wait short was itself queued
+	// behind it. Ordered this way the drain can only ever wait on a
+	// subprocess already being killed.
```
tomo-chan, line 310: **This invariant only holds on the stream-error path.** `cancel()` above is inside `if streamErr != nil`, so on a successful turn the drain runs with nothing having cancelled the context — the drain can then wait on a subprocess that is *not* being killed.

Concretely: a subprocess that emits a well-formed final `result` line and then keeps its stdout pipe open without exiting (the exact shape this branch was written for, just without a stream error) leaves `streamErr == nil`. `io.Copy(io.Discard, stdout)` blocks on `read(2)`, and `run()` — with it the busy flag and the turn's terminal frame — waits out the full `QueryTimeout`, same as before the change.

Not a regression (the old deferred drain behaved identically here), but the comment claims more than the code delivers, and `docs/behavior.md:703` states it unconditionally too ("The rest of the subprocess's output is drained only after that cancellation"). Either scope both to the stream-error path, or hoist `cancel()` out of the `if` so the claim becomes true for every turn.

---

bootstrap_test.go
```diff
@@ -159,7 +159,8 @@ func TestBootstrapWatcher_AgmsgNotPresent_NoWrite_WarnsOnce(t *testing.T) {
 	w.pollOnce(context.Background())
 
 	assert.Empty(t, sess.writes)
-	assert.True(t, w.presenceWarned["pane-a"])
+	assert.True(t, w.warned["pane-a"][warnKindAbsent],
+		"three ticks against a host with no agmsg must produce one warning, not three")
```
tomo-chan, line 168: **The message claims a log count this assertion never checks.** `assert.True` here only confirms the per-kind suppression flag is set; nothing counts lines, so the test passes whether one warning or three were emitted.

Revert `warnOnce` to an unconditional `log.Printf` that still records `b.warned[paneID][kind] = true` and this test stays green while three ticks produce three warning lines — the exact behavior the message says is forbidden. Line 232's "once per streak, not per tick" has the same gap.

The property really is covered, by `TestCheckPaneLogsADetectionFailureOncePerStreakAndDropsThePane` in `bootstrap_branches_test.go`, which counts. So either count occurrences here too (`captureBootstrapLog` is right there in the sibling file) or reword these two messages to describe the flag they actually assert — per DEVELOPMENT.md's own anti-pattern table, a message that overstates its assertion reads as coverage and is worth nothing.

---

.golangci.yml
```diff
@@ -86,11 +97,17 @@ linters:
...
+        linters:
+          - forbidigo
```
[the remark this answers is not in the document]
tomo-chan: Correct, and the hole was real. Both path exclusions are gone; the four intentional calls carry a line-level `//nolint:forbidigo` instead (`dirFn` in each package, and the one call in each package's own default-behavior test).

Confirmed by perturbation rather than by reading: adding `var dirFn2 = os.UserHomeDir` inside `internal/cachedir` now fails lint with the homedir rule's message, where it passed before. Each seam's doc comment records why the waiver is on the line rather than on the package.

---

internal/boardmcp/server.go
tomo-chan: Right — and that is the more important half of the finding, because the fix as written moved that response from one invalid shape to another while the commit message claimed it had made it valid. Applied.

`MarshalJSON` now substitutes `struct{}{}` for a nil `Result`, so `notifications/initialized` sent as a request answers `{"result":{},"id":7,"jsonrpc":"2.0"}`. The test is renamed `TestInitializedSentAsARequestIsAnsweredWithAnEmptyResultObject` and asserts by value (`map[string]any{}`) rather than only by key presence, since "present but null" is exactly the shape that had to stop being produced. E11 and agent-board.md's wire-shape bullet both name the MCP half now, including why §5 alone was not enough.

---

bootstrap.go
tomo-chan: Agreed, and the framing is the part worth keeping: a pane being torn down and recreated is the strongest recovery signal there is, and it was the one case that produced silence rather than one line.

`warned` is now `map[string]warnedPane`, where `warnedPane` carries the suppressed kinds *and* the session they were recorded against. `checkPane` reaches its session through a new `sessionFor` helper, which drops the whole entry when `manager.Get` fails and resets it when the identity changes — so the map also stops growing for panes that no longer exist.

One consequence worth flagging, because it drove the shape: `clearWarning` no longer deletes the entry when its last kind clears. It cannot — the entry is what carries the identity, so deleting it would leave the next warning recorded against no session, and the following tick would read that as a replacement and re-warn, every tick. Its doc comment says so.

Two tests: `TestARestartedPaneWarnsAboutTheSameFailureAgain` (same pane ID, new `Session`, the failure is reported again) and `TestAPaneThatDisappearsDropsItsSuppressedWarnings`. C9 and agent-board.md both mention the identity keying now.

`checkPane` crossed `funlen`'s 60-line limit with the extra bookkeeping inline, which is what `sessionFor` is for rather than a `//nolint`.

---

DEVELOPMENT.md
tomo-chan: Fixed, both of them: the list now names `internal/fileops` and `internal/cachedir`, and ends with "The `Makefile`'s `COVERAGE_PKGS` is the authority; this list follows it" — so the next reader knows which side to trust when they drift again, rather than only this one being current.

---

internal/commandcenter/runner_test.go
tomo-chan: Fixed — and the cause is worth recording, because it will do the same to the next person: **gofmt moves a `//directive`-shaped comment (`//word:word`, no space) to the end of its doc-comment block.** I wrote each marker first with the reason running on after it; `gofmt -w` reordered them, which is why every one of them came out split the same way.

So "put the whole reason on and after the marker line" is not reachable while the file has to be gofmt-clean. Each block now reads: prose paragraph, blank `//`, then a single self-contained marker line carrying a one-line reason — in order, with the marker last where gofmt keeps it. The `runner_test.go:764` one puts its paragraph inside the function body instead, since a comment block there also shifts the red-check's line attribution for the *preceding* test.

Two further exemptions came out of the other review fixes: `internal/homedir`'s `TestDirDefaultsToTheOperatingSystemHomeDirectory` (this branch touches only the `//nolint:forbidigo` on its call, now that the package-wide waiver is gone), and none removed. `EFFICACY_BASE=origin/main make efficacy` is green — 19 changed tests, all red on revert.

---

internal/commandcenter/runner.go
tomo-chan: **Withdrawing this finding — it was wrong, and leaving it standing risks a change that would make things worse.** The comment and `docs/behavior.md:703` are correct as written; nothing here needs fixing.

The premise I argued from does not hold. `streamOutput` is a `for scanner.Scan()` loop, so while the subprocess keeps stdout open it blocks *inside* `Scan()` and never returns — the "clean final `result` line, pipe still open" case therefore never reaches the drain at all. `streamOutput` returns a nil error in exactly one situation: `Scan()` returned false with `scanner.Err() == nil`, i.e. EOF. At that point `io.Copy(io.Discard, stdout)` finds the pipe already at EOF and returns immediately, waiting on nothing. Every non-EOF exit (`json.Unmarshal` failure, `scanner.Err()`, including `ErrTooLong`) yields `streamErr != nil` and so has already gone through `cancel()`.

So both arms satisfy the invariant: with `streamErr != nil` the drain waits only on a subprocess being killed, and with `streamErr == nil` it does not wait at all.

Also worth saying explicitly, since it was one of the two remedies I proposed: **do not hoist `cancel()` out of the `if`.** On a successful turn that would cancel the context between stdout EOF and `cmd.Wait()`, i.e. kill a subprocess that was about to exit cleanly, and hand `finishAfterStream` a cancelled context on the normal path. The current placement is the correct one.

---

.golangci.yml
tomo-chan: Confirmed fixed at `70e896d`, and verified by perturbation in both directions rather than by reading the config:

- `os.UserHomeDir()` added inside `internal/cachedir` → now reported (`use of os.UserHomeDir forbidden … call homedir.Dir() instead`).
- `os.UserCacheDir()` added inside `internal/homedir` → now reported with the symmetric message.

Both passed silently before the change, so the hole is closed on both sides. The line-level `//nolint:forbidigo` route taken here is stronger than the `text:`-scoped exclusion I suggested: it waives the four intentional calls individually instead of every `os.UserCacheDir` mention in the directory, and the waiver sits next to the call a reader is looking at. `golangci-lint run ./...` is still 0 issues.

---

bootstrap_branches_test.go
```diff
@@ -112,9 +112,166 @@ func TestCheckPaneLogsADetectionFailureAndDropsThePane(t *testing.T) {
...
+	require.NotEmpty(t, w.warned["pane-a"].kinds)
+
+	require.NoError(t, manager.Remove("pane-a"))
+	w.pollOnce(context.Background())
+
+	assert.NotContains(t, w.warned, "pane-a")
+}
+
+// Clearing one kind's streak must leave the others' suppression intact: the
+// per-pane entry is only dropped once nothing is suppressed for that pane, so
```
tomo-chan: **This sentence became false in the same commit that added the tests above it.** "the per-pane entry is only dropped once nothing is suppressed for that pane" describes the `clearWarning` that `c08ed30` replaced — it no longer deletes the entry when its last kind clears, deliberately, because the entry is what carries the session identity `noteSession` compares against. Its own doc comment in `bootstrap.go` now says exactly that, and `sessionFor`'s disappeared-pane arm is the only place the entry is dropped.

The assertion itself is unaffected and still pins the right property; only the explanation above it now contradicts the code. It reads as a claim about the map's lifecycle, which is the one part of this change a later reader is most likely to take at face value and then "fix" back to the deleting form — the shape that re-warns every tick.

Something like: *the per-pane entry outlives its kinds — it carries the session identity — so clearing one kind leaves both the entry and the other kinds' suppression intact, and a recovered detection does not re-open the presence warning the operator has already been shown.*

---

bootstrap.go
```diff
@@ -192,6 +226,22 @@ func (b *bootstrapWatcher) pollOnce(ctx context.Context) {
 // checkPane runs the bootstrap decision for one pane. See
 // docs/agent-board.md's Bootstrap flow section for the algorithm this
 // implements.
+// sessionFor returns the pane's live session, or false when the manager no
```
tomo-chan, line 226: **`sessionFor`'s godoc now opens with someone else's sentence.** Lines 226-228 are `checkPane`'s doc comment, and with no blank line between them the two blocks are one comment attached to this function — so `go doc` renders `sessionFor` as "checkPane runs the bootstrap decision for one pane. See docs/agent-board.md's Bootstrap flow section for the algorithm this implements. sessionFor returns the pane's live session…".

Pre-existing in the sense that the comment was already detached — before this branch it sat above `modeFor` with the same missing blank line — so this is not a regression, just the same wart now landing on the function this PR introduces, and inside the diff a reviewer is reading. Two options, both a one-line change:

- move lines 226-228 down to immediately above `func (b *bootstrapWatcher) checkPane` (line 259), where they belong; or
- if that is more churn than this PR wants, insert a blank `//`-less line between 228 and 229 so at least the two blocks stop merging — though that leaves the comment orphaned between two unrelated functions, so the first is the real fix.

---

tomo-chan (commented):
Re-reviewed the two fix commits (`c08ed30`, `70e896d`) against `9042b22`.

**Five of the six findings are fixed, and I verified each by perturbation or by running the gate rather than by reading the diff:**

- **`boardmcp` empty result object** — `MarshalJSON` substitutes `struct{}{}` for a nil `Result`, and `resultResponse.Result` carries no `omitempty`, so `{"result":{},...}` really is what goes on the wire. Test renamed and asserting by value; E11 and agent-board.md updated.
- **`bootstrap` session-identity keying** — both new tests are load-bearing, confirmed by breaking the implementation under them: neutering `noteSession`'s identity comparison fails `TestARestartedPaneWarnsAboutTheSameFailureAgain` (1 warning where 2 are required), and removing `delete(b.warned, paneID)` fails `TestAPaneThatDisappearsDropsItsSuppressedWarnings`.
- **`.golangci.yml` forbidigo hole** — closed on both sides; details in the thread.
- **`DEVELOPMENT.md` coverage list** — `internal/fileops` and `internal/cachedir` added, with the Makefile named as the authority so the next drift is resolvable.
- **`//efficacy:exempt` ordering** — all five blocks read in order now. The gofmt directive-comment behavior recorded in the reply is the useful half; it explains why the obvious layout is unreachable.

**One finding I am withdrawing: the `runner.go` drain invariant.** It was wrong — `streamOutput` blocks inside `Scan()` while stdout is open, so a nil `streamErr` means EOF was reached and the drain waits on nothing. The comment and `docs/behavior.md` are correct as written, and the `cancel()` hoist I floated would actively harm the success path. Full reasoning in that thread, now resolved.

**Still open:** the `bootstrap_test.go:163`/`:232` assertion messages, which claim a log count neither assertion checks. Lowest-value item here, and the counting coverage genuinely exists elsewhere — but it is a two-line reword.

**Two new, minor observations** from the fix commits themselves, inline below: a doc comment in `bootstrap_branches_test.go` that this change made false, and `checkPane`'s doc comment now merging into `sessionFor`'s.

Gates run at `70e896d`: `gofmt -s` clean, `go vet ./...` clean, `golangci-lint run ./...` 0 issues (local binary is 2.5.0 against the pinned v2.12.2, so CI remains the authority there), `go test ./...` all green, `scripts/scenarios_check.sh` resolves 233 names across 98 auto rows, `EFFICACY_BASE=origin/main` red-check 19/19 red, and diff-scoped `coverage-blocks` reports nothing unexecuted.

---

bootstrap_test.go
tomo-chan: Right, and the revert you describe is exactly the check I ran rather than taking the reasoning on trust. Both assertions now count.

`TestBootstrapWatcher_AgmsgNotPresent_NoWrite_WarnsOnce` asserts `strings.Count(logs.String(), "agmsg not found on host") == 1`, and the transport-error test the same for `"checking agmsg presence on host"`; the flag assertion stays underneath, reworded to say what it actually is — that the count came from *that* kind's suppression rather than from something else not firing.

Confirmed load-bearing by neutering `warnOnce`'s suppression (`if false && entry.kinds[kind]`): both fail with `expected: 1 / actual: 2`, where before they stayed green.

---

bootstrap_branches_test.go
tomo-chan: Correct — the sentence survived the change that falsified it, and the risk you name is the real one: it reads as a lifecycle claim, and the "fix" it invites is the deleting form that re-warns every tick.

Rewritten roughly as you suggest, plus the consequence spelled out so the next reader has the reason and not only the rule: the entry outlives its kinds because it carries the session identity `noteSession` compares against, `sessionFor`'s disappeared-pane arm is the only place it is dropped, and deleting it here would leave the next warning recorded against no session — which the following tick reads as a replacement.

---

bootstrap.go
tomo-chan: Took the first option — the comment is now directly above `func (b *bootstrapWatcher) checkPane`, where it belongs. It is three lines of movement, and leaving an orphaned block between two unrelated functions would have preserved the thing that made this reachable in the first place.

`go doc` now renders `sessionFor` with its own sentence first.

---

PR #235, opened 2026-09-12
fix: report a mutant that reached no verdict as undecided, not as passing

tomo-chan:
Closes the first of the two 保留 items left open on #180's roadmap item 6, ahead of stage 4.

## The defect

`scripts/mutation.sh` matched `LIVED`, routed `SKIPPED` to an "unanalysed" list, and let a catch-all `*)` arm drop every other status. `TIMED OUT` and `NOT VIABLE` fell into it silently, and so would any status a later gremlins invents.

`TIMED OUT` is why this matters rather than a tidiness point. It is the status the pinned `--timeout-coefficient` / `--workers` settings exist *because* of: with gremlins' defaults, 465 of 1059 runnable mutants on this repository come back timed out, and #180's measurement found those timeouts hiding 51 survivors. So a run in which every mutant on the diff timed out printed

```
mutation: no surviving mutants on lines this branch changed.
```

— a true sentence about a run that measured nothing. The script's own header already forbids exactly this for the run as a whole ("a warning that could not run must not look like a warning that found nothing"); it just never applied the rule one mutant at a time.

## The fix

**Which arm carries the catch-all.** The enumeration that has to be exhaustive is the *silencing* one, and it was on the wrong side.

- `LIVED` → a survivor, as before.
- `KILLED`, `NOT COVERED`, `NOT VIABLE` → enumerated as the three states the gate deliberately says nothing about, each with its reason inline. `NOT VIABLE` belongs here rather than among the unknowns: a mutant that did not compile is not a hole in anyone's tests, and listing it would put noise in the one section whose whole value is that everything in it is genuinely unknown.
- **everything else** → reported as undecided, carrying the status string that produced it.

**In the headline, not only in a section below it.** "0 survivors" and "0 survivors, 12 of 12 mutants undecided" are different results, and the headline is the line a reviewer reads:

```
mutation: no surviving mutants on lines this branch changed, 2 of 3 mutant(s) undecided.

  Undecided — these mutants sit on lines this branch changed and never
  reached a verdict, so nothing is known about them either way:
    pkg/new.go:4  CONDITIONALS_BOUNDARY  (TIMED OUT)
```

Exit code is unchanged: stage 3 warns. Whether an undecided mutant should *fail* is now a question stage 4 inherits, and D9 records it as such.

## Also in this change

`--help` printed a hand-counted line range (`sed -n '2,80p'`), which this very change pushed out of date — it began printing `set -u` as documentation. It now prints to the first non-comment line, so the header can grow without the help going quietly wrong. Test 21 pins both ends: too little is the worse failure, since nobody notices a missing paragraph.

## Two test-side corrections, found by perturbation rather than by reading

Both were pre-existing, and both made assertions pass for the wrong reason:

- **`findings_only()` stopped only at the `Exempt by` header.** On a report with no exempt section it returned the whole output, so every "is not reported as a finding" assertion held regardless. It now stops at either section header — in `awk`, because the two-header form needs alternation and BRE's `\|` is a GNU extension macOS's `sed` lacks (the portability class of bug #233 just fixed in `scenarios_check.sh`).
- **Asserting a status by grepping the whole output matched the section's own explanatory prose.** A perturbation replacing every recorded status with a placeholder left all three status assertions passing. They now require the status on the row that names the file and line.

## Verification

- 21 checks, all passing under `dash`, `bash` and `sh`.
- The five new/changed assertions go red against the pre-change script; case 18 (`NOT VIABLE` stays silent) pins deliberate existing behaviour and is not expected to.
- Six perturbations, each caught by exactly the assertion it should be: catch-all restored; `NOT VIABLE` moved to undecided; headline note removed; status dropped from the record; undecided leaking into findings; `findings_only()` reverted.
- `make check` green.

`make efficacy` does not apply — it considers `*_test.go`, and this branch changes no Go.

## Test plan

- [x] `sh scripts/mutation_test.sh` — 21 checks pass
- [x] Same under `dash` and `bash`
- [x] New assertions confirmed red against the pre-change script
- [x] Six perturbations confirmed caught
- [x] `make check`
- [x] `sh scripts/mutation.sh --help` prints the whole header and no code
- [x] `sh scripts/mutation.sh --base origin/main` against the real repository

https://claude.ai/code/session_01VTfbAi8NhyieDyRgcaC7qN

---


scripts/mutation.sh
```diff
@@ -108,7 +119,11 @@ while [ $# -gt 0 ]; do
 		shift 2
 		;;
 	-h | --help)
-		sed -n '2,80p' "$0" | sed 's/^#\{1,2\} \{0,1\}//'
+		# The header, however long it is. A hand-counted line range was
+		# silently wrong the moment the header grew — it had already fallen
```
tomo-chan: **The rationale here states a failure that never happened.** `scripts/mutation.sh` has exactly one prior revision (`124a64d`), and in it the header ended at line 80 with `set -u` at line 82 — so `sed -n '2,80p'` printed the header *exactly*. The stale range could only ever **truncate**, never leak code.

I reconstructed the old range against this PR's longer header: `--help` then stops mid-"PINNED GREMLINS SETTINGS", prints **0** `set -u` lines and never reaches `Exit codes:`.

The change itself is right; only the stated history is wrong, and a reader will go looking for the wrong failure mode.

```suggestion
		# The header, however long it is. A hand-counted line range went
		# stale the moment the header grew — it silently truncated the
		# tail, dropping whole paragraphs with nothing to show for it.
```

---

scripts/mutation_test.sh
```diff
@@ -606,6 +619,181 @@ else
...
+checks=$((checks + 1))
+out=$(sh "$checker" --help 2>&1)
+rc=$?
+if [ "$rc" -ne 0 ]; then
+	fail "--help exits 0" "exit $rc: $out"
+elif ! printf '%s' "$out" | grep -q 'Exit codes:'; then
+	fail "--help prints the header through its last line" "$out"
+elif ! printf '%s' "$out" | grep -q 'UNDECIDED'; then
+	fail "--help prints the middle of the header, not just its ends" "$out"
+elif printf '%s' "$out" | grep -q '^set -u'; then
```
tomo-chan: **This assertion can never fire, under either implementation.** The new awk stops at the first non-`#` line after line 1, so `set -u` is unreachable by construction; and the old `sed -n '2,80p'` never printed it either — simulating that range against this PR's header yields **0** `set -u` lines and **0** `Exit codes:` lines. Truncation was the only regression ever possible, and the `Exit codes:` assertion above already covers it.

The case comment at line 777 ("adding the paragraph case 17 documents pushed `set -u` into the output") describes the opposite of what growing the header does — growing it drops the tail.

Worth replacing with something that can actually go red, e.g. asserting the output contains no line that is not a stripped comment (a leaked `base=${MUTATION_BASE:-}` would be caught), and fixing the comment above.

---

scripts/mutation_test.sh
```diff
@@ -606,6 +619,181 @@ else
...
+EOF
+write_report "$repo/rep.json" '{"go_module":"example","files":[
+ {"file_name":"pkg/new.go","mutations":[
+   {"type":"CONDITIONALS_BOUNDARY","status":"TIMED OUT","line":4,"column":5},
+   {"type":"CONDITIONALS_NEGATION","status":"TIMED OUT","line":4,"column":5},
+   {"type":"CONDITIONALS_BOUNDARY","status":"KILLED","line":7,"column":5}]}]}'
+commit_on_branch "$repo" "add new.go"
+out=$(run_checker "$repo" --base main --report rep.json)
+headline=$(printf '%s\n' "$out" | head -1)
+if ! printf '%s' "$headline" | grep -q '2'; then
```
tomo-chan: **These match bare digits, not the sentence.** `grep -q '2'` and `grep -q '3'` cannot tell the numerator from the denominator, nor either from an unrelated number in the headline.

Concretely: if a regression swapped the operands and the headline read `3 of 2 mutant(s) undecided`, or the denominator came from the wrong counter and it read `23 of 3`, all three assertions still pass and case 20 reports `ok`. Since the whole point of the case is that the headline is the line a reviewer reads, it should pin the phrasing:

```suggestion
if ! printf '%s' "$headline" | grep -q '2 of 3 mutant(s) undecided'; then
	fail "the headline says how many mutants were left undecided, of how many" "$out"
```

(and the two `elif` arms below collapse into it).

---

scripts/mutation.sh
```diff
@@ -321,19 +340,32 @@ for f in $changed; do
...
 
+		scoped_count=$((scoped_count + 1))
+
+		# THE ENUMERATION THAT HAS TO BE EXHAUSTIVE IS THE SILENCING ONE, and
+		# the catch-all used to be on the other arm. A status this script did
+		# not recognise fell through to `continue` and vanished — TIMED OUT and
+		# NOT VIABLE did exactly that, and so would any status a later gremlins
+		# invents. The default is now "nothing is known about this", which is
+		# the only honest thing to say about a verdict string nobody matched.
 		case $status in
```
tomo-chan, line 362: **Reworking these arms invalidates the comment at lines 327-332, which was not updated.** That comment defends the tab `OFS` by saying a whitespace split of `"NOT COVERED"` into `status="NOT"` "lands in the same branch either way, so nothing visibly breaks — which is precisely why it would have survived until a status this gate does act on gained a space."

After this change that is no longer true. `NOT` matches neither `LIVED` nor `KILLED | "NOT COVERED" | "NOT VIABLE"`, so it falls into the new `*)` arm and every `NOT COVERED` / `NOT VIABLE` mutant would be reported as **undecided** — loudly, immediately, and wrongly. The hazard the comment calls invisible is exactly the kind of thing the new default arm surfaces.

The code is fine; the comment now understates why the tab separation is load-bearing and should say so.

---

scripts/mutation.sh
tomo-chan: You're right, and I checked it rather than taking it on faith. `124a64d`'s header ends at line 80 with `set -u` at 82, so `sed -n '2,80p'` printed that header exactly. Simulated against this branch's header, the old range prints 0 `set -u` lines and 0 `Exit codes:` lines — it stops mid-sentence in "PINNED GREMLINS SETTINGS", exactly as you describe.

The leak was real but self-inflicted: a hand-count of `2,92p` I made *while writing this change*, against a header that then ended at 89. I then wrote that history into the comment as if the repository's own range had done it.

Fixed in `abf03ef`, phrased as what a stale range actually does — it truncates, quietly. I kept your framing rather than your exact wording only because the simulated evidence was worth naming in the comment.

---

scripts/mutation_test.sh
tomo-chan: Agreed on the substance, with one correction to the margin of it.

The arm is not *wholly* unable to fire: perturbing the range to `2,95p` does print `set -u` and does fail it. What it cannot do is distinguish the two implementations it sat between — neither `2,80p` nor the awk ever prints it — and the history written above it was false either way, which is the part that matters.

Replaced in `abf03ef` with an assertion on the **last non-blank line** of the output. That one carries both failure modes: a range that stops short ends on some other sentence, a range that overshoots ends on `set -u`. Verified in both directions — restoring the original `2,80p` and an overshooting `2,95p` each fail it, and nothing else.

The case comment now records the real history, including that the leak was a hand-count made while writing the change rather than anything in the repository.

---

scripts/mutation_test.sh
tomo-chan: Correct, and your example reproduces exactly. I swapped the operands to make the headline read `3 of 2 mutant(s) undecided` and ran the suite against the old assertions: **case 20 reported `ok`**. All three arms passed on a headline with the numbers inverted.

Fixed in `abf03ef` with your suggestion — the three arms collapse into one pinned phrase, and the swap now fails it.

One knock-on worth flagging since these PRs are stacked: #237 moves the denominator out of the undecided note and into the scope note, so the sentence there becomes `among 3 on lines this branch changed, 2 undecided`. Merging this fix down surfaced that as a **failing** case rather than letting it slide on a substring, which is the behaviour wanted from a pinned phrase; #237 carries the updated sentence and I re-verified the swap still fails it there.

---

scripts/mutation.sh
tomo-chan: This is the sharpest of the four, and it reproduces. Dropping the reader's `IFS="$tab"` alone — so `NOT COVERED` splits to `status="NOT"` — now fails **four** cases:

```
FAIL: NOT COVERED is not reported as a survivor
FAIL: the undecided list says WHY THIS mutant has no verdict
FAIL: a NOT VIABLE mutant is not reported as undecided
FAIL: an unrecognised status is quoted back, so a reader can act on it
```

with every one of them showing `1 of 1 mutant(s) undecided` — the split status announced as an unknown, exactly as you said. Before this change the same perturbation would have failed none, because `NOT` and `NOT COVERED` both fell into the one `continue`. That is the comment's own argument turned around: the hazard it called invisible is precisely what the new default arm surfaces.

Rewritten in `216ee4d`. The note now says the stake rose with the catch-all — the separator went from load-bearing and silent to load-bearing and loud — and cites the four-case perturbation so the claim is checkable rather than asserted.

Thanks for catching this one; the code was fine and I would not have re-read that comment.

---

scripts/mutation.sh
```diff
@@ -44,6 +44,17 @@
 # that could not run must not look like a warning that found nothing, which is
 # the rule scripts/efficacy.sh and scripts/coverage_blocks.sh both state.
 #
+# THE SAME RULE, ONE MUTANT AT A TIME. gremlins reports six statuses. This
```
tomo-chan: **`gremlins reports six statuses` miscounts — there are seven, and the missing one is `RUNNABLE`.**

v0.6.0's `internal/mutator/mutator.go` defines:

```go
const (
	NotCovered Status = iota // "NOT COVERED"
	Runnable                 // "RUNNABLE"
	Skipped                  // "SKIPPED"
	Lived                    // "LIVED"
	Killed                   // "KILLED"
	NotViable                // "NOT VIABLE"
	TimedOut                 // "TIMED OUT"
)
```

Seven. The census here adds up to six exactly because `RUNNABLE` is absent: 1 acted on (`LIVED`) + 3 silenced (`KILLED`, `NOT COVERED`, `NOT VIABLE`) + 2 named in the catch-all (`SKIPPED`, `TIMED OUT`). The same sentence is repeated in `DEVELOPMENT.md:182` and `docs/quality-gateway.md:367`.

The behaviour is safe — this change is precisely what makes it safe, since `RUNNABLE` now lands in the catch-all rather than being dropped. But the count is the auditable claim of a change whose whole premise is that the *silencing* enumeration must be exhaustive and verifiable, so a reader checking the three enumerations against gremlins finds a discrepancy with no way to tell which side is wrong.

There is also a user-visible edge. Running the branch's script against a report containing a `RUNNABLE` mutant on a changed line:

```
  Undecided — these mutants sit on lines this branch changed and never
  reached a verdict, so nothing is known about them either way:
    pkg/new.go:7  CONDITIONALS_NEGATION  (TIMED OUT)
    pkg/new.go:7  CONDITIONALS_BOUNDARY  (RUNNABLE)

  SKIPPED means ... TIMED OUT means ... Anything else is a status this
  gate does not recognise, which is itself worth looking at.
```

The trailer sends the reader to investigate a documented gremlins status as if it were one a later release invented. `RUNNABLE` is reachable in a real report (`--dry-run` leaves every mutant in it), so this is not purely hypothetical.

Suggest correcting the count to seven and naming `RUNNABLE` alongside `SKIPPED` and `TIMED OUT` in the catch-all — in all three places, and in `undecided_list`'s trailer, which is the one a reader actually sees.

---

tomo-chan: Both halves confirmed against the module source, and both fixed in `1b53627`.

**The count.** `internal/mutator/mutator.go`'s const block is exactly as you quote it — seven. And your diagnosis of *why* it read as six is right: my census was 1 + 3 + 2, and it balanced only because RUNNABLE was missing from it. That is the worst way for a miscount to hide, since the arithmetic looks self-checking.

**The reachability.** I traced it rather than taking `--dry-run` on trust: `internal/engine/executor.go:171` returns early on `m.dryRun` *before* `m.mutant.Apply()`, so the mutant is emitted with its status untouched. RUNNABLE genuinely survives into the JSON.

Corrected in four places, not three — the script header, the catch-all arm's own comment, `DEVELOPMENT.md` and `docs/quality-gateway.md` — plus the trailer. The trailer now reads:

```
  hiding real survivors. RUNNABLE means the mutant was identified and
  covered but never run, which is what --dry-run leaves behind.
  Anything else is a status this gate does not recognise, which is
  itself worth looking at.
```

Case 22 pins the trailer half specifically. The row already named RUNNABLE — the catch-all did that much — so the assertion excludes the row and looks only at the explanation; removing that sentence fails case 22 and nothing else.

Propagated to #236 (`6583f7b`) and #237 (`d9b82c0`). Both merges conflicted, for the same reason each time: an independent case landed at the same insertion point with the same number. Kept both sides and renumbered, so RUNNABLE is case 30 there and 34 on #237. 22 / 30 / 34 checks green, `make check` green on all three.

---

